PR #14, opened 2012-12-19
Adding support for .Net 40 Directory methods

pranavkm:
A few methods were introduced to the System.IO types in .Net40. I was specifically interested in the Directory.EnumerateFiles method. 

There seems to be a bug in NuGet.exe at this point which causes it to not pick up the assembly from the Net40 project. I can get you a fix for that independently.

### DIFF
--- a/System.IO.Abstractions.Net40/DirectoryBase.Net40.cs
+++ b/System.IO.Abstractions.Net40/DirectoryBase.Net40.cs
@@ -1,0 +1,11 @@
+﻿using System.Collections.Generic;
+
+namespace System.IO.Abstractions
+{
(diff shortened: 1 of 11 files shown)


tathamoddie:
This PR has been around forever, mostly because I never went and supported multi-framework builds. In 94a9fbcbe67c86b187c6d19de36d20482203602d, I've now just bumped the baseline up to .NET 4.0 across the board, and then in 00b077bfbbd1b405a76e4cadfd956b4eeb14de07 I've added the `EnumerateFiles` signatures.

laedit:
Thanks! :+1:

PR #60, opened 2014-02-22
Created FileSystemWatcherFactory class to to make it easier to test code...

KallDrexx:
... that needs to generate file system watchers

Sorry, I realized after I started integrating this into my project that I needed some testable way to generate file system watchers.  

Not totally sure the MockFileSystemWatcher is useful, but i needed to create something for the MockFileSystemWatcherFactory to create.

### DIFF
--- a/TestingHelpers/MockFileSystemWatcher.cs
+++ b/TestingHelpers/MockFileSystemWatcher.cs
@@ -1,0 +1,38 @@
+﻿using System.ComponentModel;
+
+namespace System.IO.Abstractions.TestingHelpers
+{
+    public class MockFileSystemWatcher : FileSystemWatcherBase
+    {
+        public override bool IncludeSubdirectories { get; set; }
+        public override bool EnableRaisingEvents { get; set; }
+        public override string Filter { get; set; }
+        public override int InternalBufferSize { get; set; }
+        public override NotifyFilters NotifyFilter { get; set; }
+        public override string Path { get; set; }
+#if NET40
+        public override ISite Site { get; set; }
+        public override ISynchronizeInvoke SynchronizingObject { get; set; }
+#endif
+
+#if NET40
+        public override void BeginInit()
+        {
+        }
+
+        public override void EndInit()
+        {
+        }
+#endif
+
+        public override WaitForChangedResult WaitForChanged(WatcherChangeTypes changeType)
+        {
+            throw new NotImplementedException();
+        }
+
+        public override WaitForChangedResult WaitForChanged(WatcherChangeTypes changeType, int timeout)
+        {
+            throw new NotImplementedException();
+        }
+    }
+}
--- a/System.IO.Abstractions/FileSystem.cs
+++ b/System.IO.Abstractions/FileSystem.cs
@@ -39,5 +39,11 @@
         {
             get { return driveInfoFactory.Value; }
         }
+		
+        private IFileSystemWatcherFactory fileSystemWatcherFactory;
+        public IFileSystemWatcherFactory FileSystemWatcher
+        {
+            get { return fileSystemWatcherFactory ?? (fileSystemWatcherFactory = new FileSystemWatcherFactory()); }
+        }
     }
 }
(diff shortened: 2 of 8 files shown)


jpreese:
@KallDrexx, we actually have an open issue for this currently (#167) if you have any desire to resolve the conflicts and work towards getting this merged. Otherwise we can try and take it over. Thanks for putting this together!

KallDrexx:
I just did a rebase and fixed the merge conflicts.  However, that's a lot of history to go through (especially going from a .net framework to dnf + dnc conversion) so I'd double check my work just in case.  All tests did pass for me though.

KallDrexx:
Woops removed Rider's .idea directory

jpreese:
Thanks for the response, @KallDrexx. Just wanted to pop in real quick to say I'm looking at this.

KallDrexx:
Sorry I just noticed the merge tool did some funkyness with the whitespace in some of the files.  I can normalize those instances in a few.

TestingHelpers/MockFileSystemWatcher.cs
```diff
@@ -0,0 +1,38 @@
...
+        public override bool IncludeSubdirectories { get; set; }
+        public override bool EnableRaisingEvents { get; set; }
+        public override string Filter { get; set; }
+        public override int InternalBufferSize { get; set; }
+        public override NotifyFilters NotifyFilter { get; set; }
+        public override string Path { get; set; }
+#if NET40
+        public override ISite Site { get; set; }
+        public override ISynchronizeInvoke SynchronizingObject { get; set; }
+#endif
```
jpreese: This is probably mostly for me, but where are you finding that certain members need to be specific to 4.0? I've dug through the documentation for the FileSystemWatcher (https://msdn.microsoft.com/en-us/library/system.io.filesystemwatcher(v=vs.110).aspx) as well as a couple others, but I'm not seeing anything that definitively says to wrap these.

I thought it was going to be the "Version Information" towards the bottom, but then say DriveInfo's GetAccessControl (https://msdn.microsoft.com/en-us/library/yk1cww0c(v=vs.110).aspx) only lists 4.0, but we don't have the NET40 check around that method.

jpreese: To follow up, it looks like it's in the FileSystemWatcherBase, though I'm still not even sure how the original author found that requirement.

jpreese:
Thanks @KallDrexx 🎉

TestingHelpers/MockFileSystemWatcher.cs
KallDrexx: Yeah I only put those in because of the base class, otherwise it wouldn't compile for both .net framework and .net standard. 

It appears that .net standard doesn't have the `ISite` and `ISynchronizeInvoke` interfaces defined in them (as Microsoft pulled them out into separate nuget packages).  That may or may not have existed when the base class was written and so the `#if` was put in place (at least that's my guess).

visorz:
I would like to have a working FileSystemWatcher that reacts to changes made on the MockFileSystem as in SO question [here](https://stackoverflow.com/questions/47031711/how-can-i-use-filesystemwatcher-from-system-io-abstractions-to-monitor-a-mock-fi).

```
/// create mock filesystem
var filesystem = new MockFileSystem();
filesystem.Directory.CreateDirectory(MyFolderPath);

/// create and start file watcher
var watcher = filesystem.FileSystemWatcher.CreateNew();
watcher.Path = MyFolderPath;
watcher.Filter = "*.*";
watcher.NotifyFilter = NotifyFilters.LastWrite | NotifyFilters.FileName;
watcher.Created += new FileSystemEventHandler(OnCreated);
watcher.EnableRaisingEvents = true;

/// create file to raise watcher event
filesystem.File.Create(MyFilepath); // does not raise event, OnCreated() is not called
```

What do I need to do additionally to make the watcher raise OnCreated? Alternatively, could you please provide a minimal but working FileSystemWatcher on MockFileSystem example?

PR #68, opened 2014-03-08
File modify

manne:
The changes are necessary, because some methods have not persisted the existing BOM and mixed up the content of an existing file.

The default encoding for files is UTF-8 without BOM, like in .Net.
MockFileData.TextContents uses the default encoding, too. 
So setting value via TextContent can destroy the encoding of the file. I am not satisfied with this implementation. However it matches the behavior of .Net very close.

Adapted two test cases, because they were not correct.

@tathamoddie if you need help, just ask

### DIFF
--- a/TestHelpers.Tests/MockFileTests.cs
+++ b/TestHelpers.Tests/MockFileTests.cs
@@ -30,6 +30,27 @@ namespace System.IO.Abstractions.TestingHelpers.Tests
         }
 
         [Test]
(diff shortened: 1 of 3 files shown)


tathamoddie:
Hey @manne! This doesn't merge cleanly any more; sorry. :/ After your slew of wonderful pull requests though, I've just added you as a collaborator, so now I don't have to be a bottleneck any more.

manne:
resolved with #107

PR #114, opened 2015-02-22
Add EnumerateDirectories testing support

asarium:


tathamoddie:
Unit test?

asarium:
I added the unit tests. They are the same as for GetDirectories, is that a problem?

PR #120, opened 2015-03-11
Implemented WriteAllLines with input parameter IEnumerable<string>

Lyynx:
Added unit test to check output file has correct content. :)


tathamoddie:
Thanks! Can you add a unit test?

tathamoddie:
Thanks! Published to NuGet as 2.0.0.111

PR #124, opened 2015-03-23
Implemented EnumerateStuff methods

martindevans:
Implemented:
 - EnumerateDirectories()
 - EnumerateDirectories(string)
 - EnumerateDirectories(string, SearchOption)
 - EnumerateFiles() __replaced existing implementation__
 - EnumerateFiles(string)
 - EnumerateFiles(string, SearchOption)
 - EnumerateFileSystemInfos()
 - EnumerateFileSystemInfos(string)
 - EnumerateFileSystemInfos(string, SearchOption)

Every one of these is simply a call to the equivalent GetStuff method and immediately returning the result with no extra work. I have *replaced* the implementation of Enumerate files so that it calls out to GetFiles, the previous implementation seemed to be broken - as far as I can tell it would always return *all* files!

Added test coverage for most of these methods, which essentially just consisted of copying and pasting the appropriate GetStuff method and modifying it to call EnumerateStuff instead. Changed the MockDirectoryInfo_EnumerateFiles_ShouldReturnAllFiles test method so that it has some additional files (i.e. ones which should *not* be returned).

**Important**: I noticed that there is a commit from this time *last year* which I evidently never submitted a PR for and has been inadvertently included here. If you don't want those changes (adding GetFileSystemEntries, targetting .net4.0 Client) feel free to reject this PR and I'll submit a new one with just the Enumerate stuff :)


tathamoddie:
Rebased the latter commit as eca6e7608de571de4586ce4b65ceee1552233c94
Pushed to http://www.nuget.org/packages/System.IO.Abstractions/2.0.0.114

PR #126, opened 2015-04-25
Implemented EnumerateFiles methods

chkpnt:
The search pattern respects Microsofts [notes](https://msdn.microsoft.com/en-us/library/ms143327%28v=vs.110%29.aspx).

Initially, I also wanted to redirect the existing `MockDirectoryInfo#GetFiles(...)` methods to the new `EnumerateFiles(...)` methods. But as I don't fully understand the currently used 
[`GetFilesInternal`](https://github.com/tathamoddie/System.IO.Abstractions/blob/71c432a1b06be7153532293799012a68b038bf45/TestingHelpers/MockDirectory.cs#L171-L198) method, I'm first waiting for your feedback.

### DIFF
--- a/TestHelpers.Tests/MockDirectoryInfoTests.cs
+++ b/TestHelpers.Tests/MockDirectoryInfoTests.cs
@@ -111,33 +111,62 @@ namespace System.IO.Abstractions.TestingHelpers.Tests
         [Test]
         public void MockDirectoryInfo_GetParent_ShouldReturnDirectoriesAndNamesWithSearchPattern()
         {
-            // Arrange
             var fileSystem = new MockFileSystem();
             fileSystem.AddDirectory(XFS.Path(@"c:\a\b\c"));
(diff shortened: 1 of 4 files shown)


chaitanyagurrapu:
+1. This is something I ran into in my current project. Great work getting it fixed. Hope it makes it into a new release soon. We could really use it on our project.

tathamoddie:
I can see that you're fixing some bugs here, which is great, but you're duplicating the logic in https://github.com/tathamoddie/System.IO.Abstractions/blob/master/TestingHelpers/MockDirectory.cs#L171

We should fix that logic instead of introducing new logic.

Note that what's there currently is Unix-compatible, which is why I don't just want to drop yours straight over the top.

I think you need to split the Unix and Windows code paths, and then swap your logic in for the Windows front.

chkpnt:
You are right, I'm going to file a new pull request for the search-pattern-issue based on the current master-branch, which has the EnumerateFiles-methods already implemented.

PR #128, opened 2015-05-19
MockFileSystem.FileExists should not check directories

tillig:
Currently you can call `FileExists` with a directory that might exist in the mock file system and the method will return `true`, which is incorrect behavior. This fix updates the mock file system so it will only say a file exists if it's _actually a file_.


tillig:
Whoops, I think there's a better way to do this. I'll try again.

PR #130, opened 2015-05-24
Fixing interpretation of search patterns

chkpnt:
The `searchPattern` passed on to `GetFiles(...)` should be processed as Microsoft [describes](https://msdn.microsoft.com/de-de/library/wz42302f%28v=vs.110%29.aspx):

> When using the asterisk wildcard character in a searchPattern, such as "_.txt", the matching behavior when the extension is exactly three characters long is different than when the extension is more or less than three characters long. A searchPattern with a file extension of exactly three characters returns files having an extension of three or more characters, where the first three characters match the file extension specified in the searchPattern. A searchPattern with a file extension of one, two, or more than three characters returns only files having extensions of exactly that length that match the file extension specified in the searchPattern. When using the question mark wildcard character, this method returns only files that match the specified file extension. For example, given two files, "file1.txt" and "file1.txtother", in a directory, a search pattern of "file?.txt" returns just the first file, while a search pattern of "file_.txt" returns both files. 

In the current implementation of [`GetFilesInternal(...)`](https://github.com/tathamoddie/System.IO.Abstractions/blob/71c432a1b06be7153532293799012a68b038bf45/TestingHelpers/MockDirectory.cs#L171-L198), there are various regular expressions concerning invalid path chars. I think those are not neccessary:
1. It shouldn't be able to add Files and Directories with invalid chars to the _MockFileSystem_.
2. An exception should be thrown if a path containing invalid paths is passed on to `GetFiles(...)`.

For 2), I've filed an independent pull request (#131).

As the special searchPattern-behaviour for three-letter-extensions is not used by `GetDirectories(...)` I had to introduce a new method `GetDirectoriesInternal(...)`.

### DIFF
--- a/TestHelpers.Tests/MockDirectoryTests.cs
+++ b/TestHelpers.Tests/MockDirectoryTests.cs
@@ -16,12 +16,14 @@ namespace System.IO.Abstractions.TestingHelpers.Tests
             var fileSystem = SetupFileSystem();
             var expected = new[]
             {
-                XFS.Path(@"c:\a\a.txt"),
-                XFS.Path(@"c:\a\b.gif"),
-                XFS.Path(@"c:\a\c.txt"),
-                XFS.Path(@"c:\a\a\a.txt"),
-                XFS.Path(@"c:\a\a\b.txt"),
-                XFS.Path(@"c:\a\a\c.gif")
+                XFS.Path(@"c:\a\testA.so"),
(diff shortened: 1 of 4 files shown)


fgreinacher:
@chkpnt Definitely not your fault, but this branch is heavily outdated, so I'm gonna close the PR. If you're still willing to work on this feel free to re-open after it is rebased.

PR #131, opened 2015-05-24
Reworking throwing exceptions in MockDirectory

chkpnt:
Some refactorings. Also there were some checks missing in `EnumerateDirectories(...)` and `GetFiles(...)`.


jpreese:
I'm going to close this out. I think our exception throwing story is pretty Okay, and believe all of the checks are there. e.g. The searchPattern null check is slightly redundant because we're doing that check inside of CheckSearchPattern()

PR #138, opened 2015-09-22
Added TestingHelper for directory SetAttributes

kmcginnes:
I've implemented the logic for SetAttributes for directories. The code for files was already in place. And someone else had already provided the code to get attributes from directory paths. I just _borrowed_ and adapted that code.

### DIFF
--- a/TestHelpers.Tests/MockFileSetAttributesTests.cs
+++ b/TestHelpers.Tests/MockFileSetAttributesTests.cs
@@ -1,0 +1,66 @@
+﻿using System.Collections.Generic;
+using NUnit.Framework;
+using XFS = System.IO.Abstractions.TestingHelpers.MockUnixSupport;
+
+namespace System.IO.Abstractions.TestingHelpers.Tests
+{
+    [TestFixture]
+    public class MockFileSetAttributesTests
+    {
+        [Test]
+        public void MockFile_SetAttributes_ShouldSetAttributesOnFile()
+        {
+            var path = XFS.Path(@"c:\something\demo.txt");
+            var filedata = new MockFileData("test");
+            var fileSystem = new MockFileSystem(new Dictionary<string, MockFileData>
+            {
+                {path, filedata},
+            });
+
+            fileSystem.File.SetAttributes(path, FileAttributes.Hidden);
+
+            var attributes = fileSystem.File.GetAttributes(path);
+            Assert.That(attributes, Is.EqualTo(FileAttributes.Hidden));
+        }
+
+        [Test]
+        public void MockFile_SetAttributes_ShouldSetAttributesOnDirectory()
+        {
+            var fileSystem = new MockFileSystem();
+            var path = XFS.Path(@"c:\something");
+            fileSystem.AddDirectory(path);
+            var directoryInfo = fileSystem.DirectoryInfo.FromDirectoryName(path);
+            directoryInfo.Attributes = FileAttributes.Directory | FileAttributes.Normal;
+
+            fileSystem.File.SetAttributes(path, FileAttributes.Directory | FileAttributes.Hidden);
+
+            var attributes = fileSystem.File.GetAttributes(path);
+            Assert.That(attributes, Is.EqualTo(FileAttributes.Directory | FileAttributes.Hidden));
+        }
+
+        [Test]
+        [TestCase("", FileAttributes.Normal)]
+        [TestCase("   ", FileAttributes.Normal)]
+        public void MockFile_SetAttributes_ShouldThrowArgumentExceptionIfPathContainsOnlyWhitespaces(string path, FileAttributes attributes)
+        {
+            var fileSystem = new MockFileSystem();
+
+            TestDelegate action = () => fileSystem.File.SetAttributes(path, attributes);
+
+            Assert.Throws<ArgumentException>(action);
+        }
+
+        [Test]
+        public void MockFile_SetAttributes_ShouldThrowFileNotFoundExceptionIfMissingDirectory()
+        {
+            var path = XFS.Path(@"C:\something");
+            var attributes = FileAttributes.Normal;
+            var fileSystem = new MockFileSystem();
+
+            TestDelegate action = () => fileSystem.File.SetAttributes(path, attributes);
+
+            var exception = Assert.Throws<FileNotFoundException>(action);
+            Assert.That(exception.FileName, Is.EqualTo(path));
+        }
+    }
+}
--- a/TestingHelpers/MockFile.cs
+++ b/TestingHelpers/MockFile.cs
@@ -571,7 +571,23 @@ namespace System.IO.Abstractions.TestingHelpers
         {
             mockFileDataAccessor.PathVerifier.IsLegalAbsoluteOrRelative(path, "path");
 
-            mockFileDataAccessor.GetFile(path).Attributes = fileAttributes;
+            var possibleFileData = mockFileDataAccessor.GetFile(path);
+            if (possibleFileData == null)
+            {
+                var directoryInfo = mockFileDataAccessor.DirectoryInfo.FromDirectoryName(path);
+                if (directoryInfo.Exists)
+                {
+                    directoryInfo.Attributes = fileAttributes;
+                }
+                else
+                {
+                    throw new FileNotFoundException(string.Format(CultureInfo.InvariantCulture, StringResources.Manager.GetString("COULD_NOT_FIND_FILE_EXCEPTION"), path), path);
+                }
+            }
+            else
+            {
+                possibleFileData.Attributes = fileAttributes;
+            }
         }
 
         public override void SetCreationTime(string path, DateTime creationTime)
--- a/TestingHelpers/MockFileSystem.cs
+++ b/TestingHelpers/MockFileSystem.cs
@@ -64,6 +64,11 @@ namespace System.IO.Abstractions.TestingHelpers
 
         private string FixPath(string path, bool checkCaps = false)
         {
+            if (path == null)
+            {
+                throw new ArgumentNullException(nameof(path), StringResources.Manager.GetString("VALUE_CANNOT_BE_NULL"));
+            }
+            
             var pathSeparatorFixed = path.Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar);
             var fullPath = Path.GetFullPath(pathSeparatorFixed);
 


manne:
@kmcginnes the implementation looks good. Can you add the parameter checks (et al) based on the [documentation](https://msdn.microsoft.com/en-us/library/system.io.file.setattributes(v=vs.100).aspx) and the actual implementation.
it would be awesome, if you could also adapt the parameter check implementation of `MockFile.GetAttributes`.

kmcginnes:
I updated the code to the latest and fixed conflicts. I also added some tests that prove out some of the parameter preconditions. Let me know what else you want.

TestHelpers.Tests/MockFileTests.cs
```diff
@@ -269,6 +269,73 @@ public void MockFile_GetAttributeWithIllegalParameter_ShouldThrowOneArgumentExce
...
+            var attributes = FileAttributes.Normal;
+            var fileSystem = new MockFileSystem();
+            
+            TestDelegate action = () => fileSystem.File.SetAttributes(path, attributes);
+
+            Assert.Throws<PathTooLongException>(action);
+        }
+
+        [Test]
+        public void MockFile_SetAttributeOfMissingDirectory_ShouldThrowPathTooLongException()
```
manne: The name of the method does not match the thrown exception

TestingHelpers/MockFile.cs
```diff
@@ -410,7 +410,23 @@ public override void SetAccessControl(string path, FileSecurity fileSecurity)
...
+            if (possibleFileData == null)
+            {
+                var directoryInfo = mockFileDataAccessor.DirectoryInfo.FromDirectoryName(path);
+                if (directoryInfo.Exists)
+                {
+                    directoryInfo.Attributes = fileAttributes;
+                }
+                else
+                {
+                    throw new DirectoryNotFoundException();
```
manne: according to my local "manual" tests it should throw a `FileNotFoundException` (à la `throw new FileNotFoundException(string.Format(CultureInfo.InvariantCulture, "Could not find file '{0}'.", path), path);` . see https://github.com/tathamoddie/System.IO.Abstractions/blob/master/TestingHelpers/MockFile.cs#L201

TestHelpers.Tests/MockFileTests.cs
```diff
@@ -269,6 +269,73 @@ public void MockFile_GetAttributeWithIllegalParameter_ShouldThrowOneArgumentExce
...
+        }
+
+        [Test]
+        [TestCase("", FileAttributes.Normal)]
+        [TestCase("   ", FileAttributes.Normal)]
+        [TestCase(">", FileAttributes.Normal)]
+        public void MockFile_SetAttributeOfExistingDirectoryWithInvalidParameters_ShouldThrowArgumentException(string path, FileAttributes attributes)
+        {
+            var fileSystem = new MockFileSystem();
+
```
manne: can you remove the blanks

TestHelpers.Tests/MockFileTests.cs
```diff
@@ -269,6 +269,73 @@ public void MockFile_GetAttributeWithIllegalParameter_ShouldThrowOneArgumentExce
...
+            Assert.Throws<ArgumentException>(action);
+        }
+
+        [Test]
+        public void MockFile_SetAttributeOfExistingDirectoryWithPathTooLong_ShouldThrowPathTooLongException()
+        {
+            var path = "C:\\8e95d05bc8d747d194e59516f8d1493b\\aa3a4391679642f9bf83f9501f7fb0e1\\8c51cab5b63543909fb7da59a3932b47\\49910ca19d0a4b59aadd4276829cc476\\3b0540cbd57b4770a52166c0ffd72e10\\70ab25de331e473a9cffbc5e297c79c1\\607c63a6f57645aa98073e9df7c7edd8\\dc473e93228c44c08d127c36d9cd9b01";
+            var attributes = FileAttributes.Normal;
+            var fileSystem = new MockFileSystem();
+
```
manne: can you remove the blanks

TestHelpers.Tests/MockFileTests.cs
```diff
@@ -269,6 +269,73 @@ public void MockFile_GetAttributeWithIllegalParameter_ShouldThrowOneArgumentExce
...
+            Assert.Throws<PathTooLongException>(action);
+        }
+
+        [Test]
+        public void MockFile_SetAttributeOfMissingDirectory_ShouldThrowPathTooLongException()
+        {
+            var path = "C:\\Some\\Path\\That\\Does\\Not\\Exist";
+            var attributes = FileAttributes.Normal;
+            var fileSystem = new MockFileSystem();
+
```
manne: can you remove the blanks

manne:
there is a useful vs extension to recognize the trailing whitespaces
* [GitHub Repo](https://github.com/madskristensen/TrailingWhitespace)
* [VS Gallery](https://visualstudiogallery.msdn.microsoft.com/a204e29b-1778-4dae-affd-209bea658a59)

it really helps a lot. I deactivated removing the [trailing whitespaces on saving](https://github.com/madskristensen/TrailingWhitespace#remove-on-save).

TestHelpers.Tests/MockFileTests.cs
kmcginnes: If you look at the possible exceptions for SetAttributes it explicitly lists both empty string and strings with only whitespace. So I'm testing both scenarios.

TestHelpers.Tests/MockFileTests.cs
kmcginnes: What blanks?

TestHelpers.Tests/MockFileTests.cs
kmcginnes: Good catch. Copy pasta...

kmcginnes:
I've made the changes requested. I would also like to move the SetAttributes tests into their own file. Although, that might make more sense as a separate pull request. What do you think?

jpreese:
Thanks for this PR @kmcginnes. We're making progress on getting through the open PRs and issues that have been sitting around for awhile.. (ok a LONG while). This one looks like it's in pretty good shape, and you still seem pretty active. If you have interest in resolving the merge conflicts, we can re-run the checks and get this merged in for you.

kmcginnes:
I'll take a look. I'll have to reacquaint myself with these changes.

@jpreese I think it would be a good idea if I squash these commits down to one. I don't see a reason to keep the multiple commits around.

jpreese:
I agree. When it's time to merge into master, I'll be squashing them.

kmcginnes:
So, I take it you want me to merge `master`, not rebase `master`.

jpreese:
For the PR I'm indifferent. I was just going to squash before approving.

kmcginnes:
A couple of questions:

* Should I create a new file for the SetAttribute tests called `MockFileSetAttributesTests.cs` and move those tests out of `MockFileTests.cs`?
* Should I remove the `PathTooLongExceptoin` test? It's failing and I don't see any others in the code base. Maybe because Microsoft fixed it at some point?

```c#
[Test]
public void MockFile_SetAttributeOfExistingDirectoryWithPathTooLong_ShouldThrowPathTooLongException()
{
    var path = XFS.Path("C:\\8e95d05bc8d747d194e59516f8d1493b\\aa3a4391679642f9bf83f9501f7fb0e1\\8c51cab5b63543909fb7da59a3932b47\\49910ca19d0a4b59aadd4276829cc476\\3b0540cbd57b4770a52166c0ffd72e10\\70ab25de331e473a9cffbc5e297c79c1\\607c63a6f57645aa98073e9df7c7edd8\\dc473e93228c44c08d127c36d9cd9b01");
    var attributes = FileAttributes.Normal;

    TestDelegate action = () => fileSystem.File.SetAttributes(path, attributes);

    Assert.Throws<PathTooLongException>(action);
}
```

TestHelpers.Tests/MockFileTests.cs
```diff
@@ -290,6 +290,73 @@ public void MockFile_GetAttributeWithIllegalParameter_ShouldThrowOneArgumentExce
...
+            TestDelegate action = () => fileSystem.File.SetAttributes(path, attributes);
+
+            Assert.Throws<ArgumentException>(action);
+        }
+
+        [Test]
+        public void MockFile_SetAttributeOfExistingDirectoryWithPathTooLong_ShouldThrowPathTooLongException()
+        {
+            var path = XFS.Path("C:\\8e95d05bc8d747d194e59516f8d1493b\\aa3a4391679642f9bf83f9501f7fb0e1\\8c51cab5b63543909fb7da59a3932b47\\49910ca19d0a4b59aadd4276829cc476\\3b0540cbd57b4770a52166c0ffd72e10\\70ab25de331e473a9cffbc5e297c79c1\\607c63a6f57645aa98073e9df7c7edd8\\dc473e93228c44c08d127c36d9cd9b01");
```
jpreese: Instead of this approach, I think calling something like string.PadRight(<some max number here>) would be a little nicer.

TestingHelpers/MockFile.cs
```diff
@@ -527,7 +527,23 @@ public override void SetAttributes(string path, FileAttributes fileAttributes)
...
+            if (possibleFileData == null)
+            {
+                var directoryInfo = mockFileDataAccessor.DirectoryInfo.FromDirectoryName(path);
+                if (directoryInfo.Exists)
+                {
+                    directoryInfo.Attributes = fileAttributes;
+                }
+                else
+                {
+                    throw new FileNotFoundException($"Could not find file '{path}'.", path);
```
jpreese: I would take this opportunity to move this string into the Resources.resx file. Just make sure it's the same verbiage that System.IO uses, which it already looks like it is.

TestHelpers.Tests/MockFileTests.cs
```diff
@@ -290,6 +290,73 @@ public void MockFile_GetAttributeWithIllegalParameter_ShouldThrowOneArgumentExce
...
+
+            TestDelegate action = () => fileSystem.File.SetAttributes(path, attributes);
+
+            Assert.Throws<PathTooLongException>(action);
+        }
+
+        [Test]
+        public void MockFile_SetAttributeOfMissingDirectory_ShouldThrowFileNotFoundException()
+        {
+            var path = XFS.Path("C:\\Some\\Path\\That\\Does\\Not\\Exist");
```
jpreese: While I like the intent that it shows, we can probably dramatically reduce the length of this test data to simply C:\ or something along those lines. Since we never add anything to the MockFileSystem, any path is going to fail since no paths exist.

It also reduces the chance of the test failing because it depends on two artfully crafted strings (one in the arrange, and one in the assert)

jpreese:
1. Yes. I do like that approach better if you want to split it out.
2. It -should- pass, but unfortunately it's a known bug in MockFileInfo. You can see the issue here #192

TestingHelpers/MockFile.cs
jpreese: As a followup, I'm about to deploy this resource verbiage as part of another PR.

jpreese: Ok, you should be able to use StringResources.Manager.GetString("COULD_NOT_FIND_FILE_EXCEPTION")

kmcginnes: I see what I'm supposed to be able to do, but the compiler keeps telling me:

```
'StringResources' is inaccessible due to its protection level
```

And I do see that you have this defined in the `AssemblyInfo.cs`:

```c#
using System.Runtime.CompilerServices;

#if DEBUG
[assembly: InternalsVisibleTo("System.IO.Abstractions.TestingHelpers.Tests")]
#else
    [assembly: InternalsVisibleTo("System.IO.Abstractions.TestingHelpers.Tests, PublicKey=002400000480000094000000060200000024000052534131000400000100010051bf2aa00ba30d507d4cebcab1751dfa13768a6f5235ce52da572260e33a11f52b87707f858fe4bbe32cd51830a8dd73245f688902707fa797c07205ff9b5212f93760d52f6d13022a286ff7daa13a0cd9eb958e888fcd7d9ed1f7cf76b19a5391835a7b633418a5f584d10925d76810f782f6b814cc34a2326b438abdc3b5bd")]
#endif
```

So, in theory, it should work. I even tried shortening the assembly name to be `TestingHelpers.Tests`, which I believe is actually what it should be, but that didn't seem to help.

It looks like I'm the first consumer of `StringResources` in the test project. That's why I'm the first to see this.

😭

kmcginnes:
What do you think I should do about the `PathTooLongException`?

1. Comment out the test
2. Comment out the assertion
3. Mark it inconclusive
4. Delete the test
5. Add the workaround from #193 to the `SetAttributes()` method
6. Something else

TestingHelpers/MockFile.cs
jpreese: I'm not sure I follow. You're able to use it in the TestingHelpers library right? For MockInfo.cs? I wouldn't bother using it as an Assertion within the test. Just checking that it throws a FileNotFoundException is more than enough.

jpreese:
My gut says 4. You're not introducing the logic that checks the length. That's probably going to be the job of GetFullPath (or maybe even elsewhere, definitely not SetAttributes) sometime in the future. 

I think the check for a long path should be tied with the class/method that's going to do the check. Yeah?

kmcginnes:
Ok, so I've updated with the following:

* Updated with latest from `master`
* Use `StringResources` for the exception message, modeling it after some other thrown exceptions I see in the file
* Removed the test for long paths, as I agree that it's not `SetAttributes` responsibility
* Removed a test scenario for an invalid character that doesn't align with similar tests. So now it only tests for the whitespace scenarios
* Updated the names of the tests to better match the test naming convention in the project.
* Squashed all commits

TestHelpers.Tests/MockFileSetAttributesTests.cs
```diff
@@ -0,0 +1,67 @@
...
+using NUnit.Framework;
+
+namespace System.IO.Abstractions.TestingHelpers.Tests
+{
+    using XFS = MockUnixSupport;
+
+    [TestFixture]
+    public class MockFileSetAttributesTests
+    {
+
```
jpreese: Itty bitty whitespace

TestHelpers.Tests/MockFileSetAttributesTests.cs
```diff
@@ -0,0 +1,67 @@
+﻿using System.Collections.Generic;
+using NUnit.Framework;
+
+namespace System.IO.Abstractions.TestingHelpers.Tests
+{
+    using XFS = MockUnixSupport;
```
jpreese: I'd prefer if this was moved up with the other usings, for consistency.

TestHelpers.Tests/MockFileSetAttributesTests.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        [Test]
+        public void MockFile_SetAttributes_ShouldReplaceExistingFlagsIfExistingDirectory()
+        {
+            var fileSystem = new MockFileSystem();
+            var path = XFS.Path(@"c:\something");
+            fileSystem.AddDirectory(path);
+            var existingAttributes = fileSystem.File.GetAttributes(path);
+
+            fileSystem.File.SetAttributes(path, existingAttributes | FileAttributes.Hidden);
```
jpreese: The file case makes perfect sense to me. It starts out Normal, and we set the attribute to Hidden. Thus the assertion should be to make sure that the net result is that the file is Hidden (replaced the attribute).

I'm not quite certain why the directory test is pulling the existing attributes, and the File test did not?

jpreese:
Couple small nits and just a general question, otherwise looks great!

TestHelpers.Tests/MockFileSetAttributesTests.cs
kmcginnes: I prefer the same. I only put it here for consistency with other test files:

https://github.com/System-IO-Abstractions/System.IO.Abstractions/blob/0caa9382a68061a62c69ed3ef8f6f8204204ea7e/TestHelpers.Tests/MockDirectoryTests.cs#L8

TestHelpers.Tests/MockFileSetAttributesTests.cs
kmcginnes: The explanation is that I wrote the directory ones, and the file one was simply moved from the `MockFileTests.cs` file and into `MockFileSetAttributesTests.cs` to live along side my new tests.

But I think I agree with you. The file and directory tests should be consistent. And the MSDN article on `SetAttributes()` shows you should always add or remove from the existing set of attributes:

https://msdn.microsoft.com/en-us/library/system.io.file.setattributes(v=vs.110).aspx

So my next commit will update the file test with these changes.

jpreese: Ok. Let me grab your branch real quick and mess around with it in my IDE. I still have a couple concerns about how these tests are laid out, but I could be taking crazy pills.

jpreese: Ok, just went through the debugger and confirmed. So, I think how you had it first was correct, given functionality, test name, etc.

Right now you're saying "ShouldREPLACEExistingFlags...", but really what we're doing is ADDing more flags. So the end result is a file that is both "Normal" and "Hidden".

I think that SetAttributes() can be used sort of like a pseudo-RemoveAttribute. i.e. if you have a file thats Hidden -- to remove the hidden and make it normal, I'd think SetAttributes(FileAttributes.Normal) would work. And that would be "replacing".

If you want to preserve all/some of your attributes, then it probably makes sense to selectively remove and then add onto. But if you want to remove all but "Normal" or whatever, I don't think you need to -always- get the current attributes.

Hopefully that wasnt too much of a ramble. The TLDR is that I think it's possible we have two test case scenarios here "replace" and "add additional".

kmcginnes: I see your point on the file test.

But for directories, they always need to have the `FileAttributes.Directory` flag. So a full replacement isn't really possible.

How about initializing the directory's attributes to something in addition to the `FileAttributes.Directory` flag. Then we can replace it with something else.

Like this:

```c#
[Test]
public void MockFile_SetAttributes_ShouldReplaceExistingFlagsIfExistingDirectory()
{
    var path = XFS.Path(@"c:\something");
    fileSystem.AddDirectory(path);
    var directoryInfo = fileSystem.DirectoryInfo.FromDirectoryName(path);
    directoryInfo.Attributes = FileAttributes.Directory | FileAttributes.Normal;

    fileSystem.File.SetAttributes(path, FileAttributes.Directory | FileAttributes.Hidden);

    var attributes = fileSystem.File.GetAttributes(path);
    Assert.That(attributes, Is.EqualTo(FileAttributes.Directory | FileAttributes.Hidden));
}
```

jpreese: Honestly, the more I think about it and familiarize myself with everything. I don't really see anything wrong with what you're tests are testing. 

I just think the naming is less than ideal and may be a source of confusion (at least for me). Maybe we could just change them to something along the lines of MockFile_SetAttributes_ShouldSetAttributesOnDirectory and MockFile_SetAttributes_ShouldSetAttributesOnFile. Simple and to the point. No worry about what does "replace" mean or are we "adding" more attributes.

I would also drop the..
```
Attributes = FileAttributes.Normal
```
initialization on the File test since that should be set by System.IO.Abstractions for you.

What do you think about that?

kmcginnes: Sorry, I'll hopefully have a chance to take a look at this again tomorrow.

Life, amiright 😞

jpreese: No worries, we've made you wait years, I'm sure we can wait a little bit. No sweat! 👍

fgreinacher:
Let’s get this in (finally :))

fgreinacher:
I’ll merge as soon as the conflicts are resolved!

TestingHelpers/MockFileSystem.cs
```diff
@@ -91,6 +91,11 @@ public PathVerifier PathVerifier
 
         private string FixPath(string path, bool checkCaps = false)
         {
+            if (path == null)
+            {
+                throw new ArgumentNullException("path", StringResources.Manager.GetString("VALUE_CANNOT_BE_NULL"));
```
fgreinacher: Nit: `nameof(path)`

fgreinacher:
@kmcginnes Could you update this branch one last time? I’d then squash-merge, so no worries about cleaning up the history :)

kmcginnes:
The branch has been updated from upstream, and I've squashed all my commits.

We should be good to go 🚀

fgreinacher:
Thanks again, merging!

PR #140, opened 2015-10-16
Throwing DirectoryNotFound exception when directory does not exist.

BrianLanghoor:
When updating legacy code so it can be tested with unit tests I ran into a small problem. When you, in the actual fileSystem from Microsoft, call the AddFile method. The only thing it does is add the file to the already existing directory. If that directory is not present it will give you an DirectoryNotFoundException. For the MockFileSystem I would expect the same behavior as below:

I was running a test on the below code and made sure that the directory did not exist in the mocked file system.

Test:

``` C#
    [TestMethod]
    public void WhenCallingDoSomethingWithoutExistingDirectoryShouldCreateDirectory()
    {
            var foo = new Foo();
            string path = @"C:\temp\createThisDirectory";

            foo.DoSomething(path);

            //Assert using fluent assertions.
            mockFileSystem.Directory.Exists(@"C:\temp").Should().BeTrue();
    }
```

Code:

``` C#
    public class Foo
    {
        private readonly IFileSystem fileSystem;

        public Foo(IFileSystem fileSystem)
        {
            this.fileSystem = fileSystem;
        }

        public void DoSomething(string path)
        {
            //..........some code here..........

            //This should not be TRUE because the c:\temp will always exist in the real filesystem.
            if (!Directory.Exists(Path.GetDirectoryName(path)))
            {
                Directory.CreateDirectory(Path.GetDirectoryName(path));
            }

            //Below line was already mocked and seeing the code aboveI would expect
            //the test to fail below because that would have happend if it was not
            //mocked using the real filesystem. The directory is not there yet.

            fileSystem.File.WriteAllBytes(path, aByteArray);

            //.........some more code here...........
         }
    }
```

It turned out that the line: 'fileSystem.File.WriteAllBytes(path, aByteArray);', dit not fail. After some research in the 'System.IO.Abstractions' turned out that the method WriteAllBytes() makes use of the AddFile() method. The AddFile() method looked like this:

``` C#
         public void AddFile(string path, MockFileData mockFile)
         {
             var fixedPath = FixPath(path);
             lock (files)
             {
                 if (FileExists(fixedPath))
                 {
                     var isReadOnly = (files[fixedPath].Attributes & FileAttributes.ReadOnly) == FileAttributes.ReadOnly;
                     var isHidden = (files[fixedPath].Attributes & FileAttributes.Hidden) == FileAttributes.Hidden;

                     if (isReadOnly || isHidden)
                     {
                         throw new UnauthorizedAccessException(string.Format(CultureInfo.InvariantCulture, "Access to the path '{0}' is denied.", path));
                     }
                 }

                 var directoryPath = Path.GetDirectoryName(fixedPath);

                 if (!directory.Exists(directoryPath))
                 {
                      AddDirectory(directoryPath);
                 }

                 files[fixedPath] = mockFile;
             }
         }
```

The reason it dit not fail is obviously:

``` C#
                 if (!directory.Exists(directoryPath))
                 {
                      AddDirectory(directoryPath);
                 }
```

A quick fix for this was the following:

``` C#
                 if (!directory.Exists(directoryPath))
                 {
                      throw new DirectoryNotFoundException();
                 }
```

In order to make all the tests work I created the following function and replaced every occurance of AddFile() with AddFileWithCreate() in the tests. The difference here is that with the first a directory has to be present like the actual filesystem and the second creates the directory if it's not there.

``` C#
        public void AddFileWithCreate(string path, MockFileData value)
        {
            var fixedPath = FixPath(path);

            var directoryName = Path.GetDirectoryName(fixedPath);
            if (!directory.Exists(directoryName))
            {
                directory.CreateDirectory(directoryName);
            }

            AddFile(fixedPath, value);
        }
```

If in the test there was another function used like WriteAllBytes() I created the directory in the test using:

``` C#
    fileSystem.Directory.CreateDirectory(Path.GetDirectoryName(fullPath));
```

In my opinion this makes system.IO.Abstractions better suitable for a smooth simulation of the actual fileSystem. Please let me know what you think about this. 

Kind regards,
Brian Langhoor

### DIFF
--- a/TestHelpers.Tests/MockDirectoryTests.cs
+++ b/TestHelpers.Tests/MockDirectoryTests.cs
@@ -368,7 +368,7 @@ namespace System.IO.Abstractions.TestingHelpers.Tests
         {
             // Arrange
             var fileSystem = new MockFileSystem();
-            fileSystem.AddFile(XFS.Path(@"c:\foo\bar.txt"), new MockFileData("Demo text content"));
+            fileSystem.AddFileWithCreate(XFS.Path(@"c:\foo\bar.txt"), new MockFileData("Demo text content"));
 
             // Act
             var result = fileSystem.Directory.Exists(XFS.Path(@"c:\foo\"));
--- a/TestHelpers.Tests/MockFileCreateTests.cs
+++ b/TestHelpers.Tests/MockFileCreateTests.cs
@@ -21,6 +21,8 @@ namespace System.IO.Abstractions.TestingHelpers.Tests
 
             Assert.That(fileSystem.FileExists(fullPath), Is.False);
 
+            fileSystem.Directory.CreateDirectory(Path.GetDirectoryName(fullPath));
+
             sut.Create(fullPath).Close();
 
             Assert.That(fileSystem.FileExists(fullPath), Is.True);
@@ -34,6 +36,7 @@ namespace System.IO.Abstractions.TestingHelpers.Tests
             var data = new UTF8Encoding(false).GetBytes("Test string");
 
             var sut = new MockFile(fileSystem);
+            fileSystem.Directory.CreateDirectory(Path.GetDirectoryName(fullPath));
             using (var stream = sut.Create(fullPath))
             {
                 stream.Write(data, 0, data.Length);
@@ -52,6 +55,8 @@ namespace System.IO.Abstractions.TestingHelpers.Tests
             var fileSystem = new MockFileSystem();
 
             var mockFile = new MockFile(fileSystem);
+
+            fileSystem.Directory.CreateDirectory(Path.GetDirectoryName(path));
 
             // Create a file
             using (var stream = mockFile.Create(path))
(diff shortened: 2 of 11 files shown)


manne:
yes, the behavior of the `MockFileSystem` is not correct.
This should fail (pre condition: there is not a directory with the path "C:\tmp\sioa\140\123123", directory "C:\tmp\sioa\140" can exist):
```c#
var fileSystem = new MockFileSystem(); // or new FileSystem()
fileSystem.File.WriteAllText(XFS.Path(@"C:\tmp\sioa\140\123123\demo.txt"), "Hello");
```

What is your intension of the additional method `MockFileSystem.AddFileWithCreate`?

From my point of view there is no need for this method.
The check that the directory does not exist, should be done inside `MockFile.WriteAllText(...)`.
It is in the responsibility of each test, to bring the `MockFileSystem` in a suitable state.
So `MockFileSystem.AddFile(...)` always adds the file, even if there is not a directory for the file.
For example this should always add the file:
```c#
fileSystem.AddFile(XFS.Path(@"c:\something\demo.txt"), new MockFileData("Hello"));
```

jpreese:
Note, this is a pretty important PR to mimic System.IO behavior.

jpreese:
After looking at this again. WriteAllBytes and WriteAllText now have this check as well as some Stream creation. We'll take up the other methods as they appear if any exist.

PR #159, opened 2016-03-28
Add support for framework v3.5 and v2.0.

PhilipDaniels:
While I love System.IO.Abstractions for the boost it gives to testability, I am in the unfortunate position of having to support some libraries that need to be built for older framework versions. This makes it impossible to incorporate System.IO.Abstractions into them at the moment.

This pull request adds support for 3.5 and 2.0 (for which LinqBridge is also needed). All public APIs are preserved and have the same semantics, but the performance and memory usage characteristics might be somewhat different, because the technique used is to delegate the Enumerate\* methods (which don't exist in older frameworks) back to the Get\* methods, which do.

I updated the nuspec to incorporate all 3 versions, but it should really be changed to use the Release builds rather than the Debug ones, but I don't see to be able to build in Release mode, something to do with the key.

### DIFF
--- a/System.IO.Abstractions.sln
+++ b/System.IO.Abstractions.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 2013
-VisualStudioVersion = 12.0.30723.0
+# Visual Studio 14
+VisualStudioVersion = 14.0.24720.0
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{5F3BDA62-8052-4C37-97A0-94354AA971B0}"
 	ProjectSection(SolutionItems) = preProject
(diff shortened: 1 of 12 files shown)


PhilipDaniels:
Hi, @tathamoddie do you have any feedback on this? It is a severe problem for us. It is a great boon if the authors of fundamental building-block libraries such as this try and make them available in as many versions as possible. Take a look at Newtonsoft.JSON for example:

json


I am more than happy to make adjustments to the code if you require it.
Any feedback would be appreciated.

SingleCopy:
Just to add, would it not be better to add a net35 project that could reference the majority of the files via a file link and add your own changes to the new project. No merge issues that way. 

If you look at at dapper, that has a lot of preprocessor directives in the code and makes it quite difficult to read and maintain.

fgreinacher:
Triage: For me this has become irrelevant and I would rather consider getting rid of the one full framework target that currently exists. .NET Standard has become the de factor standard for libs and is clearly the future.

PhilipDaniels:
Agreed. 2 years and while we still have some customers on 3.5 we have convinced the bosses we aren't going to bend over backwards for them anymore.

PR #157, opened 2016-02-09
Implement Creation, LastAccess and LastWrite dates on MockDirectoryInfo

johncmckim:
I implemented Creation, LastAccess and LastWrite times to resolve #156.

I've got different formatting settings in VisualStudio, but I tried to keep it consistent with what you use. I also added tests based on the tests in use for CreationTime on MockFileInfo.  Let me know if there are issues with the formatting or tests.

PR #160, opened 2016-04-17
Added support for CreationTime

TwanVanDongen:


jpreese:
Thanks for this PR! We're going to close this in favor of #157 as they both intend to achieve the same goal, but #157 also includes additional properties.

PR #164, opened 2016-07-16
added test helper support for FileInfoBase.Replace

dombrovsky:


ericnewton76:
Um, since when should a file system "ShouldCreateBackup".  That's not
normally a filesystem goal, except in very specific instances.

rkoeninger:
What about this PR needs to be changed to get it merged in?

According to the [docs for System.IO.File.Replace](https://msdn.microsoft.com/en-us/library/system.io.file.replace(v=vs.110).aspx), it looks like a backup is part of the functionality.

You can see [in the reference source](https://referencesource.microsoft.com/#mscorlib/system/io/file.cs,d4b8da02b41f19a4), it also mentions a backup path, although the actual operation is in native code.

I need this functionality to get my tests working again and would appreciate if it was in the next release.

jpreese:
@rkoeninger we're starting to get through all of the open PRs after some inactivity. 

Given that the original date on this PR is quite old, and the original author does not seem to be as active as you on GitHub -- if you would like that champion this enhancement, I'm OK with closing this out and reviewing yours if you still have the desire to do so.

jpreese:
Thanks! I'll close this one and take a look.

PR #174, opened 2016-10-13
Fix MockFileInfo reset own path on MoveTo to match functionality for …

ghendley:
…FileInfo with Windows filesystem (not sure whether this reflects functionality for Unix/Linux filesystem).


tathamoddie:
Thanks! This should hit NuGet as 2.0.0.137 in a few moments.

PR #179, opened 2016-11-17
Fixed MockFile to properly use DirectoryNotFoundException

eggenstein:
`MockFile` uses `AddFile()` which implicitly creates directories.
Added checks for existing directories to throw `DirectoryNotFoundException` accordingly.

Similar to [PullRequest#140](https://github.com/tathamoddie/System.IO.Abstractions/pull/140), but without changing the behavior of `AddFile()`.


MCCoolMoDAue:
this is actually a fix for the [Issue#151](https://github.com/tathamoddie/System.IO.Abstractions/issues/151)

Issue #170, opened 2016-09-29
Throw IOException on FileInfoBase.OpenRead()


dgee2:
Hi,

I am trying to engineer a situation for a test case where a file cannot be opened when calling OpenRead on the FileInfoBase object. Is it possible to force an exception when opening it from a MockFileSystem? I have looked into opening the file using OpenWrite and then trying to open it again using OpenRead but this didn't seem to work:

```
var fileSystem = new MockFileSystem(new Dictionary<string, MockFileData> {
    { @"C:\temp\tempFile.txt", MockFileData.NullObject }
};

var file = fileSystem.FileInfo.FromFileName(@"C:\temp\tempFile.txt");

using (var writeLock = fileSystem.File.OpenWrite(@"C:\temp\tempFile.txt")){
    file.OpenRead();
}
```

If anyone could think of another way to make OpenRead throw an IOException or if there is a way to mock the FileInfo that I am creating for file, it would be greatly appreciated.

Thanks

Mike-EEE:
I am looking for exactly the same thing.  Any luck with this?

AlexanderPruss:
Yeah, I need this as well. I'll add a fork for it hopefully sometime today.

AlexanderPruss:
Added a PR - https://github.com/tathamoddie/System.IO.Abstractions/pull/180

Mike-EEE:
Nice work, @AlexanderPruss!  Thank you for contributing to the community.  FWIW, I ended up using @kzu's excellent [Moq framework](https://github.com/moq/moq) to get the behavior that I was after. 😄

AlexanderPruss:
@Mike-EEE Does the Moq Framework allow mocking of MSCORLIB classes?

Mike-EEE:
@AlexanderPruss yes you can Moq any interface and any class as long as the member you are mocking is virtual/overridable.  That is, interfaces are completely 100% mockable, and overridable/virtual methods and members in classes are mockable.

AlexanderPruss:
Removed my PR, I'm going to merge the 7+ bug fixes I've made onto my master branch.

jpreese:
@Mike-EEE @AlexanderPruss @dgee2 -- this has been merged into master and is available as of the latest release.

PR #180, opened 2016-11-18
Issue #170: Opening a MockFileStream to read from a nonexistent file will throw an exception.

AlexanderPruss:
Implementation note - instead of using System.IO.FileAccess/FileMode, I
ended up creating an internal StreamType enum instead to keep the
interface simple. That's potentially something to change upon on review,
but I think it's good the way it is.

### DIFF
--- a/TestHelpers.Tests/MockFileStreamTests.cs
+++ b/TestHelpers.Tests/MockFileStreamTests.cs
@@ -15,7 +15,7 @@
             // Arrange
             var filepath = XFS.Path(@"c:\something\foo.txt");
             var filesystem = new MockFileSystem(new Dictionary<string, MockFileData>());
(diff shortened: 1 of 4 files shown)


AlexanderPruss:
This does change the functionality of the FileInfoBase.OpenRead(). If you'd prefer, I could make this new functionality optional. I do think it's important to be able to have opening nonexistent mocked files throw exceptions, as that's what the System.IO FileInfo does.

AlexanderPruss:
Looks like nothing is happening with this PR. I'm removing it, as I'm going to be using my master branch to hold lots of fixes, including one that will be potentially breaking (https://github.com/tathamoddie/System.IO.Abstractions/issues/189).  I'll keep my other PR's open for now, in case you want to use my fixes.

Issue #178, opened 2016-11-12
Assembly signed


ianselmi:
Hi, from version 2.0.0.124 the assembly signed is missing in nuget package assembly.

Can you resign the assembly? In my project my assembly are all signed and I can't installed new updates

Thanks

AlexanderPruss:
https://github.com/tathamoddie/System.IO.Abstractions/pull/182

MCCoolMoDAue:
The System.IO.Abstractions.dll in the package version 138 is still unsigned.
Is this supposed to be like that?
Maybe I am understanding wrong?

jpreese:
As of v2.1.x.x the package should be signed.

PR #182, opened 2016-11-21
#178: Resigning assemblies with an snk

AlexanderPruss:
@Master: This uses a .snk instead of a .pfx file.


flcdrg:
Will this produce the same public key as the old versions?

AlexanderPruss:
Good point. No, this would change the public key.

flcdrg:
So if someone knew the password for the pfx, then you could use `sn -p strongname.pfx System.IO.Abstractions.snk` to extract the snk file, and I think that would probably result in keeping the same public key.

@tathamoddie ?

tathamoddie:
I have the password for the PFX.

It used to be installed on the old TeamCity box, so the build process would pick it up.

When we moved to AppVeyor, I didn't have a solution for it.

I can easily unlock the pfx, or publish the password, but that feels wrong. I don't know why.

I'm not terribly familiar with AppVeyor yet, but I do know that they have a way of storing encrypted values. Can somebody help with a PR / pointers on how to securely use the current PFX?

flcdrg:
Yes, you can define encrypted things with AppVeyor. Maybe it would need to be set as an environment variable that the csproj could pick up - but you'd want to do that in a way that didn't end up including the password in the build logs!

flcdrg:
So looking at http://codecoding.com/using-appveyor-to-build-a-vsto-project/, the way they do it is:

- add an encrypted environment variable
- in a pre-build script, grab this value and use it to install the pfx in the user store

eg. 
```
	$psw = ConvertTo-SecureString $env:certificatePass -AsPlainText -Force
	Import-PfxCertificate -FilePath .\{Path-to-your-pfx-in-your-github-project}\Certificate.pfx -CertStoreLocation cert:\CurrentUser\My -Password $psw
```

jpreese:
This PR is quite old and I have not been able to get in touch with the original author. Going to close to explore different avenues with AppVeyor

PR #183, opened 2016-11-21
181 moving a file to itself will no longer throw an exception

AlexanderPruss:
See http://stackoverflow.com/questions/40722205/should-mscorlibs-fileinfo-moveto-throw-an-exception-if-trying-to-move-a-file


jpreese:
Due to activity, this will be handled in #299

PR #185, opened 2016-11-22
184 read only mocked files should not be deletable

AlexanderPruss:


jpreese:
To be handled in #303

PR #193, opened 2016-11-29
Issue 192 mockfileinfo wil throw exceptions for too long paths

AlexanderPruss:


MCCoolMoDAue:
you can create a long path with the string.Pad functions
string longPath = @"C:\".PadRight(260, 'x');
that way you know it is 260 char long and it doesn't reach outside the editor ;)

AlexanderPruss:
Neat. It's more poetic this way though :D

TestHelpers.Tests/MockDirectoryInfoTests.cs
```diff
@@ -264,5 +264,17 @@ public void MockDirectoryInfo_Constructor_ShouldThrowArgumentException_IfArgumen
             var exception = Assert.Throws<ArgumentException>(action);
             Assert.That(exception.Message, Is.StringStarting("The path is not of a legal form."));
         }
+
+        [Test]
+        [ExpectedException(typeof(PathTooLongException))]
```
manne: can you please use Assert.Throws<ArgumentException>() instead of [ExpectedException(typeof(PathTooLongException))]
see https://github.com/tathamoddie/System.IO.Abstractions/blob/0d3f11e508815321bbd1279a3889fe775723a39c/TestHelpers.Tests/MockFileWriteAllBytesTests.cs#L22

TestHelpers.Tests/MockFileInfoTests.cs
```diff
@@ -476,5 +476,18 @@ public void MockFileInfo_MoveTo_ShouldUpdateFileInfoDirectoryAndFullName()
             Assert.AreEqual(fileInfo.DirectoryName, destinationFolder);
             Assert.AreEqual(fileInfo.FullName, destination);
         }
+
+        [Test]
+        [ExpectedException(typeof(PathTooLongException))]
```
manne: see above

TestHelpers.Tests/MockFileStreamTests.cs
```diff
@@ -44,5 +44,19 @@ public void MockFileStream_Dispose_ShouldNotResurrectFile()
             Assert.AreEqual(0, fileCount2, "File should have been deleted");
             Assert.AreEqual(0, fileCount3, "Disposing stream should not have resurrected the file");
         }
+
+        [Test]
+        [ExpectedException(typeof(FileNotFoundException))]
```
manne: see above

jpreese:
I'm going to close this PR -- I think we should handle this explicitly rather than calling the base implementation. I'll fiddle with this a little bit.

PR #195, opened 2016-12-06
.NET Core version

dabide:
I managed to convert System.IO.Abstractions to .NET Core. Methods that are not part of that are disabled using `#if NET45`.

The tests were converted to XUnit, as there are compatibility issues with NUnit.

There is a bug in VS 2017 RC which prevents setting AssemblyVersion and so on, so that needs to be fixed.

All tests are green. Feel free to do whatever you please with it.


JT-Bruch:
Is there progress on this?

dabide:
See See https://github.com/tathamoddie/System.IO.Abstractions/pull/221

PR #204, opened 2017-03-01
Ensure attributes follow when moving a directory

mattdwen:
With the test helpers, when a directory is moved, any custom attributes
such as hidden or system are not moved to the destination directory.


tathamoddie:
Thanks! Should hit NuGet in the next 15 minutes or so.

(Great PR! Short, succinct, with clear description and tests!)

PR #219, opened 2017-04-04
Update MockDirectory.cs

brianestey:
Remove Exists check so Exception is not thrown if Directory already exists.


tathamoddie:
Oops! #211 already fixed this one, making this redundant. Sorry for getting you to do the excess work ☹️, but thanks for contributing! 😄

PR #220, opened 2017-04-04
Update MockDirectoryTests.cs

brianestey:
Add test case to ensure CreateDirectory will succeed even if directory already exists.


ericnewton76:
normally you'd do this as one pull request.  can you pull both of these back and submit a single pull request?

tathamoddie:
Covered by #219 and #211

PR #221, opened 2017-04-19
.NET Standard compatibility

dabide:
Added .NET Standard 1.4 compatibility.


dabide:
Not sure how to fix the AppVeyor configuration.

tathamoddie:
Deployed to NuGet as 2.1+, via #225. Some more discussion in #229.

Thanks for all this work, and sorry it took so long to land!

PR #231, opened 2017-07-17
Target .net standard (1.3, 1.6)

boekabart:
In order for us to keep using this library in the near future, it needs netstandard support. This is an attempt at making that happen.


boekabart:
Seems that only the TestingHelpers nuspec (what's that) is being picked up now, not the 'actual' one.

robichaud:
@tathamoddie please make this a priority.

jpreese:
@robichaud @boekabart there has been a lot of progress made towards .NET standard support. Is this still an issue for you?

Issue #233, opened 2017-07-23
MockFileSystem.File.Exists throws NullReferenceException


robertlarkins:
I'm trying to mock what would happen if I get a file from disk, store a `FileInfoBase` representation of it, delete the file from disk, try to move the representation, and see if the file exists in the new location. But I seem to be missing something in my understanding with the following code, as I would have expected that both `fileExists1` and `fileExists2` would operate the same and both return `false`.

```csharp
var filePath = @"C:\folder\file.txt";
var newFolder = @"C:\newFolder";

var fileSystem = new MockFileSystem(new Dictionary<string, MockFileData>
{
    { filePath, new MockFileData(new byte[5]) },
});

fileSystem.AddDirectory(newFolder);

var fileInfo = fileSystem.FileInfo.FromFileName(filePath);
var newDir = fileSystem.DirectoryInfo.FromDirectoryName(newFolder);

// file is removed
fileSystem.File.Delete(filePath);

// move file wrapper to new location
var newPath = Path.Combine(newDir.FullName, fileInfo.Name);
fileInfo.MoveTo(newPath);

// Check that the file still exists on disk before trying to do anything with it
var fileExists1 = fileSystem.FileExists(@"C:\newFolder\file.txt");
var fileExists2 = fileSystem.File.Exists(@"C:\newFolder\file.txt");
```

But what happens is that `fileExists1` returns true, and `fileExists2` throws a Null Reference Exception. Why does this happen, and what is the correct way of checking if the file exists?

ausarb:
The problem is actually with the fileInfo.MoveTo(newPath) line.  Since the underlying file was deleted, the MoveTo should have failed with a FileNotFoundException.  No exceptions were thrown and the underlying tracking code in MockFileSystem became corrupt.  

I've submiied PR https://github.com/tathamoddie/System.IO.Abstractions/pull/236 to fix this so that a FileNotFoundException is thrown.  In your case you will still need to change your code, but the real error will be clear.

ayb4btu:
Ah, that would explain it, thanks. Hopefully a merge will come soon.

ayb4btu:
On a side not, should the issue be closed now, or once the merge has occurred?

ausarb:
Thanks for reopening.  Let's leave it open until merged, since I've done the PR against this issue.  I had a similar problem as you (unexpected NullReferenceException) and found this ticket while invetigating.  My PR fixes both your issue and mine.

[...]

PR #236, opened 2017-08-07
MockFile.Copy now throws a FileNotFoundException when the source file doesn't exist.

ausarb:
Previously no exception was thrown and it appeared to work, but later requests to some MockFileSystem methods would then fail with NullReferenceException's.  

The fix mimics System.IO.File's Copy and Move methods.


jpreese:
Note, this is for #233

jpreese:
This was resolved with #299

Issue #239, opened 2017-09-05
Extend FileSystemInfoBase with FileSystem property


AlenPelin:
## Proposal:
Extend abstract `FileSystemInfoBase` class with `public IFileSystem FileSystem { get; }` property to simplify usages. 

## Example 1:

```c#
public class Program
{
  public static void Main()
  {
    var fileSystem = new FileSystem();
    var zipFile = fileSystem.FileInfo.FromFileName("C:\\1.zip");

    var extractedFiles = ExtractInnerZip(fileSystem, zipFile);

    Console.WriteLine("Files extracted: {0}", extractedFiles.GetFiles().Length);
  }

  public static DirectoryInfoBase ExtractInnerZip(IFileSystem fileSystem, FileInfoBase zipFile)
  {
    var tempDirectory1 = fileSystem.DirectoryInfo.FromDirectoryName(@"C:\temp1");
    ZipHelper.ExtractFile(zipFile, tempDirectory1);
    var tempDirectory2 = fileSystem.DirectoryInfo.FromDirectoryName(@"C:\temp2");
    ZipHelper.ExtractFile(tempDirectory1.GetFiles("Inner.zip").Single(), tempDirectory2);

    return tempDirectory2;
  }
}
```

In this example extra function argument can be removed:

```c#
public class Program
{
  public static void Main()
  {
    var zipFile = new FileSystem().FileInfo.FromFileName("C:\\1.zip");

    var extractedFiles = ExtractInnerZip(zipFile);

    Console.WriteLine("Files extracted: {0}", extractedFiles.GetFiles().Length);
  }

  public static DirectoryInfoBase ExtractInnerZip(FileInfoBase zipFile)
  {
    var fileSystem = zipFile.FileSystem;
    var tempDirectory1 = fileSystem.DirectoryInfo.FromDirectoryName(@"C:\temp1");
    ZipHelper.ExtractFile(zipFile, tempDirectory1);
    var tempDirectory2 = fileSystem.DirectoryInfo.FromDirectoryName(@"C:\temp2");
    ZipHelper.ExtractFile(tempDirectory1.GetFiles("Inner.zip").Single(), tempDirectory2);

    return tempDirectory2;
  }
}
```

## Example 2:
```c#
public class ExtractInnerZipCommand
{
  public IFileSystem FileSystem { get; set; }

  public FileInfoBase ZipFile { get; set; }

  public DirectoryInfoBase Execute()
  {
[...]

PR #242, opened 2017-09-06
Extend FileSystemInfoBase with FileSystem property

AlenPelin:
Please review proposal in issue #239 and this PR that implements it.


ericnewton76:
Storing a FileSystem instance in a static variable will cause many race conditions.  I don't recommend this PR.

There's a possibility instead of modifying the constructor for the Mocks to require the IFileSystem parent instance in order to return it back... this would then stave-off the race conditions present of a default implementation simply using an AppDomain static that the first test implementation ends up mangling upon completion.

herebebeasties:
#352 implements this for all base classes without statics. (Probably close this if that's accepted.)

fgreinacher:
Superseded by #352

PR #253, opened 2017-11-04
Move to MIT

tathamoddie:
As a more permissive and popular license than MS-PL


tathamoddie:
Paging @manne as the second most prolific contributor, and @colin-young and @jpreese as incoming maintainers

PR #261, opened 2017-11-16
Add support for Directory AccessControl

russcam:
Supercedes #259 (_meant to branch for the PR as opposed to creating on master_).

Add support for both getting and setting AccessControl on directories.

It looks like the implementation of `AddDirectory(string path)` on `MockFileSystem` should also accept a `MockDirectoryData` parameter, similar to how `AddFile(string path, MockFileData mockFile)` accepts `MockFileData`. This PR doesn't include that change

### DIFF
--- a/TestHelpers.Tests/MockDirectoryGetAccessControlTests.cs
+++ b/TestHelpers.Tests/MockDirectoryGetAccessControlTests.cs
@@ -1,0 +1,69 @@
+﻿using NUnit.Framework;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Security.AccessControl;
+
+namespace System.IO.Abstractions.TestingHelpers.Tests
+{
+    using XFS = MockUnixSupport;
+
+    [TestFixture]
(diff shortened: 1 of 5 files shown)


tathamoddie:
Thanks! Shipped to NuGet in 2.1.0.175

russcam:
Thanks!

russcam:
> It looks like the implementation of AddDirectory(string path) on MockFileSystem should also accept a MockDirectoryData parameter, similar to how AddFile(string path, MockFileData mockFile) accepts MockFileData. 

What are your thoughts on ^ @tathamoddie? Happy to submit a PR for this change if it makes sense (guess you'd bump the major if adhering to SemVer)?

tathamoddie:
Could it be added as an optional overload?

For `AddFile`, we always need the file data to know what's in it.

For `AddDirectory`, we'd only need a `MockDirectory` object if you want to init it with special access control restrictions. All existing test code could continue to default as-is.

russcam:
I can add it as an overload on the constructor?

PR #262, opened 2017-11-16
DirectoryInfo.Root.Name should return drive name

russcam:
```c#
new DirectoryInfo(@"C:\").Root.Name 
```
returns `C:\`, but `MockDirectoryInfo` will return empty string. 

The implementation should use the FullName when path is an empty string


russcam:
will submit again shortly (_branched from wrong commit, need more ☕️_)

Issue #264, opened 2017-11-16
MockDirectory.Enumerate* should search current directory if search path is not rooted


benagain:
Great library, I use it a lot!

I have found a difference in behaviour between `System.IO.Abstractions.TestingHelpers.MockDirectory` and  `System.IO.Directory` - the platform implementation will find files for a relative path under the current working directory; `MockDirectory` does not find those files.

Assuming there are some files in `c:\temp\subdir`, and the `System.IO.Directory`'s current path is `c:\temp`, then `System.IO.Directory.EnumerateFiles("subdir")` returns the list of those files.  Given the same set of files in `MockDirectory`, its `EnumerateFiles` returns an empty list.

```c#
[Test]
public void RealDirectory_EnumerateFiles_WhenFilterIsUnRooted_ShouldFindFilesInCurrentDirectory()
{
    System.IO.Directory.SetCurrentDirectory(@"c:\temp");
    var actual = System.IO.Directory.EnumerateFiles("subdir");
    Assert.That(actual, Is.Not.Empty);
}

[Test]
public void MockDirectory_EnumerateFiles_WhenFilterIsUnRooted_ShouldFindFilesInCurrentDirectory()
{
    // Arrange
    var fileSystem = new MockFileSystem(new Dictionary<string, MockFileData>
    {
        { XFS.Path(@"c:\a.txt"), MockFileData.NullObject },
        { XFS.Path(@"c:\a\a.txt"), MockFileData.NullObject },
        { XFS.Path(@"c:\a\b\b.txt"), MockFileData.NullObject },
        { XFS.Path(@"c:\a\c\c.txt"), MockFileData.NullObject },
    });

    var expected = new[]
    {
        XFS.Path(@"c:\a\b\b.txt"),
    };

    fileSystem.Directory.SetCurrentDirectory(XFS.Path(@"c:\a"));

    // Act
    var result = fileSystem.Directory.EnumerateFiles(XFS.Path("b"));

    // Assert
    Assert.That(result, Is.EquivalentTo(expected));
}
```

jpreese:
Thanks again! 🎉

PR #265, opened 2017-11-16
MockDirectory.Enumerate can search relative to current directory

benagain:
Fix for #264: `System.IO.Directory.Enumerate*` will find files matching the search path inside the current directory, when the search path is not rooted.  This commit makes `MockDirectory.Enumerate*` emulate that same behaviour.


jpreese:
Thanks for the PR @benagain !

PR #267, opened 2017-12-30
Added methods to handle loading data from embedded files

Togusa09:
Adds two additional methods for the MockFileSystem class to allow embedded files to be added:
- AddFileFromEmbeddedResource - Adds an embedded file to the file system
- AddFilesFromEmbeddedNamespace - Adds all embedded files under a namespace


tathamoddie:
Nice one! Should automatically pop up on NuGet in a few mins. Thanks for the PR.

PR #268, opened 2018-01-11
Use correct lock object for AllFiles

marcomeyer:
Avoid **"System.InvalidOperationException : Collection was modified; enumeration operation may not execute."**


tathamoddie:
Should be on NuGet as 2.1.0.178 in a couple of minutes.

Issue #272, opened 2018-02-07
TestingHelpers not usable on non-Windows platforms


fgreinacher:
The TestingHelpers package is currently not usable on non-Windows platforms as even the simplest operations are failing, e.g.:

```
Error Message:
 System.PlatformNotSupportedException : Access Control List (ACL) APIs are part of resource management on Windows and are not supported on this platform.
Stack Trace:
   at System.Security.AccessControl.ObjectSecurity..ctor()
   at System.Security.AccessControl.CommonObjectSecurity..ctor(Boolean isContainer)
   at System.Security.AccessControl.NativeObjectSecurity..ctor(Boolean isContainer, ResourceType resourceType)
   at System.Security.AccessControl.DirectorySecurity..ctor()
   at System.IO.Abstractions.TestingHelpers.MockDirectoryData..ctor()
   at System.IO.Abstractions.TestingHelpers.MockFileSystem.AddDirectory(String path)
```

skolima:
related: #252

tathamoddie:
Travis CI build now in place so that we can see what all the gaps are, and avid introducing new ones in new pull requests: https://travis-ci.org/System-IO-Abstractions/System.IO.Abstractions

fgreinacher:
Thanks

Issue #252, opened 2017-10-31
MockFileSystem.AddFile/Directory causes TypeInitializationException on Unix platforms


TAGC:
`MockFileSystem.AddFile` and `MockFileSystem.AddDirectory` create instances of `MockFileData`. When this type is first accessed on a Unix system, it causes `TypeInitializationException` to be thrown with this stack trace:

```
System.TypeInitializationException : The type initializer for 'System.IO.Abstractions.TestingHelpers.MockFileData' threw an exception.
[xUnit.net 00:00:01.6614320]       ---- System.PlatformNotSupportedException : Access Control List (ACL) APIs are part of resource management on Windows and are not supported on this platform.
```

This is probably because of [this line](https://github.com/tathamoddie/System.IO.Abstractions/blob/master/TestingHelpers/MockFileData.cs#L63) where you try instantiating a `FileSecurity` object.

Has this library even been tested on non-Windows platforms?

TAGC:
For anyone else looking for a workaround, I discovered a library called [Xunit.SkippableFact](https://github.com/AArnott/Xunit.SkippableFact) that lets you selectively skip running tests (facts or theories) based on some condition(s).

In my case I annotated all my xUnit tests that use `System.IO.Abstraction` in any way with `SkippableFact`/`SkippableTheory`:

```cs
[SkippableTheory]
[MemberData(nameof(UsbDeviceExamples))]
internal void Dev_Monitor_Should_Ignore_Usb_Device_Events_When_Not_Enabled(UsbDevice device)
{
    // GIVEN the device monitor is disabled.
    _devMonitor.Enabled = false;

    // WHEN a USB device is connected and disconnected.
    SimulateUsbDeviceArrival(device);
    SimulateUsbDeviceRemoval(device);

    // THEN no evente should have been published.
    _usbEvents.ShouldBeEmpty();
}
```

In the unit test class constructor (which is xUnit's equivalent of `setup()` and runs at the start of each test) I added this:

```cs
public DevMonitorSpec()
{
    Skip.IfNot(RunningOnWindows, "See https://github.com/tathamoddie/System.IO.Abstractions/issues/252");

    // Rest of setup.
}
```

Where `RunningOnWindows` is defined as:

```cs
private static readonly bool RunningOnWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
```

[...]

PR #280, opened 2018-03-18
Remove all refs to System.Security.AccessControl from netstandard1.4 version

verybadcat:
With these changes, this will install in a Xamarin project.

That said, I am not an expert in this area whatsoever. It is possible that I removed the refs from too many environments.

### DIFF
--- a/System.IO.Abstractions/DirectoryBase.cs
+++ b/System.IO.Abstractions/DirectoryBase.cs
@@ -1,5 +1,7 @@
 ﻿using System.Collections.Generic;
+#if NET40
 using System.Security.AccessControl;
+#endif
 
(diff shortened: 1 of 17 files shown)


pasn:
TestingHelpers don't work for me without this fix because of #252 . This looks like possible fix for aforementioned issue. Could someone take a look on it?

pasn:
Related: #272

pasn:
I can confirm that my tests using this library started to work after I applied this patch. I used xunit.runner.devices to run my xunit tests inside real Xamarin.Forms app on Android.

jpreese:
I am going to close this PR in favor of #289. Please continue the discussion there as they're both trying to solve the same problem.

Issue #281, opened 2018-03-20
Warning MSB3277 and .Net Standard 2.0


i4004:
I got this warnings when trying to compile my .NET Standard 2.0 library with dependency on System.IO.Abstractions:

```
1>C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\MSBuild\15.0\Bin\Microsoft.Common.CurrentVersion.targets(2052,5): warning MSB3277: Found conflicts between different versions of "System.Security.Principal.Windows" that could not be resolved.  These reference conflicts are listed in the build log when log verbosity is set to detailed.
1>C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\MSBuild\15.0\Bin\Microsoft.Common.CurrentVersion.targets(2052,5): warning MSB3277: Found conflicts between different versions of "System.Security.AccessControl" that could not be resolved.  These reference conflicts are listed in the build log when log verbosity is set to detailed.

```

## How to reproduce
Just create an empty .NET Standard 2.0 library project, reference latest System.IO.Abstractions and try compile.

---

The similar problem in another project: https://github.com/NLog/NLog/issues/2253

pasn:
This problem disappears when you update System.IO.FileSystem.AccessControl from version 4.3.0 (minimum required for this library) to 4.4.0 or newer. Starting from 4.4.0, System.Security.Claims and System.Security.Principal libraries are not required anymore.

PR #282, opened 2018-03-21
Fixed the Problem with .Net Standard 2.0

TclasenITVT:
I only added a compile target for .Net Standard 2.0 to the System.IO.Abstractions.csproj and System.IO.Abstractions.TestHelpers.csproj plus I updated the Version of used Nuget-Packages.
#281

### DIFF
--- a/System.IO.Abstractions/System.IO.Abstractions.csproj
+++ b/System.IO.Abstractions/System.IO.Abstractions.csproj
@@ -23,7 +23,7 @@
 
   <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
     <PackageReference Include="System.IO.FileSystem.AccessControl">
-      <Version>4.4.0</Version>
+      <Version>4.5.0</Version>
     </PackageReference>
     <PackageReference Include="System.IO.FileSystem.DriveInfo">
       <Version>4.3.1</Version>
@@ -38,10 +38,10 @@
   
     <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard1.4' ">
         <PackageReference Include="System.IO.FileSystem.AccessControl">
-            <Version>4.3.0</Version>
+            <Version>4.5.0</Version>
         </PackageReference>
         <PackageReference Include="System.IO.FileSystem.DriveInfo">
-            <Version>4.3.0</Version>
+            <Version>4.3.1</Version>
         </PackageReference>
         <PackageReference Include="System.IO.FileSystem.Watcher">
             <Version>4.3.0</Version>
--- a/System.IO.Abstractions/System.IO.Abstractions.csproj
+++ b/System.IO.Abstractions/System.IO.Abstractions.csproj
@@ -20,23 +20,8 @@
     <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
         <SignAssembly>True</SignAssembly>
     </PropertyGroup>
-
-  <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
-    <PackageReference Include="System.IO.FileSystem.AccessControl">
-      <Version>4.5.0</Version>
-    </PackageReference>
-    <PackageReference Include="System.IO.FileSystem.DriveInfo">
-      <Version>4.3.1</Version>
-    </PackageReference>
-    <PackageReference Include="System.IO.FileSystem.Watcher">
-      <Version>4.3.0</Version>
-    </PackageReference>
-    <PackageReference Include="System.Runtime.Serialization.Formatters">
-      <Version>4.3.0</Version>
-    </PackageReference>
-  </ItemGroup>
   
-    <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard1.4' ">
+    <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard1.4' or '$(TargetFramework)' == 'netstandard2.0'">
         <PackageReference Include="System.IO.FileSystem.AccessControl">
             <Version>4.5.0</Version>
         </PackageReference>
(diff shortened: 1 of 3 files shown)


richardwerkman:
This problem is breaking things for me, could somebody merge this and update nuget?

System.IO.Abstractions/System.IO.Abstractions.csproj
```diff
@@ -21,6 +21,21 @@
         <SignAssembly>True</SignAssembly>
     </PropertyGroup>
 
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
+    <PackageReference Include="System.IO.FileSystem.AccessControl">
+      <Version>4.4.0</Version>
```
pasn: I propose to use version 4.5.0 as it removes DLLs for many targets, but offers single netstandard2.0 for all these platforms added since 4.3.0. It's good that you haven't used 4.3.0 which won't work on Android

TclasenITVT: Changed it to use 4.5.0

pasn:
Hmm... seems that we don't need netstandard2.0 target. This problem disappears when you update System.IO.FileSystem.AccessControl from version 4.3.0 (minimum required for this library) to 4.4.0 or newer. Starting from 4.4.0, System.Security.Claims and System.Security.Principal libraries are not required anymore.

So it seems like switching minimum required version of System.IO.FileSystem.AccessControl for netstandard1.4 to >=4.5.0 will fix it.

tathamoddie:
So this PR is now redundant / wrong approach?

pasn:
I would bump System.IO.FileSystem.AccessControl for netstandard1.4 to >=4.5.0 to remove warning for all targeting the library in netstandard1.4. Otherwise developers of netstandard2.0 libraries utilizing System.IO.Abstractions w/o specifying any version of System.IO.FileSystem.AccessControl will get these warnings.

TclasenITVT:
Following @pasn suggestion removing NetStandard2.0 in favor of bumping up System.IO.FileSystem.AccessControl to 4.5.0 in NetStandard1.4 Target

jpreese:
I agree that the current state of the PR is much better than bumping the .NET standard requirement. The lower we can keep the .NET standard version, the more accessible we can keep the library.

pasn:
@jpreese this was never the intention of this PR to remove support for netstandard1.4. netstandard2.0 target was added in addition to existing one. Anyway, new target is now removed and I propose to merge it.

tathamoddie:
Note this comment from @davidfowl, where he says we should target _both_ 1.x and 2.0: https://twitter.com/davidfowl/status/1014560244720996352?s=19

pasn:
Then maybe it is beneficial - many libraries do this, so it may make sense. Anyway, the original issue is solved.

jpreese:
@pasn I'm just a little confused with the PR as it is now. My experience is that 1.4 works without an issue right now. I'm not getting any warnings. if I target .NET Standard 2.0, the warnings start to appear.

If I apply this change to my local environment, 1.4 still works and 2.0 still has warnings. Am I missing something?

TclasenITVT:
I have a Piece of Software that targets .netstandard 2 and for testing purposes I switched to the newest version of the System.IO.Abstrations Nuget Package and I don't get any warnings. So I guess we don't need this PR anymore. I could readd the Target for .netstandard 2.0 If that is what you want?

pasn:
@jpreese What warnings do you have when switchimg from AccessControl 4.3 to 4.5?

Original bug was about 2 warnings and one of them was about System.Security.Principal.Windows which is no longer required. Do you have the same warnings and is this library referenced (indirectly) by some other nuget in your project?

jpreese:
@TclasenITVT that's actually a little surprising.. even with the latest System.IO.Abstractions package, I get warnings in a project that targets .NET Standard 2.0

Based on Fowler's twitter response, we probably need to target .NET Standard 2.0 (and keep 1.4 as well) since it's a warning specific to .NET Standard 2.0

jpreese:
@pasn I don't get any warnings until my project targets 2.0

When my project targets 2.0, I get 

```
1>    There was a conflict between "System.Security.Principal.Windows, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" and "System.Security.Principal.Windows, Version=4.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a".
```

```
1>    There was a conflict between "System.Security.AccessControl, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" and "System.Security.AccessControl, Version=4.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a".
```

jpreese:
So to resolve this issue with .NET Standard 2, I think we need to add the target back.

pasn:
I can confirm that I have warnings mentioned in original issue #281 when creating new netstandard2.0 project. Sorry for confusion, but for my other netstandard2.0 project these warnings disappeared.

davidfowl:
I came here to say the same thing. 4.3.0 versions of packages are for ns1.x. You should try ns2.0 without any dependencies and see if that works.

TclasenITVT:
I tried it locally and could not get it to build. I guess that means we need the explizit references... 
All Errors are about Types not found. (e.g. DirectorySecurity, FileSecurity etc just to name a few)

TclasenITVT:
@davidfowl you were right.

System.IO.Abstractions/System.IO.Abstractions.csproj
```diff
@@ -21,12 +21,18 @@
         <SignAssembly>True</SignAssembly>
     </PropertyGroup>
 
-    <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard1.4' ">
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0'">
```
tathamoddie: Inconsistent whitespace in this block

tathamoddie:
Looking very close now!

tathamoddie:
Can we fix the whitespace (https://github.com/System-IO-Abstractions/System.IO.Abstractions/pull/282#pullrequestreview-135708653), and rebase it up onto the latest master? Then we can land it as a clean squash instead of the 17 commits we took to get here.

PR #286, opened 2018-05-07
Updates the Mock File System to throw DirectoryNotFound if creating a…

wleader:
… file in a directory that does not exist.

When creating a file in a real file system on Windows, if the filename is in a directory that does not exist, a DirectoryNotFoundException would be thrown. This change, causes the MockFile to behave in the same way.

Additionally some other tests had to be updated. These tests did not tell the mock filesystem that the directory existed before attempting the create.


wleader:
This request resolves the issue https://github.com/tathamoddie/System.IO.Abstractions/issues/278

PR #292, opened 2018-07-04
Remove stale packages.config

skolima:
This seems to be a leftover from before nestandard build migration. It's ignored by normal build process, but is confusing to some other tooling.


jpreese:
Thanks for this @skolima !

PR #294, opened 2018-07-04
Fix/disable tests that fail on Unix

fgreinacher:
This MR makes the Travis build green, by either
* fixing tests that were buggy (e.g. by not using MockUnixSupport consistently)
* fixing implementation issues
* ignoring tests that don't make sense on Unix (mostly ACL, UNC paths, character limitations in paths)

### DIFF
--- a/TestHelpers.Tests/MockDirectoryGetAccessControlTests.cs
+++ b/TestHelpers.Tests/MockDirectoryGetAccessControlTests.cs
@@ -16,6 +16,11 @@ namespace System.IO.Abstractions.TestingHelpers.Tests
         [TestCase("   ")]
         public void MockDirectory_GetAccessControl_ShouldThrowArgumentExceptionIfPathContainsOnlyWhitespaces(string path)
         {
+            if (MockUnixSupport.IsUnixPlatform()) 
+            {
+                Assert.Inconclusive("Unix does not support ACLs.");
+            }
+
             // Arrange
             var fileSystem = new MockFileSystem();
 
@@ -30,6 +35,11 @@ namespace System.IO.Abstractions.TestingHelpers.Tests
         [Test]
         public void MockDirectory_GetAccessControl_ShouldThrowDirectoryNotFoundExceptionIfDirectoryDoesNotExistInMockData()
         {
+            if (MockUnixSupport.IsUnixPlatform()) 
+            {
+                Assert.Inconclusive("Unix does not support ACLs.");
+            }
+
             // Arrange
             var fileSystem = new MockFileSystem();
             var expectedDirectoryName = XFS.Path(@"c:\a");
@@ -44,6 +54,11 @@ namespace System.IO.Abstractions.TestingHelpers.Tests
         [Test]
         public void MockDirectory_GetAccessControl_ShouldReturnAccessControlOfDirectoryData()
         {
+            if (MockUnixSupport.IsUnixPlatform()) 
+            {
+                Assert.Inconclusive("Unix does not support ACLs.");
+            }
+            
             // Arrange
             var expectedDirectorySecurity = new DirectorySecurity();
             expectedDirectorySecurity.SetAccessRuleProtection(false, false);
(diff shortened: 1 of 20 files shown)


TestHelpers.Tests/MockDirectoryGetAccessControlTests.cs
```diff
@@ -16,6 +16,11 @@ public class MockDirectoryGetAccessControlTests
         [TestCase("   ")]
         public void MockDirectory_GetAccessControl_ShouldThrowArgumentExceptionIfPathContainsOnlyWhitespaces(string path)
         {
+            if (MockUnixSupport.IsUnixPlatform())
```
jpreese: Instead of this block of code throughout the test project, what do we think about categorizing Windows tests?

I think this would declutter the test runner a bit more (no need for inconclusives) and would also clearly show that the feature we're testing is intended for Windows.

TestHelpers.Tests/MockDirectoryInfoTests.cs
```diff
@@ -202,8 +202,11 @@ public static IEnumerable<object[]> MockDirectoryInfo_FullName_Data
                 yield return new object[] { XFS.Path(@"c:\temp\\folder"), XFS.Path(@"c:\temp\folder") };
                 yield return new object[] { XFS.Path(@"c:\temp//folder"), XFS.Path(@"c:\temp\folder") };
                 yield return new object[] { XFS.Path(@"c:\temp//\\///folder"), XFS.Path(@"c:\temp\folder") };
-                yield return new object[] { XFS.Path(@"\\unc\folder"), XFS.Path(@"\\unc\folder") };
-                yield return new object[] { XFS.Path(@"\\unc/folder\\foo"), XFS.Path(@"\\unc\folder\foo") };
+                if (!MockUnixSupport.IsUnixPlatform())
```
jpreese: I think this would be a little better if we just removed these values from the test data, and then made a new Windows test that checked just the two unc paths. This way we're testing drive paths and unc paths in two different tests. What do you think?

TestHelpers.Tests/MockFileSystemTests.cs
```diff
@@ -153,16 +153,16 @@ public void MockFileSystem_AddFile_ShouldMatchCapitalization_PerfectMatch()
             fileSystem.AddDirectory(MockUnixSupport.Path(@"C:\LOUD"));
 
             // Act
-            fileSystem.AddFile(@"C:\test\file.txt", "foo");
-            fileSystem.AddFile(@"C:\LOUD\file.txt", "foo");
-            fileSystem.AddDirectory(@"C:\test\SUBDirectory");
-            fileSystem.AddDirectory(@"C:\LOUD\SUBDirectory");
+            fileSystem.AddFile(MockUnixSupport.Path(@"C:\test\file.txt"), "foo");
```
jpreese: Should we, for consistency, use XFS as the namespace instead of MockUnixSupport? My first gut reaction was what is the difference between MockUnitSupport and XFS? If we chose one and stuck with it, I think the experience would be a little better.

TestHelpers.Tests/MockDirectoryTests.cs
```diff
@@ -745,11 +760,16 @@ public void MockDirectory_GetFiles_ShouldThrowAnArgumentException_IfSearchPatter
 
         public static IEnumerable<string> GetSearchPatternForTwoDotsExceptions()
         {
-            yield return @"a..\b";
             yield return @"a../b";
             yield return @"../";
-            yield return @"..\";
-            yield return @"aaa\vv..\";
+
+            if (!MockUnixSupport.IsUnixPlatform())
```
jpreese: This is a little confusing to me. If we're not on a Unix platform, go ahead and add these values to the test data.. but the comment says that these aren't a problem on Unix, so why would we exclude these? If they do need to be excluded, I would go with a similar approach as outlined above where we'd just go ahead and remove these and leverage a Windows test.

jpreese:
In order to increase our velocity, I'm just going to go ahead and merge in this PR since it will immensely help moving forward with PRs and validating against Linux. I think everything will work, I just have some opinions as to how things may have been done a little cleaner. 

Regardless if we do agree with the above comments, we can revisit in a future PR.

fgreinacher:
Thanks - busy days for me. Will take care of the follow-ups!

jpreese:
Sounds good! As stated above, they were mostly just suggestions. If you feel differently, I'd love to hear your counter thoughts.

PR #296, opened 2018-07-06
Add CLS compliance

### LINKED ISSUE #279: No longer CLS compliant
Before #225 the library was marked as CLS compliant
image

Since the change, it is no longer possible to install and use v2.1
image
image

The underlying issue is that there is no longer AssemblyInfo.cs, which seems the only place where CLS compliance can be reported: https://github.com/dotnet/cli/issues/1520#issuecomment-188272775

Please restore CLS compliance.

jpreese:
Resolves #279


RussKie:
Champion!

Issue #184, opened 2016-11-22
Read-only mocked files should not be deletable.


AlexanderPruss:
See https://msdn.microsoft.com/en-us/library/system.io.file.delete(v=vs.110).aspx

A read-only mock should throw an UnauthorizedAccessException if it gets deleted.

AlexanderPruss:
https://github.com/tathamoddie/System.IO.Abstractions/pull/185

jpreese:
Should be a pretty straightforward merge. Looking into this one soon.

jpreese:
Confirmed it's an issue -- will be handled in #303. Thanks for finding this, @AlexanderPruss !

tathamoddie:
Fixed by 370052d

PR #303, opened 2018-07-08
Deleting ReadOnly File Throws Exception

jpreese:
For #184


tathamoddie:
I rearranged the existing 7 commits into 2 distinct ones: one for code formatting cleanup, and then one for the functional change.

PR #305, opened 2018-07-08
Remove XOR when encrypting/decrypting files

### LINKED ISSUE #304: Encrypt/Decrypt inconsistent with .NET Framework
The behaviour of the mock `File.Encrypt` is inconsistent with the actual file system's behaviour on .NET 4.

.NET sets the encrypted attribute in NTFS, which encrypts the file. Thus, reading the file content after encrypting it transparently decrypts the content.
```
> var fs = new FileSystem();
> fs.File.WriteAllText(@"D:\test.txt", "foo");
> fs.File.ReadAllText(@"D:\test.txt")
"foo"
> fs.File.Encrypt(@"D:\test.txt")
> fs.File.ReadAllText(@"D:\test.txt")
"hi"
```

The mock currently performs an XOR on the content bytes, so that we cannot read the file's content
```
> var mfs = new MockFileSystem();
> mfs.AddFile(@"D:\test.txt", new MockFileData("bar"));
> mfs.File.ReadAllText(@"D:\test.txt")
"bar"
> mfs.File.Encrypt(@"D:\test.txt")
> mfs.File.ReadAllText(@"D:\test.txt")
"b`p"
```

I believe the only observable change (from the .NET app's point of view) should be the added `Encrypted` attribute.

hangy:
Fixes #304 by removing XOR of the in-memory data when encrypting/decrypting files with the MockFileSystem.

### DIFF
--- a/TestHelpers.Tests/MockFileTests.cs
+++ b/TestHelpers.Tests/MockFileTests.cs
@@ -559,31 +559,6 @@ namespace System.IO.Abstractions.TestingHelpers.Tests
 
 #if NET40
         [Test]
-        public void MockFile_Encrypt_ShouldEncryptTheFile()
-        {
-            // Arrange
-            const string Content = "Demo text content";
-            var fileData = new MockFileData(Content);
-            var filePath = XFS.Path(@"c:\a.txt");
-            var fileSystem = new MockFileSystem(new Dictionary<string, MockFileData>
-            {
-                {filePath, fileData }
-            });
-
-            // Act
-            fileSystem.File.Encrypt(filePath);
-
-            string newcontents;
-            using (var newfile = fileSystem.File.OpenText(filePath))
-            {
-                newcontents = newfile.ReadToEnd();
-            }
-
-            // Assert
-            Assert.AreNotEqual(Content, newcontents);
-        }
-
-        [Test]
         public void MockFile_Encrypt_ShouldSetEncryptedAttribute()
         {
             // Arrange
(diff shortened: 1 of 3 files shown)


TestHelpers.Tests/MockFileTests.cs
```diff
@@ -558,31 +558,6 @@ public void Serializable_can_deserialize()
...
-            {
-                newcontents = newfile.ReadToEnd();
-            }
-
-            // Assert
-            Assert.AreNotEqual(Content, newcontents);
-        }
-
         [Test]
         public void MockFile_Encrypt_ShouldSetEncryptedAttribute()
```
jpreese: It probably makes sense to remove all Encrypt/Decrypt tests from MockFileTests.cs -- they seem to be testing the exact same logic, with pretty much the exact same test code. What we're ultimately testing lives in MockFileInfo anyway.

hangy: The tests do seem to be redundant at first, but on a second thought, they kinda do make sense. If you want to have a blackbox test for `MockFile.Encrypt`, then asserting that the `Encrypted` attribute is set, would be correct. Right now, I don't see another option to assert that those methods work as document.

jpreese: Yeah I can see that. While they both technically call down to the same implementation, I spose it's possible to break something in MockFile before it even gets to MockFileInfo

jpreese:
Thanks @hangy! 🎉

PR #307, opened 2018-07-09
Implement File.Replace, FileInfo.Replace for TestHelpers

rkoeninger:
Replaces #164.

This implements the 2 overloads of `Replace` in each of `MockFile` and `MockFileInfo`, which previously threw exceptions.

Added 9 unit tests, 4 for `File.Replace`, 5 for `FileInfo.Replace`.

Incidentally, also added a `File` property to `IMockFileDataAccessor` so the `FileBase` would be accessible. Didn't need to implement the property as it already existed on `MockFileSystem`.

Implementation is surrounded with `#if NET40 ... #endif` as the `Replace` methods apparently don't exist in .NET Standard.

I'm going to copy my comment from the other PR in case it preempts any questions which held it up:

---

According to the [docs for System.IO.File.Replace](https://msdn.microsoft.com/en-us/library/system.io.file.replace(v=vs.110).aspx), it looks like a backup is part of the functionality. (EDIT: and, yes, it's also supposed to delete the source file. I confirmed this behavior by testing `System.IO.File.Replace`)

You can see [in the reference source](https://referencesource.microsoft.com/#mscorlib/system/io/file.cs,d4b8da02b41f19a4), it also mentions a backup path, although the actual operation is in native code.

This PR should probably also implement the test helper for File.Replace and implement FileInfo.Replace in terms of that, as [seen in the reference source](https://referencesource.microsoft.com/#mscorlib/system/io/fileinfo.cs,c41f2d5cbd26ee89).

---

### DIFF
--- a/TestHelpers.Tests/MockFileInfoTests.cs
+++ b/TestHelpers.Tests/MockFileInfoTests.cs
@@ -464,5 +464,78 @@ namespace System.IO.Abstractions.TestingHelpers.Tests
 
             Assert.Throws<FileNotFoundException>(action);
         }
+
+#if NET40
+        [Test]
+        public void MockFileInfo_Replace_ShouldReplaceFileContents()
+        {
+            // Arrange
+            var fileSystem = new MockFileSystem();
+            fileSystem.AddFile(XFS.Path(@"c:\temp\file1.txt"), new MockFileData(@"line 1\r\nline 2"));
+            fileSystem.AddFile(XFS.Path(@"c:\temp\file2.txt"), new MockFileData(@"line 3\r\nline 4"));
+            var fileInfo1 = fileSystem.FileInfo.FromFileName(XFS.Path(@"c:\temp\file1.txt"));
+            var fileInfo2 = fileSystem.FileInfo.FromFileName(XFS.Path(@"c:\temp\file2.txt"));
+
+            // Act
+            fileInfo1.Replace(XFS.Path(@"c:\temp\file2.txt"), null);
+
+            Assert.AreEqual(@"line 1\r\nline 2", fileInfo2.OpenText().ReadToEnd());
+        }
+
+        [Test]
+        public void MockFileInfo_Replace_ShouldCreateBackup()
+        {
+            // Arrange
+            var fileSystem = new MockFileSystem();
+            fileSystem.AddFile(XFS.Path(@"c:\temp\file1.txt"), new MockFileData(@"line 1\r\nline 2"));
+            fileSystem.AddFile(XFS.Path(@"c:\temp\file2.txt"), new MockFileData(@"line 3\r\nline 4"));
+            var fileInfo1 = fileSystem.FileInfo.FromFileName(XFS.Path(@"c:\temp\file1.txt"));
+            var fileInfo2 = fileSystem.FileInfo.FromFileName(XFS.Path(@"c:\temp\file2.txt"));
+            var fileInfo3 = fileSystem.FileInfo.FromFileName(XFS.Path(@"c:\temp\file3.txt"));
+
+            // Act
+            fileInfo1.Replace(XFS.Path(@"c:\temp\file2.txt"), XFS.Path(@"c:\temp\file3.txt"));
+
+            Assert.AreEqual(@"line 3\r\nline 4", fileInfo3.OpenText().ReadToEnd());
+        }
+
+        [Test]
+        public void MockFileInfo_Replace_ShouldReturnDestinationFileInfo()
+        {
+            // Arrange
+            var fileSystem = new MockFileSystem();
+            fileSystem.AddFile(XFS.Path(@"c:\temp\file1.txt"), new MockFileData(@"line 1\r\nline 2"));
+            fileSystem.AddFile(XFS.Path(@"c:\temp\file2.txt"), new MockFileData(@"line 3\r\nline 4"));
+            var fileInfo1 = fileSystem.FileInfo.FromFileName(XFS.Path(@"c:\temp\file1.txt"));
+            var fileInfo2 = fileSystem.FileInfo.FromFileName(XFS.Path(@"c:\temp\file2.txt"));
+
+            // Act
+            var result = fileInfo1.Replace(XFS.Path(@"c:\temp\file2.txt"), null);
+
+            Assert.AreEqual(fileInfo2.FullName, result.FullName);
+        }
+
+        [Test]
+        public void MockFileInfo_Replace_ShouldThrowIfSourceFileDoesNotExist()
+        {
+            // Arrange
+            var fileSystem = new MockFileSystem();
+            fileSystem.AddFile(XFS.Path(@"c:\temp\file2.txt"), new MockFileData(@"line 3\r\nline 4"));
+            var fileInfo = fileSystem.FileInfo.FromFileName(XFS.Path(@"c:\temp\file1.txt"));
+
+            Assert.Throws<FileNotFoundException>(() => fileInfo.Replace(XFS.Path(@"c:\temp\file2.txt"), null));
+        }
+
+        [Test]
+        public void MockFileInfo_Replace_ShouldThrowIfDestinationFileDoesNotExist()
+        {
+            // Arrange
+            var fileSystem = new MockFileSystem();
+            fileSystem.AddFile(XFS.Path(@"c:\temp\file1.txt"), new MockFileData(@"line 1\r\nline 2"));
+            var fileInfo = fileSystem.FileInfo.FromFileName(XFS.Path(@"c:\temp\file1.txt"));
+
+            Assert.Throws<FileNotFoundException>(() => fileInfo.Replace(XFS.Path(@"c:\temp\file2.txt"), null));
+        }
+#endif
     }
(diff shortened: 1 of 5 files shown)


jpreese:
Are we sure System.IO.File.Replace() is not in .NET Standard? I used to know of a website that would tell you which version you needed to target given a class member, but it slips my mind. Worst case I could technically play around with the portability analyzer (https://docs.microsoft.com/en-us/dotnet/standard/analyzers/portability-analyzer)

But at first muster, I just created a .NET Standard 1.4 project and was able to call System.IO.File.Replace.

Am I missing something?

rkoeninger:
Hmmm... it would seem I mis-spoke/was wrong.

The abstract declaration of the `Replace`  methods are wrapped in `#if NET40 ... #endif` in [FileBase](https://github.com/System-IO-Abstractions/System.IO.Abstractions/blob/master/System.IO.Abstractions/FileBase.cs#L228-L231) and [FileInfoBase](https://github.com/System-IO-Abstractions/System.IO.Abstractions/blob/master/System.IO.Abstractions/FileInfoBase.cs#L26-L29), thus the implementations in the mock classes are wrapped.

Whether `Replace` is available in .NET standard, I don't know.

rkoeninger:
After some brief research, I see that it appears to be present in [.NET Standard 2.0](https://github.com/dotnet/standard/blob/master/docs/versions/netstandard2.0_ref.md), but wasn't in [.NET Standard 1.0](https://github.com/dotnet/standard/blob/master/docs/versions/netstandard1.0_ref.md). I don't know where it was introduced in between.

jpreese:
Yeah.. I'll need to think about this a bit more. The whole NET40 stuff has always been an uncertainty of mine with this library. It never seems incredibly clear why some methods were chosen, at least to me.

For example, in that same document you linked (which is pretty sweet btw, essentially what I was looking for), WriteAllLines() is there. Yet, we dont have NET40 checks around WriteAllLines..

rkoeninger:
Maybe another Issue/PR needs to be opened to discuss and work on multi-targeting instead of slipping it into this PR.

The scope of this PR was intended to just to get the mock Replace method. Is there anything wrong with the implementation relative to how the rest of the library is written?

TestingHelpers/MockFile.cs
```diff
@@ -525,12 +525,30 @@ public override IEnumerable<string> ReadLines(string path, Encoding encoding)
...
         public override void Replace(string sourceFileName, string destinationFileName, string destinationBackupFileName)
         {
-            throw new NotImplementedException(StringResources.Manager.GetString("NOT_IMPLEMENTED_EXCEPTION"));
+            Replace(sourceFileName, destinationFileName, destinationBackupFileName, false);
         }
 
         public override void Replace(string sourceFileName, string destinationFileName, string destinationBackupFileName, bool ignoreMetadataErrors)
         {
-            throw new NotImplementedException(StringResources.Manager.GetString("NOT_IMPLEMENTED_EXCEPTION"));
+            if (!mockFileDataAccessor.FileExists(sourceFileName))
```
jpreese: Looking at the documentation for Replace(), a ArgumentNullException is thrown when the destFileName parameter is null.

TestingHelpers/MockFile.cs
```diff
@@ -525,12 +525,30 @@ public override IEnumerable<string> ReadLines(string path, Encoding encoding)
...
+            }
+
+            if (!mockFileDataAccessor.FileExists(destinationFileName))
+            {
+                throw new FileNotFoundException(string.Format(CultureInfo.InvariantCulture, StringResources.Manager.GetString("COULD_NOT_FIND_FILE_EXCEPTION"), destinationFileName));
+            }
+
+            var mockFile = new MockFile(mockFileDataAccessor);
+
+            if (destinationBackupFileName != null)
```
jpreese: Sanity check -- what happens when the destinationBackupFileName doesn't exist in System.IO? i.e. The value is not null, but the directory doesn't exist. Or the backupfile already exists.

I think this would throw a DirectoryNotFoundException, which isn't listed as a possibility in the documentation. Might be a good candidate for a test.

jpreese:
@rkoeninger that's fair. I've included some review comments.

TestHelpers.Tests/MockFileInfoTests.cs
```diff
@@ -453,5 +453,78 @@ public void MockFileInfo_CopyTo_ThrowsExceptionIfSourceDoesntExist()
...
+            fileSystem.AddFile(XFS.Path(@"c:\temp\file1.txt"), new MockFileData(@"line 1\r\nline 2"));
+            fileSystem.AddFile(XFS.Path(@"c:\temp\file2.txt"), new MockFileData(@"line 3\r\nline 4"));
+            var fileInfo1 = fileSystem.FileInfo.FromFileName(XFS.Path(@"c:\temp\file1.txt"));
+            var fileInfo2 = fileSystem.FileInfo.FromFileName(XFS.Path(@"c:\temp\file2.txt"));
+            var fileInfo3 = fileSystem.FileInfo.FromFileName(XFS.Path(@"c:\temp\file3.txt"));
+
+            // Act
+            fileInfo1.Replace(XFS.Path(@"c:\temp\file2.txt"), XFS.Path(@"c:\temp\file3.txt"));
+
+            Assert.AreEqual(@"line 3\r\nline 4", fileInfo3.OpenText().ReadToEnd());
```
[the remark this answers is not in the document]
rkoeninger: Are they supposed to match? File3 (the backup file) is supposed to contain the former contents of File2 (the destination file), not the resulting contents. They are never expected to have the same contents at the same time.

Before:

* File1 contains A
* File2 contains B
* File3 doesn't exist

After:

* File1 doesn't exist
* File2 contains A
* File3 contains B

TestHelpers.Tests/MockFileInfoTests.cs
```diff
@@ -453,5 +453,78 @@ public void MockFileInfo_CopyTo_ThrowsExceptionIfSourceDoesntExist()
...
             Assert.Throws<FileNotFoundException>(action);
         }
+
+#if NET40
+        [Test]
+        public void MockFileInfo_Replace_ShouldReplaceFileContents()
+        {
+            // Arrange
+            var fileSystem = new MockFileSystem();
+            fileSystem.AddFile(XFS.Path(@"c:\temp\file1.txt"), new MockFileData(@"line 1\r\nline 2"));
```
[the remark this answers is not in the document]
rkoeninger: Sure, cleaned up tests a bit

TestingHelpers/MockFile.cs
rkoeninger: Added explicit null checks

TestingHelpers/MockFile.cs
rkoeninger: Testing of `System.IO.File.Replace` showed that it didn't care if backup file already existed, it would just override it. But if the directory that the backup file path is in didn't exist, there would be a `DirectoryNotFoundException`. Added tests to assert `MockFile` replicates this behavior.

rkoeninger:
@jpreese Addressed review comments.

There could be a another PR to do more test cleanup.

Replace also isn't using the `ignoreMetadataErrors` flag, but that's not used in managed code in the BCL either. It's a native windows function. Considering how arcane that is, I don't think MockFileSystem even emulates whatever it's supposed to handle, so I think handling it can be deferred.

Decompiled code from mscorlib, System.IO.File:

```csharp
private static void InternalReplace(string sourceFileName, string destinationFileName, string destinationBackupFileName, bool ignoreMetadataErrors)
{
	string fullPathInternal = Path.GetFullPathInternal(sourceFileName);
	string fullPathInternal2 = Path.GetFullPathInternal(destinationFileName);
	string text = null;
	if (destinationBackupFileName != null)
	{
		text = Path.GetFullPathInternal(destinationBackupFileName);
	}
	if (CodeAccessSecurityEngine.QuickCheckForAllDemands())
	{
		FileIOPermission.EmulateFileIOPermissionChecks(fullPathInternal);
		FileIOPermission.EmulateFileIOPermissionChecks(fullPathInternal2);
		if (text != null)
		{
			FileIOPermission.EmulateFileIOPermissionChecks(text);
		}
	}
	else
	{
		FileIOPermission fileIOPermission = new FileIOPermission(FileIOPermissionAccess.Read | FileIOPermissionAccess.Write, new string[]
		{
			fullPathInternal,
			fullPathInternal2
		});
		if (text != null)
		{
			fileIOPermission.AddPathList(FileIOPermissionAccess.Write, text);
		}
		fileIOPermission.Demand();
	}
	int num = 1;
	if (ignoreMetadataErrors)
	{
		num |= 2;
	}
	if (!Win32Native.ReplaceFile(fullPathInternal2, fullPathInternal, text, num, IntPtr.Zero, IntPtr.Zero))
	{
		__Error.WinIOError();
	}
}
```

TestingHelpers/MockFile.cs
jpreese: Alright cool, thats what I figured as its been a common theme with the other File create methods. They usually throw a Directory error if one doesn't exist, but I didn't see it in the documentation. Must be a gap on their side.

Issue #156, opened 2016-02-09
MockDirectoryInfo implement LastWriteTimeUtc


johncmckim:
I need to use LastWriteTimeUtc on MockDirectoryInfo which is not implemented. Could this be implemented please?

tathamoddie:
Hi! The best route to an implementation is a pull request with tests. Shout out if you need help.

johncmckim:
No worries. I'll put something together.

jpreese:
To be resolved by #311

jpreese:
#311 has been merged. Closing!

PR #311, opened 2018-07-09
Implement Creation, LastAccess and LastWrite Dates on MockDirectoryInfo

robkeim:
This contains the following changes:
- Original changes from #157 
- Auto-formatting from Visual Studio
- Moving to use properties with default values (leveraging new C# 6 syntax)


fgreinacher:
Thanks - looks good from my side. Waiting for an additional pair of eyes to review.

TestingHelpers/MockDirectoryInfo.cs
```diff
@@ -19,21 +19,30 @@ public class MockDirectoryInfo : DirectoryInfoBase
...
+        MockFileData GetMockFileData(bool returnNullObject)
         {
-            get { return mockFileDataAccessor.GetFile(directoryPath); }
+            var file = mockFileDataAccessor.GetFile(directoryPath);
+
+            if (file == null)
+            {
+                if (!returnNullObject)
+                {
+                    throw new FileNotFoundException("File not found", directoryPath);
```
jpreese: I would use the string in the Resources.resx file

TestingHelpers/MockDirectoryInfo.cs
```diff
@@ -19,21 +19,30 @@ public class MockDirectoryInfo : DirectoryInfoBase
...
-            this.mockFileDataAccessor = mockFileDataAccessor;
+            this.mockFileDataAccessor = mockFileDataAccessor ?? throw new ArgumentNullException("mockFileDataAccessor");
 
             directoryPath = mockFileDataAccessor.Path.GetFullPath(directoryPath);
 
             this.directoryPath = EnsurePathEndsWithDirectorySeparator(directoryPath);
         }
 
-        MockFileData MockFileData
+        MockFileData GetMockFileData(bool returnNullObject)
```
jpreese: Could we get private added to this?

TestHelpers.Tests/MockDirectoryInfoTests.cs
```diff
@@ -283,5 +283,346 @@ public void MockDirectoryInfo_ToString_ShouldReturnDirectoryName()
...
+
+            // Act
+            var newTime = DateTime.Now;
+            directoryInfo.LastWriteTime = newTime;
+
+            // Assert
+            Assert.AreEqual(newTime, directoryInfo.LastWriteTime);
+        }
+
+        static void ExecuteDefaultValueTest<T>(Func<MockDirectoryInfo, T> getDateValue, T expected)
```
jpreese: Could we get private added to this?

TestHelpers.Tests/MockDirectoryInfoTests.cs
```diff
@@ -283,5 +283,346 @@ public void MockDirectoryInfo_ToString_ShouldReturnDirectoryName()
...
+            var file = new MockDirectoryInfo(fileSystem, path);
+
+            // Act
+            var actual = getDateValue(file);
+
+            // Assert
+            Assert.AreEqual(expected, actual);
+        }
+
+        static void ExecuteSetDefaultValueThrowsTest(Action<MockDirectoryInfo> setDateValue)
```
jpreese: Could we get private added to this?

TestHelpers.Tests/MockDirectoryInfoTests.cs
```diff
@@ -283,5 +283,346 @@ public void MockDirectoryInfo_ToString_ShouldReturnDirectoryName()
             // Assert
             Assert.AreEqual(directoryPath, str);
         }
+
+        [Test]
+        public void MockDirectoryInfo_Attributes_ShouldReturnDefaultAttributeIfNotExists()
+        {
+            ExecuteDefaultValueTest((d) => d.Attributes, MockFileData.NullObject.Attributes);
```
jpreese: This might start a bit of an opinion war, but I'm not a huge fan of this when writing tests. I dont so much mind hiding setup stuff behind a 

```csharp
var directoryInfo = CreateDefaultDirectoryInfo();
directoryInfo.Attributes // whatever act stuff
Assert(//whatever assert)
```

But this approach seems a little confusing to read. Not sure what everyone else's thoughts are.

jpreese:
I have a couple for sure changes, small nits. And then a quick discussion about execution helpers.

TestingHelpers/MockDirectoryInfo.cs
jpreese: Oh, I also wanted to bring up as food for thought. I'm usually pretty against parameters that change behaviors within a method. Usually parameters like this where its a boolean and branches of true/false. To me, it's a little confusing to say..

```csharp
GetMockFileData(true); // gasp! what does true do?!
```

You have to dive into the implementation to figure it out, rather than just read the intention. A common fix for this is to just have two separate methods. In this case, one for returning the null object (always used in the getter), and one for not (always used in the setter). GetMockFileDataWithExceptionChecking, who knows, naming is hard.

Though after typing all this out, it got me to thinking. Why is the behavior different for getting and setting when it comes to returning a NullObject? I would think we'd return FileNotFound for either case?

TestHelpers.Tests/MockDirectoryInfoTests.cs
robkeim: I tend to agree that in *test* code favoring readability at the expense of redundancy (code duplication) is not a bad thing. I'll go up ahead and update this.

TestingHelpers/MockDirectoryInfo.cs
robkeim: Great question @jpreese, but I don't have the answer right now. I think I'm going to have to do some more digging to get a better understanding of the overall project and where this PR fits into that puzzle.

jpreese:
@robkeim Yeah this looks a lot better to me. The only lingering question I have is the GetMockFileData boolean switch between a NullObject and not. It just seems strange.. if the file can't be found why can we get the attributes, but not set them.

System.IO behavior:

```csharp
Directory.SetLastWriteTimeUtc(@"C:\foo\bar", new DateTime()); //throws DirectoryNotFoundException
Directory.GetLastWriteTimeUtc(@"C:\moo\moo"); //returns a datetime of 1/1/1601 @ 12:00AM
```

So that would explain that. With that in mind, it seems like we'd want a DirectoryNotFoundException because thats the behavior I'm seeing. Which is a little weird because it's not listed in their documentation https://msdn.microsoft.com/en-us/library/system.io.directory.setlastwritetimeutc(v=vs.110).aspx (FileException is)

And maybe we could clean it up so we're not passing in a boolean to make the code slightly more expressive? Thoughts? Other than that, this looks really good.

jpreese:
https://github.com/dotnet/docs/issues/6394

rkoeninger:
I think this Fixes #156.

jpreese:
@rkoeninger it does, yes.

jpreese:
Still waiting on a final response from https://github.com/dotnet/dotnet-api-docs/pull/336, but it seems like it can be either or.

jpreese:
It seems like the exception that is called may not ultimately matter https://github.com/dotnet/docs/issues/6431#issuecomment-404647411

jpreese:
@robkeim I think if we just resolve the merge conflicts, we should be good. The type of exception that is thrown seems to be all over the place, and may even be dependent on OS. So it's probably not a huge deal, at least now. The library has always thrown a FileNotFoundException, so there's no real compelling reason to change it until it pops up, if ever.

The only potential refactor I'd like to see is the passing in of a boolean, but even that is a small nit and something we can revisit later if you'd rather just merge this in so we can get the functionality live.

robkeim:
Sorry for not responding sooner @jpreese it's been a busy week here, but I appreciate you digging into this for me. Let me take another look at the boolean and see what I can do. I'll also resolve the merge conflicts.

robkeim:
@jpreese can you this PR another look when you get a chance? I removed the boolean argument and split it into two functions that I hope have a more clear naming.

jpreese:
Yep. This looks good to me now. Thanks! 🎉

Issue #308, opened 2018-07-09
How to find source code for specific nuget version


pasn:
It seems that package is now automatically generated on every commit to master (or do you create nuget by hand every time)?

It would be nice if there was a git tag created for every release.

pasn:
How to do it in yml file? Would sth like this work?

```
after_deploy:
- git tag -a v2.1.0.{build} -m "Version v2.1.0.{build}"
- git push origin v2.1.0.{build}
```

jpreese:
We actually just had a discussion about this not too long ago in #295. As of right now, each merge into master will create a new package release, automatically. It's possible we could tag each merge, but that might be a little extreme given our current release model. 

To answer your question though, you can go to the history in AppVeyor for this project (https://ci.appveyor.com/project/tathamoddie/system-io-abstractions/history) and each time you see a full release (no hyphen at the end) the associated commit should be visible.

pasn:
That's what I needed, thanks. :)

How can one find this link? If it is not obvious what server is being used, maybe you could add it to summary at the top of main repository's page?

jpreese:
That is an option, yes. I'm trying to decide if having AppVeyor tag every release would be a bad thing. It's a lot of tags, but it's a lot of releases.

That way were not reliant on a 3rd party to keep our version history that may get cleaned up, and it's not very searchable.

fgreinacher:
Independently of the tagging question, we could consider using https://github.com/dotnet/sourcelink for a much nicer debugging experience.

fgreinacher:
Starting now all versions deployed to NuGet.org will have corresponding [tags/releases](https://github.com/System-IO-Abstractions/System.IO.Abstractions/releases).

Issue #295, opened 2018-07-05
Release Strategy and Versioning


jpreese:
Right now each merge into master cuts a new release to NuGet. This is going to cause a lot of unneeded versions to be released and prevents us from putting together a more holistic release.

Versioning is also driven from the appveyor config file -- I'd like to see versioning driven from the releases page and tags so we're not having to redeploy the config file for each release. This also means we can get away from having releases with build numbers, and lean more on semver and bump the release numbers as needed.

We can take this a couple of ways.. we can modify the appveyor config to only release on tags, so we have the ability to merge into master without firing off a release. Or we could merge into a release branch and only merge into master when we feel we have enough content for an actual release.

I lean more in favor of cutting releases when tags are pushed.

With all of that in mind, just wanted to leave this here for a little bit of discussion before going forward.

tathamoddie:
What's the problem with too many versions? If each little bit moves us forward, what's an "unneeded" version?

When it comes to SemVer, the API surface is incredibly stable: it's defined by .NET and only has immense backwards compatibility. Thus, the major version is almost never touched anyway.

I like SemVer, I just think we get a lot of it for free here in this particular library context.

I also hate the idea of ever bundling up unshipped work and delaying it until the next 'release' event. NuGet packages are free. Ship a million of them!

jpreese:
I'm not entirely opposed to the current strategy, it's obviously worked for this long. To me an unneeded version would be say, the license change to MIT. There's really no value in bumping that up as we didn't really change anything. Though those types of merges are going to be rare. Maybe even pure refactorings where we introduce no new features or bug fixes.

I mostly approached it from the idea that if I'm a consumer of the library, and I want to update my project that uses it. I need to redeploy my app. If I want to stay current, that could be everyday.

I completely agree about wanting to ship something as soon as it's ready, but cases where we have multiple PRs to merge.. each merge will be a new version vs. combining them into a singular release. We don't delay the release, just consolidate. Though we could do this today already with some form of a release branch.

TLDR if you're ok with it, I am. There's a lot more issues out there that need attention instead of versioning.

My parting question would be, when do you typically update the minor or patch numbers in the version? Right now, I'm not certain of the criteria.

tathamoddie:
The license change legitimately triggers a different package: some people might want to update to MIT immediately, and have it in their app as that.

tathamoddie:
I bumped the major version when I broke something on the build server and stuffed a version number.

[...]

PR #313, opened 2018-07-09
Misc code cleanup

robkeim:
Leveraging some new C# features to simplify the code.

Do we have style guidelines we're following? The neat freak in me was noticing inconsistencies but didn't know how to clean them up. For example, usings are mixed between being outside and inside the namespace.

### DIFF
--- a/System.IO.Abstractions/DirectoryInfoWrapper.cs
+++ b/System.IO.Abstractions/DirectoryInfoWrapper.cs
@@ -11,7 +11,7 @@ namespace System.IO.Abstractions
 
         public DirectoryInfoWrapper(DirectoryInfo instance)
         {
-            this.instance = instance ?? throw new ArgumentNullException("instance");
+            this.instance = instance ?? throw new ArgumentNullException(nameof(instance));
         }
 
         public override void Delete()
--- a/System.IO.Abstractions/DirectoryInfoWrapper.cs
+++ b/System.IO.Abstractions/DirectoryInfoWrapper.cs
@@ -11,12 +11,7 @@ namespace System.IO.Abstractions
 
         public DirectoryInfoWrapper(DirectoryInfo instance)
         {
-            if (instance == null)
-            {
-                throw new ArgumentNullException("instance");
-            }
-
-            this.instance = instance;
+            this.instance = instance ?? throw new ArgumentNullException("instance");
         }
 
         public override void Delete()
--- a/TestingHelpers/MockFileSystem.cs
+++ b/TestingHelpers/MockFileSystem.cs
@@ -10,13 +10,6 @@ namespace System.IO.Abstractions.TestingHelpers
     public class MockFileSystem : IFileSystem, IMockFileDataAccessor
     {
         private readonly IDictionary<string, MockFileData> files;
-        private readonly FileBase file;
-        private readonly DirectoryBase directory;
-        private readonly IFileInfoFactory fileInfoFactory;
-        private readonly PathBase pathField;
-        private readonly IDirectoryInfoFactory directoryInfoFactory;
-        private readonly IDriveInfoFactory driveInfoFactory;
-        private readonly IFileSystemWatcherFactory fileSystemWatcherFactory;
 
         [NonSerialized]
         private readonly PathVerifier pathVerifier;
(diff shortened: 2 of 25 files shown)


jpreese:
Yeah I'm the same way, I like to have things consistent. I've just been cleaning up files as I add enhancements or bug fixes (moving the usings above the namespacing, make sure the indentation is right, etc).

We don't have any concept of standardization right now, so I wouldn't be opposed to something like an editorconfig

jpreese:
I don't really have any issues with this PR, my only worry is our current release methodology which we've already discussed and will be sticking with. 

If we accept a PR that is purely refactoring with no benefit aside from readability, the consumer will get pinged to update their references but won't be pulling anything of value. Maybe that's not a huge deal, but it just seems a little wrong, at least to me.

While I really appreciate code cleanup, I'm sort of leaning towards an approach where PRs should fix a bug or add a feature, and while we're in there doing it, perform a code clean up. 

Thoughts? #295 and #308 for some additional reading on the release strategy.

System.IO.Abstractions/DirectoryInfoWrapper.cs
```diff
@@ -11,12 +11,7 @@ public class DirectoryInfoWrapper : DirectoryInfoBase
 
         public DirectoryInfoWrapper(DirectoryInfo instance)
         {
-            if (instance == null)
-            {
-                throw new ArgumentNullException("instance");
-            }
-
-            this.instance = instance;
+            this.instance = instance ?? throw new ArgumentNullException("instance");
```
fgreinacher: `nameof(instance)`  - same for other changed files

jpreese:
Also you're going to see some merge conflicts with #315 -- were getting rid of the if block for Unix checks.

robkeim:
@jpreese after reading over #295 and #308, here's my $0.02 since you asked for it :)

I understand the potential concern of "over notifying" people about new package updates, but out of all of the bigger projects I've worked on, we've always more work than we can handle and only upgraded NuGet packages when we absolutely had to, so I wouldn't be too worried about a lot of "newer" versions of a package accumulating particularly when they're not major version updates. Have we had specific feedback before about having too many notifications?

From a contributor perspective, I'm not a fan of having to including refactoring as part of other changes for two reasons:
1. As an open source project, contributors have a limited amount of time that they're volunteering to help the project move forward. We should be welcoming any and all contributions with open arms and by having to have PRs be coupled with a feature/bug fix adds a barrier to entry which isn't needed IMO.
2. While I am definitely a fan of [opportunistic refactoring](https://martinfowler.com/bliki/OpportunisticRefactoring.html) or [the boy scout rule](https://medium.com/@biratkirat/step-8-the-boy-scout-rule-robert-c-martin-uncle-bob-9ac839778385) I think refactoring or code cleanup by itself can be valuable and decoupling refactoring/code cleanup with features and bug fixes as it makes both of them easier to review/validate individually by maintaining a single responsibility.

With that said, I'm going to step down off my soap box and fix @fgreinacher's feedback :) If you'd prefer to merge this with something else then I can wrap it up with #311 although I've got some other cleanup to do on that PR before it will be ready to merge.

TestingHelpers/MockFileSystem.cs
```diff
@@ -50,40 +43,19 @@ public MockFileSystem(IDictionary<string, MockFileData> files, string currentDir
...
-	
-        public IFileSystemWatcherFactory FileSystemWatcher
-        {
-            get { return fileSystemWatcherFactory; }
-        }
+        public IDriveInfoFactory DriveInfo { get; }
+
+        public IFileSystemWatcherFactory FileSystemWatcher { get; }
 
         public PathVerifier PathVerifier
```
jpreese: This could be
```csharp
```

to make it a bit more consistent with the new getters. Do we care about the newline between each property?

System.IO.Abstractions/DirectoryInfoWrapper.cs
robkeim: @fgreinacher fixed!

jpreese:
1. Yeah, as far as I'm aware there hasn't been anyone come forward saying we're a hindrance for too many releases. Which, this is my primary concern, is probably a non issue.

2. To resolve this what I'll usually do is ultimately have two commits to master when its time to submit my PR. One commit is purely aesthetics, refactoring, etc and then the other is the feature itself.

All said and done, I can be down for these PRs as I am a refactoring fiend.

TestingHelpers/MockFileSystem.cs
robkeim: Fixed! I'm not too familiar with the `[NonSerialized]` attribute, but could we drop the private readonly variable entirely and have:
```
[NonSerialized]
public PathVerifier PathVerifier { get; }
```

I personally like the newline between the properties, but I prefer consistency to any personal preference I have :)

jpreese: Not a huge deal just to leave it as going to be a matter of preference.. and we don't have any standards at the moment.

jpreese:
Looks good to me! 🎉

fgreinacher:
Merging - thanks for taking care!

robkeim:
My first PR merged, woo-hoo 🍾

PR #314, opened 2018-07-09
Move MockFileSystem to use ConcurrentDictionary for files

robkeim:
I prefer using ConcurrentDictionary instead of a regular Dictionary with locks when you need concurrent access.

Currently we have a global lock approach for the mock file system which doesn't mimic what real file systems are actually doing (they have much more fine grained locks). This is a naive attempt at moving the MockFileSystem to using a ConcurrentDictionary to avoid having to lock everywhere.

This change moves in that direction although isn't there 100%, because whenever I need to prevent concurrent add/deletes, I'm still locking the entire collection.

It's worth discussing whether we'd like to go in this direction and I'm open to feedback. If it is a direction we want to consider but this isn't the right implementation than we can scratch this PR and take another approach as well.

### DIFF
--- a/TestingHelpers/MockFileSystem.cs
+++ b/TestingHelpers/MockFileSystem.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System.Collections.Concurrent;
+using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
 using System.Reflection;
@@ -9,7 +10,7 @@ namespace System.IO.Abstractions.TestingHelpers
     [Serializable]
     public class MockFileSystem : IFileSystem, IMockFileDataAccessor
     {
-        private readonly IDictionary<string, MockFileData> files;
+        private readonly ConcurrentDictionary<string, MockFileData> files;
         private readonly FileBase file;
         private readonly DirectoryBase directory;
         private readonly IFileInfoFactory fileInfoFactory;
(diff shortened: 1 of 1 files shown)


fgreinacher:
TBH I don't see the value of `ConcurrentDictionary` here as long as we still need to `lock` explicitly. I'd prefer a better explicit lock, like `ReaderWriterLockSlim`.

ericnewton76:
@robertja probably right on this.  This would cause a subtle difference in behavior not exhibited by real file systems.

@fgreinacher that looks like a good solution I suppose.  I've always mostly agreed with @tathamoddie about trying to not create an in-memory filesystem, but we have to not introduce differences.

tathamoddie:
I just don't want to fall into the trap of re-writing NTFS from the ground up. At some point, the testing helpers are just going to be different to the real thing, and that's going to need to be ok.

Sidenote: I've pondered for numerous years about trying to create some kind of VHD template, mounting it with the OS without a drive letter (which I think can be done non-admin), and then redirecting calls through to that. That'd give us access to the full OS-implementation of the file system, but in a safe / containerised / disposable way. It'd be 99.999% the same as the real thing (mind the indirection), but perfectly encapsulated for testing, and would save us from rebuilding NTFS from scratch. That's the dream at least. I've never tried to actually build it.

robkeim:
Thanks @fgreinacher, @ericnewton76, and @tathamoddie for the feedback. I'll go ahead and close this PR because I agree going down the path of trying to re-implement NTFS would be never ending and there are always going to be some differences.

> Sidenote: I've pondered for numerous years about trying to create some kind of VHD template, mounting it with the OS without a drive letter (which I think can be done non-admin), and then redirecting calls through to that. That'd give us access to the full OS-implementation of the file system, but in a safe / containerised / disposable way. It'd be 99.999% the same as the real thing (mind the indirection), but perfectly encapsulated for testing, and would save us from rebuilding NTFS from scratch. That's the dream at least. I've never tried to actually build it.

This is a super cool idea! We might run into some snags in terms of cross platform and different OSes, but it would definitely give us an almost perfect mocking experience.

PR #315, opened 2018-07-09
Introduce WindowsOnlyAttribute and use it to simplify tests

fgreinacher:
@jpreese this addresses some of your comments on my fix PR. Do you think this goes in the right direction? I'm also very open for feedback on the naming.


jpreese:
Awesome! I think I like this approach better than the categories, a lot easier to run. My only big comment would be, if I recall correctly we have some tests that still check IsUnixPlatform() and add/remove test data based on the environment.

I think it would be a lot better if we werent modifying test data within the test (avoiding logic in tests), and just splitting those out into new tests that we can then just flat out skip for Linux. 

Does that make sense?

TestHelpers.Tests/MockDirectoryTests.cs
```diff
@@ -540,13 +540,9 @@ public void MockDirectory_CreMockDirectory_CreateDirectory_ShouldReturnDirectory
         }
```
jpreese: This is one example, I think there was at least one more.

```csharp
                 // These are no problems on Unix platforms
                 yield return @"..\";	       
                 yield return @"aaa\vv..\";	
                 yield return @"a..\b";	                 
```

I think I'd rather see this in another test and we could add this skip attribute

TestHelpers.Tests/SkipOnUnixAttribute.cs
```diff
@@ -0,0 +1,36 @@
+using System;
+using System.Linq;
+using NUnit.Framework;
+using NUnit.Framework.Interfaces;
+
+namespace System.IO.Abstractions.TestingHelpers.Tests
+{
+    internal static class SkipReason
+    {
+        public const string NoDrivesOnUnix = "Unix does not have the concept of Drives";
```
jpreese: I think there are a couple tests that are skipped due to Illegal characters

TestHelpers.Tests/SkipOnUnixAttribute.cs
```diff
@@ -0,0 +1,36 @@
+using System;
+using System.Linq;
+using NUnit.Framework;
+using NUnit.Framework.Interfaces;
+
+namespace System.IO.Abstractions.TestingHelpers.Tests
+{
+    internal static class SkipReason
```
jpreese: I'd probably put this in its own class. You?

fgreinacher: Yeah, will do

TestHelpers.Tests/SkipOnUnixAttribute.cs
fgreinacher: Yes, added a new reason for that

TestHelpers.Tests/MockDirectoryTests.cs
fgreinacher: 6d58566 splits this to separate tests. I'm not sure (but also not very emotional) on what I like better 🤣

jpreese: haha I think the idea is better, anyway. You typically want to avoid any sort of logic within a test. 

For the approach, I don't think I'd have a test that calls another test with different input. I was more thinking just explicitly put the input in the test and run the test, even if the arrange/act are similar or exact.

So you'd have something like 

```csharp
[Skip on Linux]
public void MockDirectoryInfo_FullName_UNCPaths_ShouldReturnNormalizedPath()
{
  var input = object array of uncpaths {} // arrange
  // act
  // assert
}
```

jpreese:
Looking great! This will be so much better than that if(linux) block everywhere.

rkoeninger:
Most minor of quibbles, but I would call it `WindowsOnly` instead, as Windows is the odd one out, and Linux, BSD, macOS, iOS, Android collectively form the norm.

And I feel that rhyming grants this a little extra legitimacy.

fgreinacher:
That’s actually a good idea @rkoeninger - makes it much clearer IMO! Will change

TestHelpers.Tests/MockDirectoryTests.cs
jpreese: If that's still a little bit confusing, let me know and I could whip up an example. I just think its better to be a little be repetitive in tests vs. trying to avoid repeating yourself. When it comes to tests, I think it's generally better to show as much intent/be expressive as possible.

fgreinacher: Yeah, that's what I also didn't like. b6a744c tries to improve on this.

TestHelpers.Tests/MockDirectoryInfoTests.cs
```diff
@@ -202,14 +202,25 @@ public static IEnumerable<object[]> MockDirectoryInfo_FullName_Data
...
+            get
+            {
+                yield return new object[] { XFS.Path(@"\\unc\folder"), XFS.Path(@"\\unc\folder") };
+                yield return new object[] { XFS.Path(@"\\unc/folder\\foo"), XFS.Path(@"\\unc\folder\foo") };
+            }
+        }
+
+        [TestCaseSource("MockDirectoryInfo_FullName_Data_WindowsOnly")]
+        [WindowsOnly(WindowsSpecifics.UNCPaths)]
+        public void MockDirectoryInfo_FullName_ShouldReturnNormalizedPath_WindowsOnly(string directoryPath, string expectedFullName)
```
jpreese: Another case of a test calling a test.

TestHelpers.Tests/MockDirectoryTests.cs
jpreese: Woo! I just saw one more, and then I think we're good. Thanks for this! 🎉

jpreese:
I'll merge this when the build starts passing. I'm excited for this one!

jpreese:
Woo! Awesome 🎉

fgreinacher:
Done 🎉 Please squash this crappy commits when merging 🤖

PR #316, opened 2018-07-09
Implementing MockDirectoryInfo.CreationTime

rkoeninger:
Implemented `MockDirectoryInfo.CreationTime` by having the property look up the `MockFileData` for the directory on the instance field `mockFileDataAccessor`, like other properties and methods in the same class do.

Added unit tests to confirm date is assignable, retained and correctly translates between local time and UTC, so it can be assigned in either time zone and retrieved in either time zone and stay accurate.

**Potential Bug Identified**

The test `MockDirectoryInfo_CreationTime_ShouldWorkIfPathIsActuallyAFile` is currently failing. In the BCL, it is possible to create a `new DirectoryInfo` for a path to a file and the `CreationTime` is accessible and accurate, even though it's a `DirectoryInfo` pointing to a file (on Windows in .NET Framework anyway).

I wrote a sanity test to confirm the BCL exhibits this behavior:

```csharp
[Test]
public void WorksInBCL()
{
    var temp = Path.GetTempFileName();
    Console.WriteLine(temp); // "C:\Users\username\AppData\Local\Temp\tmpE451.tmp"
    Console.WriteLine();
    var dir = new DirectoryInfo(temp);
    Console.WriteLine(dir.FullName); // "C:\Users\username\AppData\Local\Temp\tmpE451.tmp"
    Console.WriteLine(dir.CreationTime); // shows approx the current time
}
```

This doesn't work in System.IO.Abstractions because the `MockFileSystem` disambiguates file and directory paths by looking for a trailing slash. `MockDirectoryInfo` adds a ensures there is a trailing slash on the path so `MockFileSystem` can make this distinction.

A separate issue can be opened for this if we want to deal with it. The test currently has the `Ignore` attribute. We could leave it there as a reminder or just remove it.

### DIFF
--- a/TestHelpers.Tests/MockDirectoryInfoTests.cs
+++ b/TestHelpers.Tests/MockDirectoryInfoTests.cs
@@ -348,7 +348,7 @@ namespace System.IO.Abstractions.TestingHelpers.Tests
             Assert.AreEqual(currentTime.ToUniversalTime(), fileSystem.DirectoryInfo.FromDirectoryName(directoryPath).CreationTimeUtc);
         }
 
-        [Test]
+        [Test, Ignore("Failing because MockDirectoryInfo adds trailing slash")]
         public void MockDirectoryInfo_CreationTime_ShouldWorkIfPathIsActuallyAFile()
         {
             // Arrange
--- a/TestHelpers.Tests/MockDirectoryInfoTests.cs
+++ b/TestHelpers.Tests/MockDirectoryInfoTests.cs
@@ -283,5 +283,85 @@ namespace System.IO.Abstractions.TestingHelpers.Tests
             // Assert
             Assert.AreEqual(directoryPath, str);
         }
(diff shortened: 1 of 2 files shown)


jpreese:
@rkoeninger thanks for the PR. At first muster this looks like a duplicate of #311. I can look into this a bit deeper in a little bit.

rkoeninger:
@jpreese Oh yeah, #311 opened while I was working on this one. And seeing as it's more comprehensive, it should replace this one. Feel free to close this if so.

Should still create an issue for the trailing slash if not already noted or being addressed.

jpreese:
I'm not aware of any open issues for this, I'd have to test it on my local to see what's going on. If you want to open a ticket with the behavior differences, I can look into it more.

I'm going to close this in favor of the other PR, but you're more than welcome to join in on the conversation in #311.

PR #206, opened 2017-03-04
Fix/issue 201

ericnewton76:
Deals with Issue #201 by creating a deferred execution enumerable.

PR #317, opened 2018-07-09
Make EnumerateFiles lazy

### LINKED ISSUE #201: DirectoryInfoWrapper does not support lazy file enumeration
The following methods of class [DirectoryInfoWrapper ](https://github.com/tathamoddie/System.IO.Abstractions/blob/master/System.IO.Abstractions/DirectoryInfoWrapper.cs)returns results as an IEnumerable which can be lazily evaluated somewhat, but current code returns results as an array which has to be fully populated before it can return.

- [EnumerateFileSystemInfos()](https://github.com/tathamoddie/System.IO.Abstractions/blob/master/System.IO.Abstractions/DirectoryInfoWrapper.cs#L151)
- [EnumerateFileSystemInfos(string searchPattern)](https://github.com/tathamoddie/System.IO.Abstractions/blob/master/System.IO.Abstractions/DirectoryInfoWrapper.cs#L156)
- [EnumerateFileSystemInfos(string searchPattern, SearchOption searchOption)](https://github.com/tathamoddie/System.IO.Abstractions/blob/master/System.IO.Abstractions/DirectoryInfoWrapper.cs#L159)

In some cases, especially when you are working with remote file systems this difference is crucial from performance stand of point.

As far as I can see problem is in the implementation of [Converters ](https://github.com/tathamoddie/System.IO.Abstractions/blob/master/System.IO.Abstractions/Converters.cs)class, namely all methods of this class returns results as an array instead of IEnumerable

For more information please see here:

1. https://msdn.microsoft.com/en-us/library/dd383571.aspx
2. http://stackoverflow.com/questions/5669617/what-is-the-difference-between-directory-enumeratefiles-vs-directory-getfiles

fgreinacher:
Fixes #201
Supersedes #206

### DIFF
--- a/TestHelpers.Tests/ConvertersTests.cs
+++ b/TestHelpers.Tests/ConvertersTests.cs
@@ -1,0 +1,54 @@
+using System.Collections;
+using System.Collections.Generic;
+using NUnit.Framework;
+
+namespace System.IO.Abstractions.TestingHelpers.Tests
+{
+    [TestFixture]
+    public class ConvertersTests
+    {
+        private sealed class CrashingEnumerable<T> : IEnumerable<T>
+        {
+            private sealed class CrashingEnumerator : IEnumerator<T>
+            {
+                object IEnumerator.Current => throw new NotSupportedException();
+
+                public T Current => throw new NotSupportedException();
+
+                public bool MoveNext() { throw new NotSupportedException(); }
+
+                public void Reset() { }
+
+                public void Dispose() { }
+            }
+            
+            public IEnumerator<T> GetEnumerator() => new CrashingEnumerator();
+
+            IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
+        }
+
+        [Test]
+        public void WrapFileSystemInfos_with_IEnumerable_is_lazy()
+        {
+            var crashingFileSystemInfos = new CrashingEnumerable<FileSystemInfo>();
+        
+            Assert.DoesNotThrow(() => Converters.WrapFileSystemInfos(crashingFileSystemInfos));
+        }
+
+        [Test]
+        public void WrapFiles_with_IEnumerable_is_lazy()
+        {
+            var crashingFileInfos = new CrashingEnumerable<FileInfo>();
+        
+            Assert.DoesNotThrow(() => Converters.WrapFiles(crashingFileInfos));
+        }
+        [Test]
+        public void WrapDirectories_with_IEnumerable_is_lazy()
+        {
+            var crashingDirectoryInfos = new CrashingEnumerable<DirectoryInfo>();
+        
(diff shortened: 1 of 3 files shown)


TestHelpers.Tests/ConvertersTests.cs
```diff
@@ -0,0 +1,54 @@
+using System.Collections;
+using System.Collections.Generic;
+using NUnit.Framework;
+
+namespace System.IO.Abstractions.TestingHelpers.Tests
+{
+    [TestFixture]
+    public class ConvertersTests
```
jpreese: Alright, I think I understand everything that's going on here, so this is my opinion.

To me, since the Converters are internal to the System.IO.Abstractions, and the consumers of our library will never directly interact with the Converters. I don't think we should directly test them. 

I think we should add lazy tests, but they should be against the public interface, EnumerateFiles(). That way we're calling it the same way our end users would be. Just because our Converters work and are lazy, doesn't mean the public interface always will be. Someone could technically come along and gut the EnumerateFiles() method and replace the implementation with something else. Our tests would still pass, but the method would no longer be lazy.

Thoughts?

jpreese:
Thanks for working on this @fgreinacher. I added some comments. Let me know your thoughts!

TestHelpers.Tests/ConvertersTests.cs
fgreinacher: Hm, I agree with your thoughts regarding testing private methods, but I have a different opinion for internal classes. In my opinion the "unit" of a unit test should be somehow self-contained and testable in isolation and does not not have to be the whole API. 

Testing the lazyness at the API level for me would bring unnecessary complexity to the tests without any clear value.

Regarding "Just because our Converters work and are lazy, doesn't mean the public interface always will be" and "Someone could technically come along and gut the EnumerateFiles() method and replace the implementation with something else. Our tests would still pass, but the method would no longer be lazy.": To mitigate exactly these kinds of problems you have the concept of integration tests. Those verify that a bunch of (unit tested) parts are composed in the right way.

jpreese: I don't think it would necessarily add complexity. We would have a test that simply calls EnumerateFiles with a test name of something like EnumerateFiles_ByDefault_IsLazy(), and that would validate that the method truley is lazy. 

We have a product gap right now where EnumerateFiles is not lazy. Tests prove that a gap/bug has been fixed, but we don't have a test for EnumerateFiles being lazy. We've added code to support it, but we're not proving the method is now lazy. I think thats my biggest catchup right now is that we're not proving what we've done.

Regardless, it's just a matter of opinion so if you're OK with it, then I am. 👍

fgreinacher: Yeah, that's one of the very few places where we do more than just forwarding calls to `System.IO`. So maybe not worth putting too much time in it. I'll merge this as is now - we might revisit any time later :)

PR #322, opened 2018-07-12
Don't allow files and directories with the same name

### LINKED ISSUE #208: MockFileSystem should not require trailing slash for directories
Create a `MockFileSystem` with a directory that does not have a trailing slash:

````c#
var fs = new MockFileSystem(new Dictionary<string, MockFileData>
    {
        { @"C:\test1\test1.txt", MockFileData.NullObject },
        { @"C:\test2", new MockDirectoryData() }
    });

fs.File.Copy(@"C:\test1\test1.txt", @"C:\test2\test2.txt");
````

Running this code produces the following exception:

````
System.IO.DirectoryNotFoundException : Could not find a part of the path 'C:\test2\test2.txt'.
   at System.IO.Abstractions.TestingHelpers.MockFile.Copy(String sourceFileName, String destFileName, Boolean overwrite)
   at System.IO.Abstractions.TestingHelpers.MockFile.Copy(String sourceFileName, String destFileName)
````

Adding a trailing slash to the folder definition fixes the problem:

````c#
...
        { @"C:\test2\", new MockDirectoryData() }
...
````

Now it runs successfully.

Using a trailing slash should not be necessary (`MockDirectoryData` should suffice to identify this as a folder).

### LINKED ISSUE #223: MockDirectory.SetCreationTime throws NullReferenceException
Attempting to set the creation date of a mock directory results in a null reference exception. Here is an NUnit test that demonstrates the issue:
```csharp
[TestCase(true)]
[TestCase(false)]
public void TestDirectorySetCreationTime(bool realFileSystem)
{
  const string PATH = @"C:\test123";
  IFileSystem fs = realFileSystem ?
    (IFileSystem) new FileSystem() :
    (IFileSystem) new MockFileSystem();
  fs.Directory.CreateDirectory(PATH);
  fs.Directory.SetCreationTime(PATH, DateTime.Now);  // throws if fs is a MockFileSystem
  fs.Directory.Delete(PATH);
}
```
More information about the exception:
```
System.NullReferenceException was unhandled by user code
  HResult=-2147467261
  Message=Object reference not set to an instance of an object.
  Source=System.IO.Abstractions.TestingHelpers
  StackTrace:
       at System.IO.Abstractions.TestingHelpers.MockFile.SetCreationTime(String path, DateTime creationTime)
       at System.IO.Abstractions.TestingHelpers.MockDirectory.SetCreationTime(String path, DateTime creationTime)
  InnerException:
```
I am new to GitHub (and Git itself), but will attempt a fix.

### LINKED ISSUE #318: Possible to create file and directory with the same name
**Main Issue**

This test passes, and it really shouldn't:

```csharp
[Test]
public void CreateSamePathAsDirectoryAndFile()
{
    fileSystem.AddDirectory(@"C:\");
    var path = @"C:\path";
    fileSystem.File.Create(path); // Create path as File
    Console.WriteLine(fileSystem.File.Exists(path)); // True
    fileSystem.Directory.CreateDirectory(path); // Create path as Directory
    Console.WriteLine(fileSystem.Directory.Exists(path)); // True
    Console.WriteLine(fileSystem.Path.GetFullPath(path)); // C:\path
}
```

You can't create a directory and a file with the same name in the same parent on either Windows or Unix (and probably most other OS/file systems), but `MockFileSystem` allows it.

`MockFileSystem` contains a `IDictionary<string, MockFileData>` with all the file system entries. It uses a trailing slash on all directories, so they have distinct keys from files. Should probably remove trailing slashes from all keys and do a check if a file system entry is a file or directory before doing a file overwrite or directory move operation.

The `AllFiles` and `AllDirectories` properties on `MockFileSystem` don't actually depend on the trailing slash, they filter using the `IsDirectory` property on the `MockFileData`.

---

**Secondary Issue**

(This is what I was originally going to report, less of an issue, and might be resolved just by fixing the above)

While working on what turned out to be a redundant PR ( #316 ), I discovered this:

The test `MockDirectoryInfo_CreationTime_ShouldWorkIfPathIsActuallyAFile` (EDIT: a test that was part of that PR) is currently failing. In the BCL, it is possible to create a new `DirectoryInfo` for a path to a file and the `CreationTime` is accessible and accurate, even though it's a `DirectoryInfo` pointing to a file (on Windows in .NET Framework anyway).

I wrote a sanity test to confirm the BCL exhibits this behavior:

```csharp
[Test]
public void WorksInBCL()
{
    var temp = Path.GetTempFileName();
    Console.WriteLine(temp); // "C:\Users\username\AppData\Local\Temp\tmpE451.tmp"
    var dir = new DirectoryInfo(temp);
    Console.WriteLine(dir.FullName); // "C:\Users\username\AppData\Local\Temp\tmpE451.tmp"
    Console.WriteLine(dir.CreationTime); // shows the time the test was run
}
```

This doesn't work in `System.IO.Abstractions` because the `IDictionary<string, MockFileData>` in `MockFileSystem` uses the path strings as keys, with trailing slashes on the directories, no trailing slashes on files. `MockDirectory` and `MockDirectoryInfo` ensure there is a trailing slash on the path. If you try to create a `MockDirectoryInfo` pointing at a file, the trailing slash is added and it no longer points at the file.

rkoeninger:
Fixes #208
Fixes #223
Fixes #318

Effects of this PR:
* Can no longer have a file and a directory with the same name exist at the same time.
* Trailing slashes are acceptable but not required.
* Directories can be created, removed and queried with a trailing slash without any difference in behavior.
* Tests that assumed directories would have trailing slashes have been fixed.
* Keys in the dictionary in `MockFileSystem` no longer have trailing slashes unless:
  * It's a Windows drive root.
  * Or the Unix file system root.
* Files enumerated by `Directory.GetFiles()` won't have trailing slashes.
* Incidentally, `UnixOnlyAttribute`/`UnixSpecifics` were added to mirror `WindowsOnlyAttribute`/`WindowsSpecifics`.
* 14 unit tests added to cover the `TrimSlashes` helper, increase coverage of existing functionality and confirm that issues #208 and #233 have been addressed as well.


TestHelpers.Tests/StringExtensionsTests.cs
```diff
@@ -56,5 +56,40 @@ public void SplitLines_InputWithOneLine_ShouldReturnOnlyOneLine()
             Assert.That(result, Is.EquivalentTo(expected));
         }
 
+        [Test]
+        public void CleanPath_DriveRoot_PreserveTrailingSlash()
+        {
+            Assert.AreEqual(@"c:\", @"c:\".CleanPath());
```
jpreese: It may seem a little silly / overly verbose, but I think I'd prefer if this was broken up into three lines for AAA (no comments for each section though).

You'll also need XFS.Path for Linux shenanigans.

TestingHelpers/MockFileSystem.cs
```diff
@@ -92,22 +92,29 @@ private string GetPathWithCorrectDirectoryCapitalization(string fullPath)
...
-
         public MockFileData GetFile(string path)
         {
             path = FixPath(path);
             return GetFileWithoutFixingPath(path);
         }
 
+        private void SetEntry(string path, MockFileData mockFile)
+        {
+            path = FixPath(path, true);
```
jpreese: I'm personally a fan of pure/functional methods where possible, so I think I'd rather see this be immutable.

jpreese:
Thanks for working on this @rkoeninger. I'm really curious what you come up for this one. In the spirit of rapid feedback, I left a couple comments.

TestingHelpers/MockFileSystem.cs
rkoeninger: You mean overwriting the value of the local variable? I'm not a fan of that either.

jpreese:
I have always used a docker container to text Linux tests locally. You can mount the System.IO.Abstractions source folder and run dotnet test.

rkoeninger:
@jpreese About that test (0c6879a): I wasn't questioning the existence of the test, but the way it was written. It started to fail after I made my changes, I believe because it was depending on the ability to have a file and directory with the same name exist at the same time. That commit contains the one-line fix that fixes what may have just been a typo.

I get worried when I have to change tests to make them pass, and they're no changes I expected to have to make.

jpreese:
For that test, yeah I was agreeing with the name and how it was written.

1. Add a file at c:\temp\file.txt
2. Add a directory at c:\temp2\file.txt
3. Move c:\temp\file.txt to c:\temp2\file.txt
4. Validate the move overwrote c:\temp2\file.txt

So if I'm understanding this right, the fault here is that we're doing .AddDirectory and not adding a File. The test is validating that we can _replace_ a file, and the way its written now, does not do that since destinationFolder is empty.

The test you've written validates that we can Move a file (which I think we need, because for some reason we don't have it).

There should be a test for move to an empty directory, and a test for a move to a source that already exists.

rkoeninger:
@jpreese Ok, so the original test was split into 3 tests, and added `Copy`/`Move` tests for `MockFile`.

rkoeninger:
I think this is pretty much ready. Does anyone have any other comments/concerns?

TestHelpers.Tests/StringExtensionsTests.cs
```diff
@@ -2,6 +2,8 @@
 
 namespace System.IO.Abstractions.TestingHelpers.Tests
 {
+    using XFS = MockUnixSupport;
+
     [TestFixture]
     public class StringExtensions
```
fgreinacher: Add Tests suffix?

TestingHelpers/StringExtensions.cs
```diff
@@ -55,5 +55,33 @@ public static string Replace(this string source, string oldValue, string newValu
...
+            if (string.IsNullOrEmpty(path))
+            {
+                return path;
+            }
+
+            var trimmed = path.TrimEnd(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar);
+
+            if (MockUnixSupport.IsUnixPlatform()
+                && (path[0] == Path.DirectorySeparatorChar || path[0] == Path.AltDirectorySeparatorChar)
+                && trimmed == "")
```
fgreinacher: string.Empty?

TestHelpers.Tests/StringExtensionsTests.cs
rkoeninger: Sure

TestHelpers.Tests/MockFileSystemTests.cs
```diff
@@ -243,13 +243,24 @@ public void MockFileSystem_AllNodes_ShouldReturnAllNodes()
...
+
+        [Test]
+        [TestCase(@"C:\path")]
+        [TestCase(@"C:\path\")]
+        public void MockFileSystem_AddDirectory_TrailingSlashAllowedButNotRequired(string path)
+        {
+            var fileSystem = new MockFileSystem();
+            var path2 = XFS.Path(path);
+            fileSystem.AddDirectory(path2);
+            fileSystem.FileExists(path2);
```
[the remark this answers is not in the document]
rkoeninger: Whoops, assert added

TestingHelpers/StringExtensions.cs
rkoeninger: Eh, I prefer `""` to `string.Empty`. 2 chars vs 12.

fgreinacher:
@jpresse for a last pair of eyes 👀

rkoeninger:
I had copy/paste/modified the `MockFileInfo` tests for `Copy`/`Move` over to `MockFile` because I thought they were missing, but just noticed while looking at the next issue that they already existed in `MockFileCopyTests.cs` and `MockFileMoveTests.cs`.

Minor improvement would be to name the files like `MockFile.Copy.cs` so they get sorted in proper alpha order. Or in nested classes, not sure what the best approach would be.

TestHelpers.Tests/MockDirectoryTests.cs
```diff
@@ -95,7 +95,7 @@ public void MockDirectory_GetFiles_ShouldFilterByExtensionBasedSearchPatternWith
             var fileSystem = SetupFileSystem();
             fileSystem.AddFile(additionalFilePath, new MockFileData(string.Empty));
             fileSystem.AddFile(XFS.Path(@"c:\a\a\c.gifx.xyz"), new MockFileData(string.Empty));
-            fileSystem.AddFile(XFS.Path(@"c:\a\a\c.gifx\xyz"), new MockFileData(string.Empty));
+            fileSystem.AddFile(XFS.Path(@"c:\a\a\c.gifz\xyz"), new MockFileData(string.Empty));
```
jpreese: Why did we change the test data for this?

TestHelpers.Tests/MockFileCopyTests.cs
```diff
@@ -96,14 +96,9 @@ public void MockFile_Copy_ShouldThrowArgumentNullExceptionWhenSourceIsNull_Param
         }
 
         [Test]
+        [WindowsOnly(WindowsSpecifics.StrictPathRules + "; Path.GetInvalidChars() does not return anything on Mono")]
```
jpreese: I think the StrictPathRules const is good enough here (and the others)

jpreese:
Wel, I started a review but then realized this PR is pretty large.. and I'm a little uneasy with how many tests were changing. It's going to take me a bit to look this over.

TestHelpers.Tests/MockDirectoryTests.cs
rkoeninger: That test depended on the ability to have a file and directory with the same name. 9fad978a0a4d185764000351467e60a37bfd55b1

fgreinacher:
@jpreese :shipit: ?

rkoeninger:
@fgreinacher @jpreese :ship: ?

fgreinacher:
OK, I’m gonna merge this now - we have no negative feedback except the amount of tests this impacts - but as I mentioned above all the changes look unproblematic to me.

Thanks again for working on this @rkoeninger!

rkoeninger:
@fgreinacher No prob, thanks for getting this merged

PR #213, opened 2017-03-07
Fix #209: Add AllNodes

cmeeren:
Add AllNodes property with tests, as well as tests for AllPaths, AllNodes, AllFiles, and AllDirectories.

PR #323, opened 2018-07-12
Implement AllNodes

### LINKED ISSUE #209: Add AllLeaves to MockFileSystem
My testing would be simplified if I had access to an array of only the "leaves" of the file system (all the files, as well as empty folders).

For example, if `AllPaths` contain

````
C:\
C:\foo\
C:\bar\
C:\bar\baz\
C:\bar\baz\test1.txt
C:\bar\baz\test2.txt
````

then I would expect `AllLeaves` to contain

````
C:\foo\
C:\bar\baz\test1.txt
C:\bar\baz\test2.txt
````

Here's a quick and dirty implementation I made for testing, using LINQ:

````c#
private string[] AllLeaves(MockFileSystem fs)
{
    Func<string, bool> isNotStartOfAnotherPath =
        path => !fs.AllPaths.Any(otherPath => otherPath.StartsWith(path) && otherPath != path);

    return fs.AllPaths.Where(isNotStartOfAnotherPath).ToArray();
}
````

jpreese:
Replaces PR #213 
Fixes #209


fgreinacher:
👍

PR #325, opened 2018-07-14
Make package/assembly/directory/project names match

### LINKED ISSUE #271: Assembly name of TestingHelpers
At least in the current `System.IO.Abstractions.TestingHelpers-2.1.0.178` the assembly itself is only named `TestingHelpers.dll`, i.e. no `System.IO.Abstractions` prefix. This is a inconsistency with the main `System.IO.Abstractions` and yells for name clashes: "TestingHelpers.dll" is quite generic.

Please consider naming the assembly `System.IO.Abstractions.TestingHelpers.dll` in future versions.

fgreinacher:
This PR makes package/assembly/directory/project names match.

Fixes #271


TestHelpers.Tests/TestHelpers.Tests.ncrunchproject
```diff
@@ -1,20 +0,0 @@
-<ProjectConfiguration>
```
fgreinacher: @jpreese @tathamoddie Do you use NCrunch?

jpreese: I do, yes. Though I'm not convinced it needs to be in source control.

robkeim: If we're going to remove these I think it makes sense to update the `.gitignore` to ensure they don't accidentally work their way back into the repro.

It looks like we already have a section for NCrunch but I don't know why this line didn't cause these files to be ignored:
https://github.com/System-IO-Abstractions/System.IO.Abstractions/blob/e6ec3d48138fbf708c593953b3a395c23f28a84b/.gitignore#L75

Maybe that was updated after these files were already added?

fgreinacher: Ah, that’s why Git wants to remove them :)

fgreinacher:
Any objections on merging this?

jpreese:
This should be OK. As long as we're not making any changes to the namespace we shouldn't be breaking anything.

jpreese:
I think when the conflicts are resolved we're all good 🎉

PR #327, opened 2018-07-16
PathVerifier should check specifically for invalid use of volume separator and tolerate alt dir separator

### LINKED ISSUE #283: MockFileSystem's File.ReadAllText throws when given a path using forward slashes
Test case:
```csharp
{
    [@"C:\test.txt"] = new MockFileData("content")
});

Assert.IsTrue(fileSystem.Directory.Exists("C:/")); // passes
Assert.IsTrue(fileSystem.File.Exists("C:/test.txt")); // passes

// throws System.ArgumentException: 'Illegal characters in path.'
Assert.AreEqual("content", fileSystem.File.ReadAllText("C:/test.txt"));
```
From what I can tell, the root cause is `PathVerifier.ExtractFileName` not considering `Path.AltDirectorySeparatorChar`. This results in `IsLegalAbsoluteOrRelative` looking for invalid filename chars (which includes `':'`, `'\\'`, and `'/'`) in the full path.

### LINKED ISSUE #320: File.Copy and File.Move Throw Incorrect Exception
When passing in a path that contains illegal characters to either of these methods, a NotSupportedException should be thrown. Currently an ArgumentException is thrown.

Our tests inside of MockFileCopyTests.cs and MockFileMoveTests.cs are named with "...ShouldThrowNotSupportedException..." but in reality, the assertion is checking an Argument exception.

https://msdn.microsoft.com/en-us/library/system.io.file.move(v=vs.110).aspx

```csharp
      [Test]
      public void Move_InvalidCharacters_ThrowsSameExceptionAsSystemIO()
      {

        TestDelegate systemIO = () => File.Move(@"C:\a:b", @"C:\c:d");
        TestDelegate systemAbstraction = () => fileSystem.File.Move(XFS.Path(@"C:\a:b"), XFS.Path(@"C:\c:d"));

        Assert.Throws<NotSupportedException>(systemIO);
        Assert.Throws<NotSupportedException>(systemAbstraction);
      }
```
(obviously we're not going to test against System.IO in our test suite, but an easy sanity check if you're looking for one)

rkoeninger:
Fixes #283
Fixes #320

^ Lots of information on the issue (#320) about how things are handled in Unix, Mono and the .NET Framework reference implementation of `File.Copy`/`File.Move`.

* `PathVerifier.IsLegalAbsoluteOrRelative` now throws `NotSupportedException` on Windows only when the path contains an invalid use of the drive separator char.
* Renamed 8 tests to mention `ArgumentException` instead of `NotSupportedException`.
* Added 8 tests for invalid drive separator specifically.
* Updated 4 Windows only tests to test invalid path chars other than drive separator.


TestHelpers.Tests/MockFileCopyTests.cs
```diff
@@ -191,6 +197,64 @@ public void MockFile_Copy_ShouldThrowNotSupportedExceptionWhenTargetFileNameCont
             }
         }
 
+        [Test]
+        [WindowsOnly(WindowsSpecifics.StrictPathRules + "; Mono does not raise this exception")]
```
fgreinacher: Why not `WindowsSpecifics.Drives`?

TestHelpers.Tests/MockFileCopyTests.cs
```diff
@@ -107,7 +107,10 @@ public void MockFile_Copy_ShouldThrowNotSupportedExceptionWhenSourceFileNameCont
             var destFilePath = XFS.Path(@"c:\something\demo.txt");
             var fileSystem = new MockFileSystem();
 
-            foreach (var invalidChar in fileSystem.Path.GetInvalidFileNameChars().Where(x => x != fileSystem.Path.DirectorySeparatorChar))
+            foreach (var invalidChar in fileSystem.Path.GetInvalidFileNameChars()
+                .Where(x => x != fileSystem.Path.DirectorySeparatorChar
+                    && x != fileSystem.Path.AltDirectorySeparatorChar
+                    && x != fileSystem.Path.VolumeSeparatorChar))
```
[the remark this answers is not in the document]
rkoeninger: Added comments explaining why they're excluded and moved to local `char[]` show I could just use `.Except()`.

TestHelpers.Tests/MockFileCopyTests.cs
rkoeninger: Why not indeed. Used `Drives`.

TestingHelpers/PathVerifier.cs
```diff
@@ -28,6 +28,16 @@ public void IsLegalAbsoluteOrRelative(string path, string paramName)
                 throw new ArgumentException(StringResources.Manager.GetString("THE_PATH_IS_NOT_OF_A_LEGAL_FORM"), paramName);
             }
 
+            if (!MockUnixSupport.IsUnixPlatform())
+            {
+                if (path.Length > 0 && path[0] == Path.VolumeSeparatorChar
+                    || path.Length > 1 && path[1] == Path.VolumeSeparatorChar && !char.IsLetter(path[0])
+                    || path.LastIndexOf(Path.VolumeSeparatorChar) > 1)
+                {
```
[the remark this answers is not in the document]
rkoeninger: Moved to `private static` method and simplified expression. Almost added a comment, but I think the new code doesn't need it: `lastVolSepIndex == -1 || lastVolSepIndex == 1 && char.IsLetter(path[0])`

TestHelpers.Tests/MockFileCopyTests.cs
```diff
@@ -179,7 +182,10 @@ public void MockFile_Copy_ShouldThrowNotSupportedExceptionWhenTargetFileNameCont
             var sourceFilePath = XFS.Path(@"c:\something\demo.txt");
             var fileSystem = new MockFileSystem();
 
-            foreach (var invalidChar in fileSystem.Path.GetInvalidFileNameChars().Where(x => x != fileSystem.Path.DirectorySeparatorChar))
+            foreach (var invalidChar in fileSystem.Path.GetInvalidFileNameChars()
+                .Where(x => x != fileSystem.Path.DirectorySeparatorChar
+                    && x != fileSystem.Path.AltDirectorySeparatorChar
+                    && x != fileSystem.Path.VolumeSeparatorChar))
```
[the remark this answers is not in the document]
rkoeninger: Did same here

TestHelpers.Tests/MockFileCopyTests.cs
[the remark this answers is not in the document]
rkoeninger: Also handled same case in Move tests

TestHelpers.Tests/MockFileCopyTests.cs
```diff
@@ -191,6 +197,64 @@ public void MockFile_Copy_ShouldThrowNotSupportedExceptionWhenTargetFileNameCont
...
+        [Test]
+        [WindowsOnly(WindowsSpecifics.StrictPathRules + "; Mono does not raise this exception")]
+        public void MockFile_Copy_ShouldThrowNotSupportedExceptionWhenDestinationPathContainsInvalidDriveLetter()
+        {
+            var sourcePath = XFS.Path(@"C:\something\demo.txt");
+            var destinationFolder = XFS.Path(@"C:\elsewhere");
+            var badDestinationPath = XFS.Path(@"^:\elsewhere\demo.txt");
+            var fileSystem = new MockFileSystem();
+            fileSystem.AddFile(sourcePath, new MockFileData("1"));
+            fileSystem.AddDirectory(destinationFolder);
```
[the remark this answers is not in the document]
rkoeninger: The tests are necessary? They're the only place the `NotSupportedException` throws are tested, which is the impetus for the PR.

Specific lines are necessary?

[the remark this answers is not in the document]
fgreinacher: I meant just the line above (and the variable), where you add some directory that doesn’t seem to be necessary for the test. Sorry for being unclear on this!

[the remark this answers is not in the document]
rkoeninger: Ah, Ok. I re-visited all 8 of those tests and found that creating the source file and destination folder was not necessary to produce the error, so I removed those parts. They just create a file system and attempt the move/copy and assert the `NotSupportedException` is raised.

rkoeninger:
@fgreinacher Applied suggested changes. Made code simpler and easier to understand. Thanks.

TestHelpers.Tests/MockFileCopyTests.cs
```diff
@@ -106,8 +106,19 @@ public void MockFile_Copy_ShouldThrowNotSupportedExceptionWhenSourceFileNameCont
 
             var destFilePath = XFS.Path(@"c:\something\demo.txt");
             var fileSystem = new MockFileSystem();
+            var excludeChars = new []
```
jpreese: This list, comments and all, is introduced four times throughout the PR. I think there's value in pulling this out into a static constant helper file that we can just pull from.

TestHelpers.Tests/MockFileCopyTests.cs
```diff
@@ -120,7 +131,7 @@ public void MockFile_Copy_ShouldThrowNotSupportedExceptionWhenSourceFileNameCont
         }
 
         [Test]
-        public void MockFile_Copy_ShouldThrowNotSupportedExceptionWhenSourcePathContainsInvalidChars_Message()
+        public void MockFile_Copy_ShouldThrowArgumentExceptionWhenSourcePathContainsInvalidChars_Message()
         {
             if (XFS.IsUnixPlatform())
```
jpreese: While not necessarily part of the PR, we're touching the line right above it. It'd be nice to push forward and get rid of these if() checks and use the attribute.

TestHelpers.Tests/MockFileCopyTests.cs
```diff
@@ -120,7 +131,7 @@ public void MockFile_Copy_ShouldThrowNotSupportedExceptionWhenSourceFileNameCont
         }
 
         [Test]
-        public void MockFile_Copy_ShouldThrowNotSupportedExceptionWhenSourcePathContainsInvalidChars_Message()
+        public void MockFile_Copy_ShouldThrowArgumentExceptionWhenSourcePathContainsInvalidChars_Message()
```
jpreese: While not necessarily part of the PR, we're touching the line right above it. It'd be nice to push forward and get rid of these if() checks and use the attribute.

TestHelpers.Tests/MockFileCopyTests.cs
```diff
@@ -191,6 +213,50 @@ public void MockFile_Copy_ShouldThrowNotSupportedExceptionWhenTargetFileNameCont
...
+        {
+            var badSourcePath = XFS.Path(@"0:\something\demo.txt");
+            var destinationPath = XFS.Path(@"C:\elsewhere\demo.txt");
+            var fileSystem = new MockFileSystem();
+
+            Assert.Throws<NotSupportedException>(() => fileSystem.File.Copy(badSourcePath, destinationPath));
+        }
+
+        [Test]
+        [WindowsOnly(WindowsSpecifics.Drives)]
```
jpreese: Would this necessarily be a Windows only feature? I may need to look into it more, but the C:\ path will just be converted to a unix path, along the lines of /tmp/some/dir, and I would think /elsewhere:/demo.txt would be a valid case and something to test against a Unix environment.

Tests verify behavior, skipping them on an environment doesn't prevent the behavioral changes from going through, they just make us blind to the side effects.

TestHelpers.Tests/MockFileReadAllBytesTests.cs
```diff
@@ -47,5 +47,15 @@ public void MockFile_ReadAllBytes_ShouldThrowFileNotFoundExceptionIfFileDoesNotE
...
         }
+
+        [Test]
+        public void MockFile_ReadAllBytes_ShouldTolerateAltDirectorySeparatorInPath()
+        {
+            var fileSystem = new MockFileSystem();
+
+            fileSystem.AddFile(XFS.Path(@"C:\test.txt"), new MockFileData("content"));
+
+            Assert.AreEqual("content", fileSystem.File.ReadAllText(XFS.Path("C:/test.txt")));
```
jpreese: The name of the test is ReadAllBytes, but we're calling ReadAllText. It also looks like were hardcoding the forward slash. If AltDirectorySeparator were to change, this test might not make anymore sense. If you want to specifically test forward slash, I'd probably prefer calling out the way that were honestly just testing forwards and backwards slashes.

TestHelpers.Tests/MockFileCopyTests.cs
```diff
@@ -191,6 +213,50 @@ public void MockFile_Copy_ShouldThrowNotSupportedExceptionWhenTargetFileNameCont
...
 
+        [Test]
+        [WindowsOnly(WindowsSpecifics.Drives)]
+        public void MockFile_Copy_ShouldThrowNotSupportedExceptionWhenSourcePathContainsInvalidUseOfDriveSeparator()
+        {
+            var badSourcePath = XFS.Path(@"C::\something\demo.txt");
+            var destinationPath = XFS.Path(@"C:\elsewhere\demo.txt");
+            var fileSystem = new MockFileSystem();
+
+            Assert.Throws<NotSupportedException>(() => fileSystem.File.Copy(badSourcePath, destinationPath));
```
jpreese: Small nit, but it's generally preferred to split up the Act and Assert steps.

```csharp
TestDelegate action = () => fileSystem.File.Copy(badSourcePath, destinationPath);

Assert.Throws<NotSupportedException>(action);
```

Makes the debugging and readability story a bit better.

TestHelpers.Tests/MockFileCopyTests.cs
rkoeninger: According to the docs for [Path.VolumeSeparatorChar](https://msdn.microsoft.com/en-us/library/system.io.path.volumeseparatorchar(v=vs.110).aspx):

> The value of this field is a colon (:) on Windows and Macintosh, and a slash (/) on UNIX operating systems. This is most useful for parsing paths such as "c:\windows" or "MacVolume:System Folder".

So on Unix, the test would just be inserting a `/`, which wouldn't cause any exception to be raised.

`Drives` isn't specifically the reason why the test is Windows-only, but Unix doesn't have an analog to the `:` that Windows uses to separate drive letters from the rest of the path. Could add `WindowsSpecifics.VolumeSeparator`?

TestHelpers.Tests/MockFileCopyTests.cs
rkoeninger: Added a class called `Shared` to contain the building of the chars. There isn't a very good name for them though, as they're just chars that don't have `ArgumentException` to be raised. So it's called `SpecialInvalidPathChars`.

TestHelpers.Tests/MockFileCopyTests.cs
rkoeninger: Done for this and 3 other similar tests.

TestHelpers.Tests/MockFileCopyTests.cs
rkoeninger: Not my style, but sure, done.

TestHelpers.Tests/MockFileReadAllBytesTests.cs
rkoeninger: Used byte array instead and built paths using `fileSystem.Path.DirectorySeparatorChar` and `.AltDirectorySeparatorChar`.

rkoeninger:
@jpreese Made suggested changes, except for [the volume separator one](https://github.com/System-IO-Abstractions/System.IO.Abstractions/pull/327#discussion_r205952375), which I don't know what to do for.

TestHelpers.Tests/MockFileCopyTests.cs
fgreinacher: I think this approach is fine. @jpreese Any objections left?

fgreinacher:
Merging this one. If we want to do something for the volume separator disuccuisb we can do that in a separate PR.

PR #328, opened 2018-07-17
#245: Add comments via inheritdoc

robertlarkins:
This commit is part of [Issue #245](https://github.com/System-IO-Abstractions/System.IO.Abstractions/issues/245) and adds comments using the inheritdoc tag. At present visual studio does not recognise these tags, so intellisense won't display them, but other tools such as ReSharper will.

The next stage is to modify the appveyor.yml and use the InheritDoc nuget package to modify the xml documentation on deployment. This will be a separate pull request.


fgreinacher:
Thanks, merged!

PR #336, opened 2018-07-20
#245: Generate documentation file for all builds

robertlarkins:
Setting the XML documentation file through Project Properties/Build
only generated the XML documentation file for NET40 and not
netstandard1.4 nor netstandard2.0. Adding the property
GenerateDocumentationFile and setting it to true generates the
documentation for all build types.


fgreinacher:
Thanks!

Issue #339, opened 2018-08-06
Changes to MockFileSystem.RemoveFile causing issues with sub directories


mattdwen:
I've just upgraded from 2.0.0.144 to find a lot of my tests breaking.

Most relate to the changes to `MockFileSystem.RemoveFile` checking if the path has the ReadOnly attribute set and throwing a UnauthorizedAccessException if so.

`MockDirectory.Move` for example enumerates all sub directories and recursively calls `Move` again. If any child file or directory of this directory have ReadOnly set then the move fails.

I can think of four possible solutions:
- Modify `MockFileSystem.RemoveFile` to check if it's a directory and not throw the exception
- Add `MockFileSystem.RemoveDirectory` and refactor to use internally
- Add a bypass to these checks. If I know I'm accessing the `MockFileSystem` from my test code, I should be able to blow away a whole directory structure without any normal file system checks.
- Manually implement an iterator in my test code to pass the filesystem checks.

I'm not 100% sure what a real file system implementation is if a subdirectory, or even a file 5 levels below is ReadOnly, but I'm pretty sure I can move the very top folder with no issues.

A lot of my code is rebuilding directory structures, so the tests are to handle scenarios where a whole directory has been deleted. It's now getting very difficult to implement this as a test.

I'm happy to create a PR, but would good to get some input first.

Cheers!

mattdwen:
I've created a PR with a test that fails.

ericnewton76:
First need to know what the behavior is on real filesystems like NTFS.

mattdwen:
I've updated the PR thread with some test code to check this against a real file system, which passes on both NTFS and FAT32.

fgreinacher:
Merged with a072541 - thanks!

PR #340, opened 2018-08-06
Test that a directory with read only sub directories can be moved

mattdwen:
Test that should fail for issue #339

### DIFF
--- a/System.IO.Abstractions.TestingHelpers.Tests/MockDirectoryTests.cs
+++ b/System.IO.Abstractions.TestingHelpers.Tests/MockDirectoryTests.cs
@@ -1150,6 +1150,31 @@ namespace System.IO.Abstractions.TestingHelpers.Tests
         }
 
         [Test]
+        public void MockDirectory_Move_ShouldMoveDirectoryWithReadOnlySubDirectory()
+        {
+            // Arrange
+            var sourceDirName = XFS.Path(@"a:\folder1\");
+            var sourceSubDirName = XFS.Path(@"a:\folder1\sub\");
+
+            var destDirName = XFS.Path(@"a:\folder2\");
+            var destSubDirName = XFS.Path(@"a:\folder2\sub\");
+
+            var fileSystem = new MockFileSystem();
+            fileSystem.AddDirectory(sourceSubDirName);
+
+            var subDirectoryInfo = fileSystem.DirectoryInfo.FromDirectoryName(sourceSubDirName);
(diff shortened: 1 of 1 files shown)


ericnewton76:
Need to know what NTFS does in this case.  And by read-only are we talking about `attrib +r dir` or ACL readonly?  This is the problem with these esoteric behaviors that are bubbling up from the filesystem itself... FAT and NTFS might do differently, and ReFS might do something altogether different.

I suspect they dont issue recursive subdirectory checks like the code naively does.  Your implementation will probably need to iterate the subfolders to rewrite the parent paths probably (havent looked at the base implementation, but if I remember from a long time ago, the parent path is stored right as a string.  I would recommend an iterative approach to rewrite all the subfolder's base paths.

Assuming NTFS allows for this behavior as a baseline for behavior mimicry.

mattdwen:
I've written a test which:
 - creates a folder
 - creates a text file in a sub directory
 - sets the file to read only
 - moves the main folder

I've run this against both an NTFS and FAT32 volume, and both pass the test. So it appears it will be okay to just rewrite the paths.

```
[Fact]
    public void Test1()
    {
      // Arrange
      const string testPath = @"E:\TestBase";
      var sourceDirectoryPath = $@"{testPath}\Source";
      var destDirectoryPath = $@"{testPath}\Dest";

      const string fileName = @"Sub\Directory\File.txt";
      var filePath = $@"{sourceDirectoryPath}\{fileName}";

      Assert.True(Directory.Exists(testPath));
      Assert.False(Directory.Exists(sourceDirectoryPath));
      Assert.False(Directory.Exists(destDirectoryPath));

      Directory.CreateDirectory(sourceDirectoryPath);
      Directory.CreateDirectory($@"{sourceDirectoryPath}\Sub");
      Directory.CreateDirectory($@"{sourceDirectoryPath}\Sub\Directory");
      using (var writer = File.CreateText(filePath))
      {
        writer.Write("Hello, world!");
      }
      Assert.True(File.Exists(filePath));

      File.SetAttributes(filePath, FileAttributes.ReadOnly);

      // Act
      Directory.Move(sourceDirectoryPath, destDirectoryPath);

      // Assert
      Assert.False(Directory.Exists(sourceDirectoryPath));
      Assert.True(Directory.Exists(destDirectoryPath));
      Assert.True(File.Exists($@"{destDirectoryPath}\{fileName}"));
      Assert.Equal(FileAttributes.ReadOnly, File.GetAttributes($@"{destDirectoryPath}\{fileName}"));
    }
```

PR #344, opened 2018-08-07
Rewrite internal path keys when using MockDirectory.Move

mattdwen:
Issue #339, Replaces PR #340.

I had to add a new method to `MockFileSystem` which isn't ideal, but there was nothing which exposed the internal `files` dictionary.

I've also added some additional checks to the mock directory move which I've observed on a real NTFS system, specifically a `DirectoryNotFound` exception if the source directory doesn't exist, and an `IOException` if the destination directory does exist.


ericnewton76:
Looks like a good set of commits, if my opinion matters... lol

PR #348, opened 2018-08-13
Implicit conversion from FileInfo to FileInfoBase

### LINKED ISSUE #347: Implicit conversion from System.IO.FileInfo to FileInfoBase should not fail when the value is null.
When you pass a null FileInfo to a method expecting a FileInfoBase you will get an ArgumentNullException.

In FileInfoBase, instead of

```
        public static implicit operator FileInfoBase(FileInfo fileInfo)
        {
            return new FileInfoWrapper(fileInfo);
        }
```

it should be

```
        public static implicit operator FileInfoBase(FileInfo fileInfo)
        {
            if (fileInfo == null)
            {
                return null;
            }

            return new FileInfoWrapper(fileInfo);
        }
```

The other conversions (e. g. from DirectoryInfo to DirectoryInfoBase) seem to have this check.

m-ringler:
Fixes #347

### DIFF
--- a/System.IO.Abstractions.TestingHelpers.Tests/MockFileInfoTests.cs
+++ b/System.IO.Abstractions.TestingHelpers.Tests/MockFileInfoTests.cs
@@ -10,6 +10,17 @@ namespace System.IO.Abstractions.TestingHelpers.Tests
     public class MockFileInfoTests
     {
         [Test]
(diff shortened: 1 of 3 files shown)


fgreinacher:
Thanks!

m-ringler:
Here's the requested test...

fgreinacher:
Thanks again!

PR #349, opened 2018-08-22
Prevent duplicate dispose exception

gitting:
In situations where multiple `using`'s are stacked against a stream and a reader/writer a "Cannot access a closed Stream" exception is thrown which is not present in a default `FileSystem`. In particular it's
observed in netcore 2.0, so a new test run with `dotnet test` against original source will expose it.

The fix just keeps track of disposal attempts.

### DIFF
--- a/System.IO.Abstractions.TestingHelpers.Tests/MockFileStreamTests.cs
+++ b/System.IO.Abstractions.TestingHelpers.Tests/MockFileStreamTests.cs
@@ -89,5 +89,20 @@
             // Assert
             Assert.DoesNotThrow(() => stream.Close());
         }
+
+        [Test]
+        public void MockFileStream_Dispose_MultipleCallsDontThrow()
+        {
+            var fileSystem = new MockFileSystem();
+            var path = XFS.Path("C:\\test");
+            fileSystem.AddFile(path, new MockFileData("Bla"));
+            var stream = fileSystem.File.OpenRead(path);
+
+            // Act
+            stream.Dispose();
+
+            // Assert
+            Assert.DoesNotThrow(() => stream.Dispose());
+        }
     }
 }
--- a/System.IO.Abstractions.TestingHelpers.Tests/MockFileStreamTests.cs
+++ b/System.IO.Abstractions.TestingHelpers.Tests/MockFileStreamTests.cs
@@ -74,5 +74,20 @@
             Assert.IsFalse(stream.CanWrite);
             Assert.Throws<NotSupportedException>(() => stream.WriteByte(1));
         }
+
+        [Test]
+        public void MockFileStream_Close_MultipleCallsDontThrow()
+        {
+            var fileSystem = new MockFileSystem();
+            var path = XFS.Path("C:\\test");
+            fileSystem.AddFile(path, new MockFileData("Bla"));
+            var stream = fileSystem.File.OpenRead(path);
+
+            // Act
+            stream.Close();
+
+            // Assert
+            Assert.DoesNotThrow(() => stream.Close());
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


System.IO.Abstractions.TestingHelpers.Tests/MockFileStreamTests.cs
```diff
@@ -74,5 +74,20 @@ public void MockFileStream_Constructor_ReadTypeNotWritable()
...
+            var fileSystem = new MockFileSystem();
+            var path = XFS.Path("C:\\test");
+            fileSystem.AddFile(path, new MockFileData("Bla"));
+            var stream = fileSystem.File.OpenRead(path);
+
+            // Act
+            stream.Close();
+
+            // Assert
+            Assert.DoesNotThrow(() => stream.Close());
```
fgreinacher: Is this the same behavior that a real `FileStream` shows?

fgreinacher: Probably an additional test which calls `Dispose` two times would make sense.

System.IO.Abstractions.TestingHelpers/MockFileStream.cs
```diff
@@ -49,12 +50,14 @@ public MockFileStream(IMockFileDataAccessor mockFileDataAccessor, string path, S
         public override void Close()
         {
             InternalFlush();
+            disposed = true;
         }
 #else
         protected override void Dispose(bool disposing)
         {
             InternalFlush();
```
fgreinacher: Should we just call `Close`here?

System.IO.Abstractions.TestingHelpers/MockFileStream.cs
```diff
@@ -65,6 +68,11 @@ public override void Flush()
 
         private void InternalFlush()
         {
+            if (disposed)
```
fgreinacher: I wouldn't evaluate this here. Maybe better to do it within `Close`/`Dispose`. This code is also executed when people call `Flush` directly, where I'd rather expect an exception when the stream is already closed.

System.IO.Abstractions.TestingHelpers.Tests/MockFileStreamTests.cs
gitting: Yes. These would pass:

``` CS

        [Test]
        public void DefaultFileStream_Close_MultipleCallsDontThrow()
        {
            var tempPath = Path.GetTempPath();
            var fileStream = File.Create(Path.Combine(tempPath, "so_i_can_see_it_close"));

            // Act
            fileStream.Close();

            // Assert
            Assert.DoesNotThrow(() => fileStream.Close());
        }

        [Test]
        public void DefaultFileStream_Dispose_MultipleCallsDontThrow()
        {
            var tempPath = Path.GetTempPath();
            var fileStream = File.Create(Path.Combine(tempPath, "so_i_can_see_it_dispose"));

            // Act
            fileStream.Dispose();

            // Assert
            Assert.DoesNotThrow(() => fileStream.Dispose());
        }
```

System.IO.Abstractions.TestingHelpers/MockFileStream.cs
gitting: Moved to just the `Dispose` for the failing netcore branch anyway.

System.IO.Abstractions.TestingHelpers/MockFileStream.cs
gitting: Not sure what you're suggesting, calling `Close` from the `Dispose`? The branching seems to be here for some purpose as the double disposing does not crash in NET40, only in netcore.

System.IO.Abstractions.TestingHelpers.Tests/MockFileStreamTests.cs
gitting: Done.

Issue #351, opened 2018-08-24
FileBase, DirectoryBase, PathBase, etc. should expose an IFileSystem property


herebebeasties:
We have an internal extensions library for `IFileSystem` that adds various nice extension methods like `Unzip(...)` and `DeleteDirectoryContents(...)`.

Because these need to use operations on a mixture of `FileBase`, `DirectoryBase` and `PathBase`, these currently need to be extension methods on `IFileSystem` which isn't ideal:
 * It's much less discoverable, particularly for things that really ought to live alongside a closely-related API like `DirectoryBase.Delete`.
 * It causes API breaks if you add functionality to an extension method on `FileBase` and suddenly discover you need to move it to an `IFileSystem` one instead.

It would be extremely useful to expose the underlying `IFileSystem` on all these base classes as a property, which would fix this.

It would be a pretty trivial change, obviously, although as with any API addition would break anyone who has a custom implementations of FileBase, etc. as obviously they'd need to add it in (pretty trivial to fix, though).

Thoughts?

fgreinacher:
IMO that’s a good idea which is very similar to what’s proposed in #239.

herebebeasties:
OK, great - am happy to work up a PR. It looks like adding this to DirectoryInfo / FileInfo stuff is a bit tricky, because they have an implicit casting operator:
```c#
public static implicit operator DirectoryInfoBase(DirectoryInfo directoryInfo)
public static implicit operator FileInfoBase(FileInfo fileInfo)
```
On the basis that this implicitly casts things to the FileSystem implementation, which may well not be what you want anyway, I'm tempted to just new up a `new FileSystem()` for those. (I don't think having implicit casts like this is a good idea anyway, as it'll silently do the wrong thing if you actually meant for all your usages to use something like a `LongPathFileSystem`.)

Also - `DirectoryWrapper`, `FileWrapper`, etc. are marked as `[Serializable]`. If they suddenly have an instance field for the `IFileSystem` then that will be serialized as well. I can see this was added in #22 by @ericnewton76 but it doesn't have much context as to why. Can anyone shed any light on why you'd want to serialize a `DirectoryBase` instance? It doesn't make much sense to me - feels similar to trying to serialize a Socket or a FileStream or a DNS resolver, or something else ephemeral like that.

fgreinacher:
Implemented by #352

PR #352, opened 2018-08-27
#351: Add IFileSystem property to base classes

herebebeasties:
This allows people to get at the underlying IFileSystem implementation when developing extension methods.

For example, it may make sense to put your extension method on DirectoryBase, but you may require FileBase to actually implement it. This makes that possible.

In addition it fixes a bug in MockFileInfo.Replace where the returned FileInfo was for the concrete FileSystem not the mock one (bug caused by implicit casting).

I think this commit actually removes the need for supporting implicit casting at all.

### DIFF
--- a/System.IO.Abstractions.TestingHelpers/IMockFileDataAccessor.cs
+++ b/System.IO.Abstractions.TestingHelpers/IMockFileDataAccessor.cs
@@ -62,5 +62,7 @@ namespace System.IO.Abstractions.TestingHelpers
         IDriveInfoFactory DriveInfo { get; }
 
         PathVerifier PathVerifier { get; }
+
+        IFileSystem FileSystem { get; }
     }
 }
--- a/System.IO.Abstractions.TestingHelpers/MockDirectory.cs
+++ b/System.IO.Abstractions.TestingHelpers/MockDirectory.cs
@@ -18,7 +18,7 @@ namespace System.IO.Abstractions.TestingHelpers
 
         private string currentDirectory;
 
-        public MockDirectory(IMockFileDataAccessor mockFileDataAccessor, FileBase fileBase, string currentDirectory)
+        public MockDirectory(IMockFileDataAccessor mockFileDataAccessor, FileBase fileBase, string currentDirectory) : base(mockFileDataAccessor?.FileSystem)
         {
             this.currentDirectory = currentDirectory;
             this.mockFileDataAccessor = mockFileDataAccessor ?? throw new ArgumentNullException(nameof(mockFileDataAccessor));
--- a/System.IO.Abstractions/DirectoryBase.cs
+++ b/System.IO.Abstractions/DirectoryBase.cs
@@ -9,7 +9,7 @@ namespace System.IO.Abstractions
     {
         protected DirectoryBase(IFileSystem fileSystem)
         {
-            this.FileSystem = fileSystem;
+            FileSystem = fileSystem;
         }
 
         /// <summary>
--- a/System.IO.Abstractions/DirectoryBase.cs
+++ b/System.IO.Abstractions/DirectoryBase.cs
@@ -7,6 +7,16 @@ namespace System.IO.Abstractions
     [Serializable]
     public abstract class DirectoryBase
     {
+        protected DirectoryBase(IFileSystem fileSystem)
+        {
+            this.FileSystem = fileSystem;
+        }
+
+        /// <summary>
+        /// Exposes the underlying filesystem implementation. This is useful for implementing extension methods.
+        /// </summary>
+        public IFileSystem FileSystem { get; }
+
         /// <inheritdoc cref="Directory.CreateDirectory(string)"/>
         public abstract DirectoryInfoBase CreateDirectory(string path);
 
--- a/System.IO.Abstractions/DirectoryInfoBase.cs
+++ b/System.IO.Abstractions/DirectoryInfoBase.cs
@@ -7,6 +7,10 @@ namespace System.IO.Abstractions
     [Serializable]
     public abstract class DirectoryInfoBase : FileSystemInfoBase
     {
+        protected DirectoryInfoBase(IFileSystem fileSystem) : base(fileSystem)
+        {
+        }
+
         /// <inheritdoc cref="DirectoryInfo.Create()"/>
         public abstract void Create();
 #if NET40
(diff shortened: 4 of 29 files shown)


herebebeasties:
This implements #351.

System.IO.Abstractions.TestingHelpers/IMockFileDataAccessor.cs
```diff
@@ -62,5 +62,7 @@ public interface IMockFileDataAccessor
         IDriveInfoFactory DriveInfo { get; }
 
         PathVerifier PathVerifier { get; }
+
+        IFileSystem FileSystem { get; }
```
herebebeasties: This feels slightly icky to me, but it was the least invasive change to make, as the various constructors for all the mock classes can then remain the same.

System.IO.Abstractions.TestingHelpers/MockDirectory.cs
```diff
@@ -18,7 +18,7 @@ public class MockDirectory : DirectoryBase
 
         private string currentDirectory;
 
-        public MockDirectory(IMockFileDataAccessor mockFileDataAccessor, FileBase fileBase, string currentDirectory)
+        public MockDirectory(IMockFileDataAccessor mockFileDataAccessor, FileBase fileBase, string currentDirectory) : base(mockFileDataAccessor?.FileSystem)
```
herebebeasties: The reason the above feels a bit icky is that you end up with these `null` checks which don't make all that much sense. The alternative here is obviously to pass in both the `IMockFileDataAccessor` and the `IFileSystem` to all these. Would people prefer we do that instead?

System.IO.Abstractions/DirectoryInfoBase.cs
```diff
@@ -102,7 +106,7 @@ public abstract class DirectoryInfoBase : FileSystemInfoBase
         {
             if (directoryInfo == null)
                 return null;
-            return new DirectoryInfoWrapper(directoryInfo);
+            return new DirectoryInfoWrapper(new FileSystem(), directoryInfo);
```
herebebeasties: I can't see much alternative to this given how the implicit cast works.
We could have a readonly static `FileSystem.Instance` field which we encourage people to use, I guess.

(To be honest I'm not a huge fan of implicit casts like this - they make dependency injection harder and tend to hide bugs, as witnessed by the one I fixed as part of this PR in `MockFileInfo`. Part of the public API, though. :-/)

System.IO.Abstractions/FileSystem.cs
```diff
@@ -30,26 +36,25 @@ public IFileStreamFactory FileStream
...
         }
 
         DirectoryInfoFactory directoryInfoFactory;
         public IDirectoryInfoFactory DirectoryInfo
         {
-            get { return directoryInfoFactory ?? (directoryInfoFactory = new DirectoryInfoFactory()); }
+            get { return directoryInfoFactory ?? (directoryInfoFactory = new DirectoryInfoFactory(this)); }
         }
 
-        private readonly Lazy<DriveInfoFactory> driveInfoFactory = new Lazy<DriveInfoFactory>(() => new DriveInfoFactory());
```
herebebeasties: I'm not sure why this is the only one done as a `Lazy<T>`, but I've left it as-is.
I can fix this up to make is all consistent in another PR if you like?
(Or do it as part of this one if you don't mind the slight diff noise.)

System.IO.Abstractions.TestingHelpers/IMockFileDataAccessor.cs
fgreinacher: I like it this way!

System.IO.Abstractions/FileSystem.cs
fgreinacher: Would be great to make this consistent (I slightly prefer `Lazy<T>`)

System.IO.Abstractions/FileBase.cs
```diff
@@ -8,6 +8,16 @@ namespace System.IO.Abstractions
     [Serializable]
     public abstract class FileBase
     {
+        protected FileBase(IFileSystem fileSystem)
+        {
+            this.FileSystem = fileSystem;
```
fgreinacher: Nit: Redundant `this.`

System.IO.Abstractions/DirectoryBase.cs
```diff
@@ -7,6 +7,16 @@ namespace System.IO.Abstractions
     [Serializable]
     public abstract class DirectoryBase
     {
+        protected DirectoryBase(IFileSystem fileSystem)
+        {
+            this.FileSystem = fileSystem;
```
fgreinacher: Nit: Redundant `this.`

System.IO.Abstractions/DriveInfoBase.cs
```diff
@@ -3,6 +3,13 @@
     [Serializable]
     public abstract class DriveInfoBase
     {
+        protected DriveInfoBase(IFileSystem fileSystem)
+        {
+            FileSystem = fileSystem ?? throw new ArgumentNullException(nameof(fileSystem));
+        }
+
+        public IFileSystem FileSystem { get; }
```
fgreinacher: Comment is missing

System.IO.Abstractions/FileSystem.cs
```diff
@@ -3,22 +3,28 @@
     [Serializable]
     public class FileSystem : IFileSystem
     {
-        DirectoryBase directory;
+        public FileSystem()
+        {
+            driveInfoFactory = new Lazy<DriveInfoFactory>(() => new DriveInfoFactory(this));;
```
fgreinacher: Nit: Double `;` at the end

System.IO.Abstractions/FileSystemInfoBase.cs
```diff
@@ -4,6 +4,13 @@
     [Serializable]
     public abstract class FileSystemInfoBase
     {
+        protected FileSystemInfoBase(IFileSystem fileSystem)
+        {
+            FileSystem = fileSystem ?? throw new ArgumentNullException(nameof(fileSystem));
+        }
+
+        public IFileSystem FileSystem { get; }
```
fgreinacher: Missing comment

System.IO.Abstractions/FileSystem.cs
herebebeasties: I've binned all the lazy entirely and just construct these instead - given all the constructors do basic field init (added in this PR) and nothing else, I really can't understand why these were made lazy in the first place!

herebebeasties:
I've addressed all your comments, Florian. Thanks. Merged up to master, tests are all green.

fgreinacher:
Great, thanks again!

PR #354, opened 2018-09-21
Ensure MockFile.Copy clones contents

### LINKED ISSUE #353: TestingHelper.File.Copy initialises contents via object reference
When you copy a file the file contents gets passed as a reference instead of being cloned. If I copy FileA.txt to FileB.txt, they will both point to the same contents object. When I then change the contents of FileA.txt, FileB.txt will also update.

          
            var mockFileSystem = new MockFileSystem();
            mockFileSystem.AddFile("C:\\Test1.txt", "");
            mockFileSystem.File.Copy("C:\\Test1.txt", "C:\\Test2.txt");

            var before1 = mockFileSystem.FileInfo.FromFileName("C:\\Test1.txt").Length;
            var before2 = mockFileSystem.FileInfo.FromFileName("C:\\Test2.txt").Length;

            using (var stream = mockFileSystem.File.Open("C:\\Test1.txt", FileMode.Open, FileAccess.ReadWrite))
            {
                var binaryWriter = new System.IO.BinaryWriter(stream);

                binaryWriter.Seek(0, SeekOrigin.Begin);
                binaryWriter.Write("Some text");
            }

            var after1 = mockFileSystem.FileInfo.FromFileName("C:\\Test1.txt").Length;
            var after2 = mockFileSystem.FileInfo.FromFileName("C:\\Test2.txt").Length;

fgreinacher:
Fixes #353

### DIFF
--- a/System.IO.Abstractions.TestingHelpers.Tests/MockFileCopyTests.cs
+++ b/System.IO.Abstractions.TestingHelpers.Tests/MockFileCopyTests.cs
@@ -28,6 +28,27 @@ namespace System.IO.Abstractions.TestingHelpers.Tests
         }
 
         [Test]
+        public void MockFile_Copy_ShouldCloneContents()
+        {
+            var sourceFileName = XFS.Path(@"c:\source\demo.txt");
+            var destFileName = XFS.Path(@"c:\source\demo_copy.txt");
+
+            var mockFileSystem = new MockFileSystem();
+            mockFileSystem.AddFile(sourceFileName, "Original");
+            mockFileSystem.File.Copy(sourceFileName, destFileName);
+            
+            using (var stream = mockFileSystem.File.Open(sourceFileName, FileMode.Open, FileAccess.ReadWrite))
+            {
+                var binaryWriter = new System.IO.BinaryWriter(stream);
+
+                binaryWriter.Seek(0, SeekOrigin.Begin);
+                binaryWriter.Write("Modified");
+            }
+
+            Assert.AreEqual("Original", mockFileSystem.File.ReadAllText(destFileName));
+        }
+
+        [Test]
         public void MockFile_Copy_ShouldCreateFileAtNewDestination()
         {
             string sourceFileName = XFS.Path(@"c:\source\demo.txt");
--- a/System.IO.Abstractions.TestingHelpers/MockFileData.cs
+++ b/System.IO.Abstractions.TestingHelpers/MockFileData.cs
@@ -81,6 +81,24 @@ namespace System.IO.Abstractions.TestingHelpers
             Contents = contents ?? throw new ArgumentNullException(nameof(contents));
         }
 
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="MockFileData"/> class by copying the given <see cref="MockFileData"/>.
+        /// </summary>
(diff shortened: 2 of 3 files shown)


System.IO.Abstractions.TestingHelpers/MockFileData.cs
```diff
@@ -81,6 +81,24 @@ public MockFileData(byte[] contents)
...
 
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="MockFileData"/> class by copying the given <see cref="MockFileData"/>.
+        /// </summary>
+        /// <param name="template">The template instance.</param>
+        /// <exception cref="ArgumentNullException">Thrown if <paramref name="template"/> is <see langword="null" />.</exception>
+        public MockFileData(MockFileData template)
+        {
+            if (template == null) throw new ArgumentNullException(nameof(template));
```
robkeim: Minor: for consistency for one line conditional blocks do we prefer this syntax or the following:
```
if (condition)
{
    // Statement to execute
}
```

I see the later being used in MockFileData.cs:
https://github.com/System-IO-Abstractions/System.IO.Abstractions/blob/55a82b00781139b041d4a2b9a59a4ad9956d28a7/System.IO.Abstractions.TestingHelpers/MockFile.cs#L31

System.IO.Abstractions.TestingHelpers.Tests/MockFileCopyTests.cs
```diff
@@ -27,6 +27,27 @@ public void MockFile_Copy_ShouldOverwriteFileWhenOverwriteFlagIsTrue()
...
+        public void MockFile_Copy_ShouldCloneContents()
+        {
+            var sourceFileName = XFS.Path(@"c:\source\demo.txt");
+            var destFileName = XFS.Path(@"c:\source\demo_copy.txt");
+
+            var mockFileSystem = new MockFileSystem();
+            mockFileSystem.AddFile(sourceFileName, "Original");
+            mockFileSystem.File.Copy(sourceFileName, destFileName);
+            
+            using (var stream = mockFileSystem.File.Open(sourceFileName, FileMode.Open, FileAccess.ReadWrite))
```
jpreese: I'd probably use a more direct approach to modifying the file. Straight up deleting it, clearing its contents, etc. Seems a little overkill to leverage a binarywriter, and set the contents to modified.

fgreinacher: Hm, for a repro the byte array holding the contents needs to be directly modified. I think most other APIs will go through the string path. Will check later if there's something simpler that'l repro.

jpreese: I didn't actually mess around with it, but looking at the issue, it sounds like the copied file is actually just a reference to the original. To me that means if we were to say.. delete the original, the copied file would be deleted as well (because of the found bug).

fgreinacher: No, deleting the old file has no effect, because this just removes the old entry from the dictionary (without touching the data). This behavior is only observable when using `MockFileStream` to modify the contents of the file directly.

PR #359, opened 2018-09-28
Add internal, parameterless ctors for generated proxies

### LINKED ISSUE #355: Breaking change in version 2.1.0.230
Before version 2.1.0.230 you could do this:
`_fileSystem = new Mock<IFileSystem>();`
`_fileSystem.Setup(x => x.Directory.Exists(It.IsAny<string>())).Returns(true);`

But that doesn't work any longer, you get this error:
> Castle.DynamicProxy.InvalidProxyConstructorArgumentsException: Can not instantiate proxy of class: System.IO.Abstractions.DirectoryBase.
Could not find a parameterless constructor..

How should you write this code to get this to work now?

fgreinacher:
Mocking libraries, like Moq, are based on
Castle.Core and rely on its functionality for
generating dynamic proxies. This is only possible
when there is an accessible parameterless constructor.
This changes re-adds such constructors to several
base types. They are marked internal and obsolete
so that nobody accidentially uses them in "real" code.
To allow Castle.Core to access them altough they are
internal we expose our internals to the dynamic proxy
assembly.

Long-term we should expose interfaces instead of
abstract classes at `IFileSystem`.

Fixes #355


fgreinacher:
@herebebeasties This is my take on #355. It's ugly but it works :/

fgreinacher:
I added some tests to ensure that all properties exposed at `IFileSystem` can be mocked to prevent running into this problem again. Doing this I found that one of them was missing, but now it seems to be complete.

PR #360, opened 2018-09-30
Let null MockFileData create empty file fixes #356

### LINKED ISSUE #356: Null MockFileData makes EnumerateDirectories return false
Drop the following test in `MockDirectoryTests.cs` and see it fail.

        [Test]
        public void MockDirectory_EnumerateDirectories_WithNoData_ShouldReturnDirectory()
        {
            // Arrange

            // Normally one would use `new MockFileData(string.Empty));` but I use `null`.
            fileSystem.AddFile(XFS.Path(@"C:\Folder\bar"), null);  // <----- there is the culprit

            // Act
            var actualResult = fileSystem.Directory.EnumerateDirectories(XFS.Path(@"c:\"), "Folder");

            // Assert
            // This assertion should succeed but fails.
            Assert.That(actualResult.Count(), Is.EqualTo(1));
        }

I tracked it to class `MockFileSystem`, method `AllDirectories`.

        public IEnumerable<string> AllDirectories
        {
            get
            {
                lock (files)
                {
                    return files.Where(f => f.Value.IsDirectory).Select(f => f.Key).ToArray();
                }
            }
        }
where `f.Value.IsDirectory` fails due to `Value` being null.
Method `Exists` then handles any exceptions and returns false.

I have verified that if one has a real folder with a file of length 0, the folder will be returned when `System.IO.Directory.EnumerateDirectories` is called. At least on Windows10.

An easy solution would be to throw a `NullArgumentException` when `AddFile` is called without `MockFileData` but I am not in the position to deem it the correct one. This might be a breaking change.
Another solution could be to `mockFile = mockFile ?? new MockFileData(string.Empty);` in method `AddFile`.
A third solution could be to update the failing code with `f.Value?.IsDirectory??true` but that might have side effects.

LosManos:
The, technically, obvious place to put a test for this issue was for `MockFileSystem.AddFile` method.
But user wise the test belonged to the constructore.
So I added both.

I have *not* tested in a *nix environment.


System.IO.Abstractions.TestingHelpers.Tests/MockFileSystemTests.cs
```diff
@@ -10,6 +10,22 @@ namespace System.IO.Abstractions.TestingHelpers.Tests
     [TestFixture]
     public class MockFileSystemTests
     {
+        [Test]
+        public void MockFileSystem_Constructor_ShouldAllowNullForEmptyFile()
```
fgreinacher: I don’t get what’s the difference between this test and the one below. Seems to be enough to keep the latter.

System.IO.Abstractions.TestingHelpers.Tests/MockFileSystemTests.cs
```diff
@@ -54,6 +70,19 @@ public void MockFileSystem_GetFile_ShouldReturnFileRegisteredInConstructorWhenPa
...
 
+        [Test]
+        public void MockFileSystem_AddFile_ShouldHandleNullFileDataAsEmpty()
+        {
+            var fileSystem = new MockFileSystem(new Dictionary<string, MockFileData>
+            {
+                { @"c:\something\nullish.txt", null }
+            });
+
+            var result = fileSystem.GetFile(@"c:\SomeThing\nullish.txt").TextContents;
```
fgreinacher: I’d suggest using `IFileSystem` APIs for the assert parts.

System.IO.Abstractions.TestingHelpers.Tests/MockFileSystemTests.cs
LosManos: I remove the constructor. After resolving conversation I'll try to find out how to update the PR.

System.IO.Abstractions.TestingHelpers.Tests/MockFileSystemTests.cs
LosManos: What do you mean?

fgreinacher: I mean using some method exposed via `IFileSystem` to get the file contents, in this case probably `fileSystem.File.ReadAllText`.

fgreinacher:
Thanks again!

PR #362, opened 2018-10-02
Bug356

bkrumme:
Fix for bug 356 - instantiated mockFile with an empty string if it is null.


fgreinacher:
Thanks for your contribution, but the same thing is already being worked on in #360, so I’m closing this one. Feel free to add your comments to that PR!

PR #369, opened 2018-10-14
WIP: Improve versioning

fgreinacher:
The goal of this PR us to decouple versioning from the CI system. 

It uses `Nerdbank.GitVersioning` to automatically generate a version number  by combining a base version set in a config file with the number if commits since the last change to the base version file.

That makes version numbers predictable and reproducible. 

I’ll also provide a document that describes how a release would work if we decide to follow this path.

Happy to get feedback on this!


fgreinacher:
Microsoft published some new guidance on library versioning at https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/versioning. This confirms what @ericnewton76 proposed in #343:

> ✔️ CONSIDER only including a major version in the AssemblyVersion.

So I'll gladly change my opinion here and adapt the config so that AssemblyVersion is MAJOR.0.0.0.

fgreinacher:
@jpreese @rkoeninger @ericnewton76 I'd consider this ready for review now.

ericnewton76:
Looks great. This will go a long ways towards better package handling and less issues with inclusion of additional libraries that might base themselves on System.IO.Abstractions and inclusion into your application product that uses System.IO.Abstractions.

fgreinacher:
OK, here we go - let's see how this goes :)

Issue #350, opened 2018-08-24
MockFileSystem sets current directory to actual FS


dzarda:
It appears that the `MockFileSystem` constructor calls real filesystem API when the `currentDirectory` parameter is left empty.

The current directory is then set as `C:\Users\...\AppData\Local\Temp`

I suspect the following code
```csharp
    public MockFileSystem(IDictionary<string, MockFileData> files, string currentDirectory = "")
    {
        if (string.IsNullOrEmpty(currentDirectory))
        {
            currentDirectory = IO.Path.GetTempPath();
        }
        ...
```
in https://github.com/System-IO-Abstractions/System.IO.Abstractions/blob/4b8ff76fb410e15d26ab2cab807117323de08299/System.IO.Abstractions.TestingHelpers/MockFileSystem.cs#L22

fgreinacher:
Thanks for reporting this! I see a few options here:

- Not setting the current directory unless explicitly specified. Unsure what would be the effect of this though.
- Setting it to some hardcoded path, e.g. `c:\temp`. Should be safe and is predictable.
- Leaving it as is. I don't see too big a problem with the current approach as getting the temp path is non-invasive.

Thoughts?

josteink:
> Setting it to some hardcoded path, e.g. `c:\temp`. Should be safe and is predictable.

That's a very Windows-centric point of view :)

ewlloyd:
After a quick glance, I think that not setting it at all would cause issues in the `MockDirectory.EnsureAbsolutePath()` method (used by `MockDirectory.GetFiles()` and `MockDirectory.EnumerateDirectories()`).

The only question I'd have with leaving it as is, is whether there'd ever be a case where you wouldn't have the real `IO.Path` available.

dzarda:
I feel having a safe default is desired. Why can't we use the "root" concept for this; "C:\" and "/" respectively?

I'd like the current behavior be changed, since. Reason why this issue exists is because I spent an hour figuring out the unexpected behavior.

ewlloyd:
Fair enough. What was expected behavior, in your case?

dzarda:
The struggle boiled down to something along the lines...
```cs
var fs = new MockFileSystem();
var dirInfo = fs.DirectoryInfo.FromDirectoryName("."); // System.IO.DirectoryNotFoundException : Could not find a part of the path 'C:\Users\....\AppData\Local\Temp
```

[...]

PR #372, opened 2018-10-16
Stop calling real IO.Path.GetTempPath() when creating a MockFileSystem

tehmantra:
Fixes: #350 
Tried to do this with the smallest change possible. 
Hard-coding to `"C:\"` (which will get translated to `"/"` on unix) as per discussion on the issue.

Let me know if this is okay.

### DIFF
--- a/System.IO.Abstractions.TestingHelpers.Tests/MockDirectoryTests.cs
+++ b/System.IO.Abstractions.TestingHelpers.Tests/MockDirectoryTests.cs
@@ -1187,8 +1187,10 @@ namespace System.IO.Abstractions.TestingHelpers.Tests
 
 
         [Test]
-        public void MockDirectory_GetCurrentDirectory_ShouldReturnDefaultPathWhenNotSet() {
-            string directory = Path.GetTempPath();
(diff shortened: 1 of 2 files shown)


fgreinacher:
Thanks for your contribution, generally this seems like a safe and valuable change to make.

Could you please verify that a scenario like the one mentioned in https://github.com/System-IO-Abstractions/System.IO.Abstractions/issues/350#issuecomment-426609905 will work with your change?

fgreinacher:
I'd expect the linked snippet to blow up with our current release because the temp directory does not exist in the `MockFileSystem`. No need to automate the test though, I was just looking for a quick manual validation.

fgreinacher:
@tehmantra What about adding a test like this:

```csharp
public void MockFileSystem_DefaultState_CurrentDirectoryExists()
{

   var currentDirectory = fs.DirectoryInfo.FromDirectoryName(".");

   Assert.IsTrue(currentDirectory.Exists);
}
```

tehmantra:
@fgreinacher `MockFileSystem` will now create the current directory if it doesn't exist.
Added a couple of test cases around this too.

Cheers.

fgreinacher:
Thanks for your work @tehmantra!

PR #377, opened 2018-10-22
Correcting build status indentation

muthumalla:


fgreinacher:
It seems like these are your first steps on GitHub, so I‘m gonna merge this although it’s kinda tiny.

Thanks for the contribution and welcome on board!

Issue #284, opened 2018-04-09
MockFile.Delete does not throw exception when file does not exist


borismod:
Unit test to reproduce:

```C#
var fs = new MockFileSystem();
fs.File.Delete(@"c:\someFilePath.txt");
```

The above code does not throw any exception, while the real File.Delete from System IO will throw

jpreese:
Additional clarification.. it looks like this will break some old tests as well. We'll need to add a directory in the failing tests.

peter-sereda:
Hello @jpreese, @fgreinacher . I have tried to fix the issue in my repository fork. Could you please check the result and tell me if something needs to be corrected?
https://github.com/peter-sereda/System.IO.Abstractions/pull/1

fgreinacher:
Hey @peter-sereda - thanks for taking care of this. Please open a PR against this repository so that we get test results.

peter-sereda:
@fgreinacher , done.
https://github.com/System-IO-Abstractions/System.IO.Abstractions/pull/379

borismod:
@peter-sereda  @fgreinacher @jpreese  thanks a lot guys for your contribution. Really appreciated!

PR #379, opened 2018-10-25
#284: MockFile.Delete does not throw exception when file does not exist

peter-sereda:


peter-sereda:
Closed as out-dated. Will create a new one.

PR #383, opened 2018-10-28
Added System.IO.StreamWriter

sharpSteff:
Added abstraction layer of the streamwriter class including StreamwriterFactory, Mocks and Unittests


fgreinacher:
Any reason you didn’t reuse https://github.com/System-IO-Abstractions/System.IO.Abstractions/blob/master/System.IO.Abstractions.TestingHelpers/MockFileStreamFactory.cs here? Adding the factory at the public API surface sounds sensible to me, but I don’t see why we need to introduce that much new code.

sharpSteff:
> Any reason you didn’t reuse https://github.com/System-IO->Abstractions/System.IO.Abstractions/blob/master/System.IO.Abstractions.TestingHelpers/MockFileStre>amFactory.cs here? 

Hi @fgreinacher, 
Can you be more precise? 
In my understanding, you want to include the StreamWriterFactory methods in the FileStreamFactory.
Is this correct?


I really like your project and introduced it into our Application. However we are using a lot of StreamWriters and StreamReaders(which i want to add after this pullrequest ;) , so I decided to include these System.IO-Classes, since writing my one Textwriter with the System.IO.Abstraction.FileStream would have result in a lot of more redundant code.

fgreinacher:
Ah sorry, ignore my comment - I looked at this on my mobile and didn't see that you actually used `MockFileStream` :/

Could you please create a separate issue that describes your overall goal and link it to this PR so that the goal is clear to everyone?

I'll be on vacation for the next couple of days, will do a deeper review after that.

herebebeasties:
Am not a project maintainer or anything, but to me this seems a bit mad.

IFileSystem already provides you with Streams. I can't see why you'd want to mock out StreamWriter, and if you do, why you'd need to push that into System.IO.Abstractions as a thing that IFileSystem implementations themselves need to implement.

Surely the implementations aren't filesystem-specific? If you want something like this you should just write some extension methods on IFileSystem that do the obvious thing:

```c#
public static StreamWriter OpenWriter(this IFileSystem fs, string path)
{
    return new StreamWriter(fs.File.Open(path));
} 
```
...kinda thing.

Doesn't strike me this need to be in the base library at all.

fgreinacher:
I agree with what @herebebeasties is saying. The existing APIs provide everything that's necessary for most use cases. So I'm gonna close this.

Nevertheless thanks a lot for your work on this!

PR #387, opened 2018-10-30
Fix for #284: MockFile.Delete does not throw exception when file does not exist.

peter-sereda:
Attempt to fix the subject bug.

### DIFF
--- a/System.IO.Abstractions.TestingHelpers.Tests/MockFileDeleteTests.cs
+++ b/System.IO.Abstractions.TestingHelpers.Tests/MockFileDeleteTests.cs
@@ -57,7 +57,7 @@
             string filePath = XFS.Path("C:\\temp\\somefile.txt");
 
             // Delete() returns void, so there is nothing to check here beside absense of an exception
-            fileSystem.File.Delete(filePath);
+            Assert.DoesNotThrow(() => fileSystem.File.Delete(filePath));
         }
     }
 }
--- a/System.IO.Abstractions.TestingHelpers.Tests/MockFileDeleteTests.cs
+++ b/System.IO.Abstractions.TestingHelpers.Tests/MockFileDeleteTests.cs
@@ -1,6 +1,7 @@
 ﻿namespace System.IO.Abstractions.TestingHelpers.Tests
 {
-    using NUnit.Framework;
+  using System.Collections.Generic;
+  using NUnit.Framework;
 
     using XFS = MockUnixSupport;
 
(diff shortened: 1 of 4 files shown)


System.IO.Abstractions.TestingHelpers.Tests/MockFileDeleteTests.cs
```diff
@@ -35,5 +36,28 @@ public void MockFile_Delete_ShouldThrowArgumentExceptionIfPathContainsOnlyWhites
...
+        public void MockFile_Delete_ShouldSilentlyReturnIfNonExistingFileInExistingFolder()
+        {
+            var fileSystem = new MockFileSystem(new Dictionary<string, MockFileData>()
+            {
+                { XFS.Path("C:\\temp\\exist.txt"), new MockFileData("foobar") },
+            });
+
+            string filePath = XFS.Path("C:\\temp\\somefile.txt");
+
+            // Delete() returns void, so there is nothing to check here beside absense of an exception
```
fgreinacher: Consider using `Assert.DoesNotThrow` to support your comment by an actual assertion :)

System.IO.Abstractions.TestingHelpers.Tests/MockFileStreamTests.cs
```diff
@@ -29,7 +29,7 @@ public void MockFileStream_Flush_WritesByteToFile()
         public void MockFileStream_Dispose_ShouldNotResurrectFile()
         {
             var fileSystem = new MockFileSystem();
-            var path = XFS.Path("C:\\test");
+            var path = XFS.Path("C:\\some_folder\\test");
```
fgreinacher: Why is this change necessary?

peter-sereda: Because of potential bug. Consider this code:
```
            var path = XFS.Path("C:\\test.txt");
            DirectoryInfoBase dir = fileSystem.Directory.GetParent(path);
```
It works ok on Windows but throws NullReferenceException on Linux. And after my changes GetParent() method is called inside File.Delete(), so this test started to fail. Adding a new intermediate folder "some_folder" is the fastest way to make the test work again.

I can create a new issue with this if you think it is a real defect.

System.IO.Abstractions.TestingHelpers.Tests/MockFileDeleteTests.cs
peter-sereda: Fixed.

System.IO.Abstractions.TestingHelpers.Tests/MockFileStreamTests.cs
fgreinacher: Ah OK, thanks. Could you please create an issue for this problem and add a comment to the changed test case that refers to the issue?

fgreinacher: Thanks!

fgreinacher:
Thanks again!

peter-sereda:
Welcome!

PR #389, opened 2018-10-30
Adding MockFileData overload to serialize objects

padillah:
Added a simple overload to include object serialization to the MockFileData class. I moved the common code to a private method and called it from both the old constructor and the new overload.

### DIFF
--- a/System.IO.Abstractions.TestingHelpers/MockFileData.cs
+++ b/System.IO.Abstractions.TestingHelpers/MockFileData.cs
@@ -1,6 +1,10 @@
 ﻿using System.Linq;
 using System.Security.AccessControl;
 using System.Text;
+#if NET40 || NETSTANDARD_20
+using System.Runtime.Serialization;
+using System.Runtime.Serialization.Formatters.Binary;
+#endif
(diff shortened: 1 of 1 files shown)


fgreinacher:
Thanks for your work, but I don't think this is something that needs to be in `System.IO.Abstractions` for the following reasons:
- the discoverability of the new overload is low, I wouldn't intuitively know what it does 
- this is a low-level library for IO operations that should stay focused
- it is easy (and IMO more flexible) to do the serialization on top of `S.IO.A`

So, I'm sorry, but I'll reject this PR. 

For the future I'd suggest you to create an issue that allows discussing the idea before starting a PR.

padillah:
Quick question: If you wanted discussion is there any reason it couldn't be done... right here? In the PR. Where GitHub has provided a space for it?

fgreinacher:
Sure, PR is also fine, but needs more effort on your end (writing the code) and also for the maintainers (reviewing the changes). For new things it’s IMO worthwhile to discuss the idea first (only providing pseudo code if necessary).

Issue #401, opened 2018-11-14
File.Create throws exception for relative file paths


Gav-Brown:
#### Package Version

2.1.0.256

#### Repro steps

1. Create `MockFileSystem` 
2. Create working directory
3. Set working directory
4. Call
 
`using (var file = fileSystem.File.Create("File.txt"))
{
}`

#### Expected behavior

File is created in current working directory (which does exist).

#### Actual behavior

Exception is thrown.
Could not find a part of the path 'File.txt'.

This worked in version 2.0.0.144. I suspect it is due to change for #278 an #286
which assumes an absolute file path rather than a relative one.

.NET Framework version handles both cases, as does 

`fileSystem.File.WriteAllText("File.txt", null);`

fgreinacher:
Thanks for reporting this! Could you test if the problem also occurs in [the latest prerelease version](https://www.nuget.org/packages/System.IO.Abstractions/2.2.8-beta)? #372 might already have fixed it.

Gav-Brown:
Thanks. Alas it still throws in latest beta.

```
            string workingDirectory = @"C:\folder";
            string relativeFile = "file.txt";
            var fileSystem = new MockFileSystem();

            fileSystem.Directory.CreateDirectory(workingDirectory);
            fileSystem.Directory.SetCurrentDirectory(workingDirectory);

            // Throws.
            using(fileSystem.File.Create(relativeFile))
            {
            }

            // Works.
            fileSystem.File.WriteAllText(relativeFile, null);
```

ThomasBarnekow:
This seems to be related to my issue #404. 

The funny thing is that issue 404 is related to something that is not found ;-)

fgreinacher:
@Gav-Brown The problem is fixed with ec1b673. Could you verify your use case with the prerelease version https://www.nuget.org/packages/System.IO.Abstractions/2.2.10-beta?

Gav-Brown:
@fgreinacher Thanks. I've tested and can confirm this is now resolved.

PR #407, opened 2018-11-27
Fix file operations that fail when used with relative paths 

### LINKED ISSUE #404: Pull Request #352 introduces breaking change while NuGet package only increases patch version
Today, I tried to update to the latest NuGet package, which is v2.1.0.256 at the time of writing. I was using v2.1.0.178 and did not expect any breaking changes, since only the patch number was updated. However, the projects using the previously parameterless `PathWrapper` constructor no longer compiled.

Further, even after providing a `FileSystem` object to the constructor, some of my unit tests then failed. They have been passing with all the previous versions, including v2.1.0.178. This might or might not be related to the breaking API change.

Are you using semantic versioning for your NuGet packages?

fgreinacher:
Root cause for the problems is https://github.com/wleader/System.IO.Abstractions/commit/95ead3a2a3fbe848e34642931f2e5fafc2368c5a#diff-e76db4b13ed41e41943b80f8f1075287 where we started using `Path.GetDirectoryName` to get the parent directory, but unfortunately this method returns an empty string when the path does not contain directory information (e.g. a plain file name).

The changes here make the paths absolute before passing them on to `Path.GetDirectoryName`.

Fixes #404 and #401

### DIFF
--- a/System.IO.Abstractions.TestingHelpers.Tests/MockFileCreateTests.cs
+++ b/System.IO.Abstractions.TestingHelpers.Tests/MockFileCreateTests.cs
@@ -279,5 +279,16 @@ namespace System.IO.Abstractions.TestingHelpers.Tests
             Assert.IsTrue(fileInfo.Attributes.HasFlag(FileAttributes.Encrypted));
         }
 #endif
+
+        [Test]
+        public void MockFile_Create_ShouldWorkWithRelativePath()
+        {
+            var relativeFile = "file.txt";
+            var fileSystem = new MockFileSystem();
+
+            fileSystem.File.Create(relativeFile).Close();
+
+            Assert.That(fileSystem.File.Exists(relativeFile));
+        }
     }
 }
--- a/System.IO.Abstractions.TestingHelpers.Tests/MockFileCreateTests.cs
+++ b/System.IO.Abstractions.TestingHelpers.Tests/MockFileCreateTests.cs
@@ -155,12 +155,12 @@ namespace System.IO.Abstractions.TestingHelpers.Tests
         {
             // Arrange
             var fileSystem = new MockFileSystem();
+            var file = XFS.Path("C:\\path\\NotFound.ext");
 
             // Act
-            TestDelegate action = () => fileSystem.File.Create("C:\\Path\\NotFound.ext");
+            TestDelegate action = () => fileSystem.File.Create(file);
 
             // Assert
-            Assert.IsFalse(fileSystem.Directory.Exists("C:\\path"));
             var exception = Assert.Throws<DirectoryNotFoundException>(action);
             Assert.That(exception.Message, Does.StartWith("Could not find a part of the path"));
         }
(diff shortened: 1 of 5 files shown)


fgreinacher:
@rkoeninger @jpreese @robkeim 👀❓

System.IO.Abstractions.TestingHelpers.Tests/MockFileCreateTests.cs
```diff
@@ -155,12 +155,12 @@ public void MockFile_Create_ShouldThrowDirectoryNotFoundExceptionIfCreatingAndPa
         {
             // Arrange
             var fileSystem = new MockFileSystem();
+            var file = XFS.Path("C:\\path\\NotFound.ext");
```
fgreinacher: This test was not normalizing the path which caused test failures on Unix, so I fixed this.

System.IO.Abstractions.TestingHelpers.Tests/MockFileCreateTests.cs
```diff
@@ -155,12 +155,12 @@ public void MockFile_Create_ShouldThrowDirectoryNotFoundExceptionIfCreatingAndPa
...
 
             // Act
-            TestDelegate action = () => fileSystem.File.Create("C:\\Path\\NotFound.ext");
+            TestDelegate action = () => fileSystem.File.Create(file);
 
             // Assert
-            Assert.IsFalse(fileSystem.Directory.Exists("C:\\path"));
```
fgreinacher: This doesn't belong in this test case so I decided to remove it.

System.IO.Abstractions.TestingHelpers.Tests/MockFileOpenTests.cs
```diff
@@ -249,14 +249,25 @@ public void MockFile_Open_ShouldThrowDirectoryNotFoundExceptionIfFileModeCreateA
         {
             // Arrange
             var fileSystem = new MockFileSystem();
+            var file = XFS.Path("C:\\path\\NotFound.ext");
```
fgreinacher: Same comment as above

System.IO.Abstractions.TestingHelpers.Tests/MockFileOpenTests.cs
```diff
@@ -249,14 +249,25 @@ public void MockFile_Open_ShouldThrowDirectoryNotFoundExceptionIfFileModeCreateA
...
 
             // Act
-            TestDelegate action = () => fileSystem.File.Open("C:\\Path\\NotFound.ext", FileMode.Create);
+            TestDelegate action = () => fileSystem.File.Open(file, FileMode.Create);
 
             // Assert
-            Assert.IsFalse(fileSystem.Directory.Exists("C:\\path"));
```
fgreinacher: Same comment as above

rkoeninger:
@fgreinacher Looks good to me!

PR #410, opened 2018-11-30
Deploy only from release branch

fgreinacher:
The overall idea I have is to trigger Nuget.org push and GitHub release creation by merging changes from `master` to `release` via a PR. This will reduce the number of GitHub releases and package versions on NuGet.org. Also I hope that AppVeyor will then be able to create release notes containing all changes since the latest release.

Thoughts?


fgreinacher:
OK, this is now ready for :eyes:

fgreinacher:
@jpreese @robkeim @ericnewton76 Could you have a look? I'd like to deploy 3.0 (#408) and would like to have this in (or out 😄) before.

ericnewton76:
Sorry to come late to this party.

I've found to generate a release with a new tag is best.  A new tag push indicates an actual release, and the release version is directly dictated by the tag.  

Triggering a release from a branch commit creates chicken and egg problem with the release version... Its difficult to lock down the actual release version without a tag.

PR #422, opened 2018-12-15
Fix MockDirectoryInfo.GetFiles for UNC paths

DeveloperGuo:
Regression introduced in v3.0.2 by #418 StringExtensions.NormalizeSlashes.  
Unc paths are getting stripped to a single separator.


fgreinacher:
Thanks, this looks good to me.

@rkoeninger, additional pair of eyes?

rkoeninger:
@fgreinacher I must be missing something obvious, but I'm confused.

I'm not sure why the `>=` helps. The existing logic was:

> if there's more than 2, cut it down to just 2

now it's:

> if there's 2 or more, cut it down to just 2

@DeveloperGuo How was the existing code trimming it to 1 slash when the statement is `.Substring(0, 2)`, and that's unchanged?

rkoeninger:
Instead of 

```csharp
            // UNC Paths start with double slash but no reason
            // to have more than 2 slashes at the start of a path
            if (XFS.IsWindowsPlatform() && prefixSeps.Length >= 2)
            {
                prefixSeps = prefixSeps.Substring(0, 2);
            }
            else if (prefixSeps.Length > 1)
            {
                prefixSeps = prefixSeps.Substring(0, 1);
            }
```

can we make it something like

```csharp
            // UNC Paths start with double slash but no reason
            // to have more than 2 slashes at the start of a path
            if (XFS.IsWindowsPlatform() && prefixSeps.Length >= 2)
            {
                prefixSeps = doubleSep;
            }
            else if (prefixSeps.Length == 1)
            {
                prefixSeps = sep;
            }
            else
            {
                prefixSeps = "";
            }
```

Otherwise, it appears that `doubleSep` variable is unused. I was going to make it a ternay, but a doubly-nested ternary might be a little much.

rkoeninger:
@fgreinacher Does that `while (true)` look sketchy to you? It could be `for (var i = 0; i < 100; ++i)` to prevent infinite loop. That could be implemented more cleanly, too.

Single-pass slash de-duplicate logic:

```csharp
var builder = new StringBuilder(path.Length);
var previousWasSep = false;

foreach (var ch in path)
{
    if (ch != sep || !previousWasSep)
    {
        builder.Append(ch);
    }

    previousWasSep = ch == sep;
}

path = builder.ToString();
```

DeveloperGuo:
> @DeveloperGuo Instead of
> 
> ```cs
>             // UNC Paths start with double slash but no reason
>             // to have more than 2 slashes at the start of a path
>             if (XFS.IsWindowsPlatform() && prefixSeps.Length >= 2)
>             {
>                 prefixSeps = prefixSeps.Substring(0, 2);
>             }
>             else if (prefixSeps.Length > 1)
>             {
>                 prefixSeps = prefixSeps.Substring(0, 1);
>             }
> ```
> 
> can we make it something like
> 
> ```cs
>             // UNC Paths start with double slash but no reason
>             // to have more than 2 slashes at the start of a path
>             if (XFS.IsWindowsPlatform() && prefixSeps.Length >= 2)
>             {
>                 prefixSeps = doubleSep;
>             }
>             else if (prefixSeps.Length == 1)
>             {
>                 prefixSeps = sep;
>             }
>             else
>             {
>                 prefixSeps = "";
>             }
> ```
> 


@rkoeninger -  The prefixSep.Length == 0 and 1 cases are already handled by the prefixSeps initialization.  

`   var prefixSeps = new string(path.TakeWhile(c => c == Path.DirectorySeparatorChar).ToArray());
            path = path.Substring(prefixSeps.Length);
`
Based on the comment above the if/else, the intention here was to trim paths starting with more than 1 separator for non-UNC paths.  Your suggestion would break this in the case of !XFS.IsWindowsPlatform() && prefixSeps.Length == 2

`           // UNC Paths start with double slash but no reason
            // to have more than 2 slashes at the start of a path
            if (XFS.IsWindowsPlatform() && prefixSeps.Length >= 2)
            {
                prefixSeps = prefixSeps.Substring(0, 2);
            }
            else if (prefixSeps.Length > 1)
            {
                prefixSeps = prefixSeps.Substring(0, 1);
            }`

It may make sense to create separate pull request(s) for the other issue(s) brought up as this one is specifically addressing the regression to UNC paths.

rkoeninger:
> The prefixSep.Length == 0 and 1 cases are already handled by the prefixSeps initialization

@DeveloperGuo right, that `else if` is never going to be true now.

Your fix should work, so it's fine as is, I'm just started thinking out loud.

rkoeninger:
👍 Go for it

fgreinacher:
Thanks @DeveloperGuo!

Issue #274, opened 2018-02-08
Cannot set properties on DriveInfoBase objects returned by MockFileSystem.DriveInfo.GetDrives


z4ns4tsu:
The `MockDriveInfo` type allows the caller to set various properties on instances of the class (e.g. `AvailableFreeSpace` or `DriveType`), but these properties are not included in the objects output by a call to `MockFileSystem.DriveInfo.GetDrives()` since the array is built from the list of files in the mock without checking if a `MockDriveInfo` object is defined.

*Creating the mock*
```CSharp
var fileSystem = new MockFileSystem();
fileSystem.AddFile(@"C:\path\to\file.txt", MockFileData.NullObject);
var driveInfo = new MockDriveInfo(fileSystem, "c:") { DriveType = DriveType.Fixed };
```

*Checking the drive is local*
```CSharp
List<string> localDiskDrives = 
(
    from driveInfo in fileSystem.DriveInfo.GetDrives()
    where driveInfo.DriveType == DriveType.Fixed & driveInfo.IsReady
    select driveInfo.RootDirectory.ToString()
).ToList();
```

*Expected*
List containing "c:"

*Actual*
Empty list

sharpSteff:
I stumbled over this bug, when trying to set `AvailableFreeSpace` of the MockDriveInfo

```cs
// Set some free space
(mockFileSystem.DriveInfo.GetDrives().First() as MockDriveInfo).AvailableFreeSpace = 1024 * 1024;
var availableSpace = (mockFileSystem.DriveInfo.GetDrives().First() as MockDriveInfo).AvailableFreeSpace; 
````
After setting the `AvailableFreeSpace` and trying to get `AvailableFreeSpace` after its '0'

fgreinacher:
Hm, I wouldn't consider this a bug, but rather a bad API. The properties at `MockDriveInfo` shouldn't be set-able. Allowing such a feature would force us to somehow maintain the used space so that we could check it against the available space on each write operation.

Other thoughts?

ThoMAsThe:
I think it's possible if you do it through the constructor directly. Something like this would allow you to set the DriveType:

`        private class MockDriveInfo : DriveInfoBase
        {
            public MockDriveInfo(IFileSystem fileSystem, DriveType driveType) : base(fileSystem)
            {
                base.DriveType = driveType;
            }
        }`

That means you can set the DriveType as such (using Moq for FileSystem):

[...]

PR #423, opened 2018-12-15
Fix for bug #274

updateaman:
Fix for bug #274 

I've tried to have a go at resolving this bug by caching the DriveInfoBase objects
Please let me know if that works


fgreinacher:
Thanks a lot for your work on this. After reading the original issue again I don't think we should support this kind of functionality. Let's continue the discussion on the issue and see how it goes.

jpreese:
I am going to agree with @fgreinacher on this one. Closing PR. If you wish to continue the discussion, please use the originating ticket or create a new one.

PR #427, opened 2018-12-17
Make sure FileNotFound exceptions contain path and proper message

### LINKED ISSUE #424: FileNotFoundException thrown by MockFileSystem and by real System.IO doesn't match
This unit test fails:

```
[TestMethod]
public void TestFileNotFoundExceptionContent() {
	var expectedMsg = "";
	var expectedFileName = "";
	var actualMsg = "";
	var actualFileName = "";

	var absentFileNameFullPath = @"c:\you surely don't have such file.hope-so";
	try {
		File.ReadAllText(absentFileNameFullPath);
	} catch (FileNotFoundException e) {
		expectedMsg = e.Message;
		expectedFileName = e.FileName;
	}

	try {
		_fileSystemMock.File.ReadAllText(absentFileNameFullPath);
	} catch (FileNotFoundException e) {
		actualMsg = e.Message;
		actualFileName = e.FileName;
	}
	
	// Expected:	@"Could not find file 'c:\you surely don't have such file.hope-so'."
	// Actual:	@"Can't find c:\you surely don't have such file.hope-so"
	Assert.AreEqual(expectedMsg, actualMsg);
	// Expected:	@"c:\you surely don't have such file.hope-so"
	// Actual:	null
	Assert.AreEqual(expectedFileName, actualFileName);
}
```

fgreinacher:
Fixes #424


rkoeninger:
👍 On the `CommonExceptions`.

That should eventually be used everywhere so there isn't any inconsistent `string.Format` strings or sometimes there's a string in `StringResources` and sometimes there are in-line string literals.

fgreinacher:
OK, I think now all commonly used exceptions are handled centrally. There are a lot of places where we don't use strings from resources or do stuff otherwise inconsistently, but I don't want to too much in this PR, so I'll leave this work for someone else :)

fgreinacher:
@rkoeninger maybe one more quick 👀 ?

rkoeninger:
Looks good!

fgreinacher:
Fixed in https://www.nuget.org/packages/System.IO.Abstractions/3.0.10

Issue #429, opened 2018-12-19
DirectoryInfoWrapper.Parent throws Exception for root directory (but should return null instead)


wexman:
Title pretty much says it all.

fgreinacher:
Fixed in https://www.nuget.org/packages/System.IO.Abstractions/3.0.10

PR #430, opened 2018-12-19
Fix (and test) for #429

wexman:


fgreinacher:
Thanks a lot!

--- 
Technically the test is wrong here (because the test assembly is targeting the testing helpers, not the wrappers). But we don't have any other place for these kinds of tests, so it's fine as-is (at least for now).

wexman:
> Technically the test is wrong here (because the test assembly is targeting the testing helpers, not the wrappers). But we don't have any other place for these kinds of tests, so it's fine as-is (at least for now).

I figured that, but would leave that up to you ;-)

PR #432, opened 2018-12-23
Pass IFileSystem into FileWrapper instead of FileSystem

kirbatious:
This will allow replacement file systems to be used and is the same pattern used in all the other classes.


fgreinacher:
Good catch, thanks!

fgreinacher:
Fixed in https://www.nuget.org/packages/System.IO.Abstractions/3.0.10

PR #188, opened 2016-11-23
#187: Directory move shoud be able to move files, not just directories

AlexanderPruss:

PR #438, opened 2019-01-07
Enable MockDirectory.Move to also move files

### LINKED ISSUE #187: Directory.Move() throws exceptions when trying to move a file, not a directory.
Directory.Move() should be able to move files: https://msdn.microsoft.com/en-us/library/system.io.directory.move(v=vs.110).aspx

The current implementation of MockDirectory, however, assumes that the moved object is a directory and throws an exception.

System.IO.Abstractions.TestingHelpers.dll!System.IO.Abstractions.TestingHelpers.MockDirectory.EnumerateDirectories(string path, string searchPattern, System.IO.SearchOption searchOption)
System.IO.Abstractions.TestingHelpers.dll!System.IO.Abstractions.TestingHelpers.MockDirectory.GetDirectories(string path, string searchPattern, System.IO.SearchOption searchOption)
System.IO.Abstractions.TestingHelpers.dll!System.IO.Abstractions.TestingHelpers.MockDirectory.GetDirectories(string path, string searchPattern) 
System.IO.Abstractions.TestingHelpers.dll!System.IO.Abstractions.TestingHelpers.MockDirectory.GetDirectories(string path)
System.IO.Abstractions.TestingHelpers.dll!System.IO.Abstractions.TestingHelpers.MockDirectory.Move(string sourceDirName, string destDirName)

fgreinacher:
Fixes #187
Supersedes #188

### DIFF
--- a/System.IO.Abstractions.TestingHelpers.Tests/MockDirectoryTests.cs
+++ b/System.IO.Abstractions.TestingHelpers.Tests/MockDirectoryTests.cs
@@ -1164,7 +1164,7 @@ namespace System.IO.Abstractions.TestingHelpers.Tests
         }
 
         [TestCaseSource("GetPathsForMoving")]
-        public void MockDirectory_Move_ShouldMove(string sourceDirName, string destDirName, string filePathOne, string filePathTwo)
+        public void MockDirectory_Move_ShouldMoveDirectories(string sourceDirName, string destDirName, string filePathOne, string filePathTwo)
         {
             // Arrange
             var fileSystem = new MockFileSystem(new Dictionary<string, MockFileData>
@@ -1180,6 +1180,27 @@ namespace System.IO.Abstractions.TestingHelpers.Tests
             Assert.IsFalse(fileSystem.Directory.Exists(sourceDirName));
             Assert.IsTrue(fileSystem.File.Exists(XFS.Path(destDirName + filePathOne)));
             Assert.IsTrue(fileSystem.File.Exists(XFS.Path(destDirName + filePathTwo)));
+        }
+
+        [Test]
+        public void MockDirectory_Move_ShouldMoveFiles()
+        {
+            string sourceFilePath = XFS.Path(@"c:\something\demo.txt");
+            string sourceFileContent = "this is some content";
+            
+            var fileSystem = new MockFileSystem(new Dictionary<string, MockFileData>
+            {
+                { sourceFilePath, new MockFileData(sourceFileContent) },
+                { XFS.Path(@"c:\somethingelse\dummy.txt"), MockFileData.NullObject }
+            });
+
+            string destFilePath = XFS.Path(@"c:\somethingelse\demo1.txt");
+
+            fileSystem.Directory.Move(sourceFilePath, destFilePath);
+
+            Assert.That(fileSystem.FileExists(destFilePath), Is.True);
+            Assert.That(fileSystem.GetFile(destFilePath).TextContents, Is.EqualTo(sourceFileContent));
+            Assert.That(fileSystem.FileExists(sourceFilePath), Is.False);
         }
 
         [Test]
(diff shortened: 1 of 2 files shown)


fgreinacher:
:eyes: anyone?

System.IO.Abstractions.TestingHelpers.Tests/MockDirectoryTests.cs
```diff
@@ -1182,6 +1182,27 @@ public void MockDirectory_Move_ShouldMove(string sourceDirName, string destDirNa
...
+        [Test]
+        public void MockDirectory_Move_ShouldMoveFiles()
+        {
+            string sourceFilePath = XFS.Path(@"c:\something\demo.txt");
+            string sourceFileContent = "this is some content";
+            
+            var fileSystem = new MockFileSystem(new Dictionary<string, MockFileData>
+            {
+                { sourceFilePath, new MockFileData(sourceFileContent) },
+                { XFS.Path(@"c:\somethingelse\dummy.txt"), MockFileData.NullObject }
```
jpreese: I think this test could be simplified a bit. The focus here is files, so we could probably just do something like C:\demo.txt, and drop the directory. There also more than likely isn't a need to include two files (I'm seeing dummy.txt and demo.txt). We just need to make sure the file gets from Point A to Point B. A rename of the file could achieve the same effect as moving directories.

System.IO.Abstractions.TestingHelpers.Tests/MockDirectoryTests.cs
```diff
@@ -1182,6 +1182,27 @@ public void MockDirectory_Move_ShouldMove(string sourceDirName, string destDirNa
...
+            });
+
+            string destFilePath = XFS.Path(@"c:\somethingelse\demo1.txt");
+
+            fileSystem.Directory.Move(sourceFilePath, destFilePath);
+
+            Assert.That(fileSystem.FileExists(destFilePath), Is.True);
+            Assert.That(fileSystem.GetFile(destFilePath).TextContents, Is.EqualTo(sourceFileContent));
+            Assert.That(fileSystem.FileExists(sourceFilePath), Is.False);
```
jpreese: I'm usually against multiple asserts, but since this is more about state after a move, it seems OK. I had a slight issue with this because it was initially kinda confusing. Maybe we move the FileExists check to the top so we're validating state together? i.e. FileExists destFile (true) and FileExists sourceFile (false) need to be together to validate the file moved on the file system.  The GetFile makes sure the source and dest were the same.

System.IO.Abstractions.TestingHelpers/MockDirectory.cs
```diff
@@ -378,6 +378,14 @@ public override void Move(string sourceDirName, string destDirName)
                 throw new IOException("Source and destination path must be different.");
             }
 
+            //if we're moving a file, not a directory, call the appropriate file moving function.
+            var fileData = mockFileDataAccessor.GetFile(fullSourcePath);
+            if (fileData != null && (fileData.Attributes & FileAttributes.Directory) == 0)
```
jpreese: I think a safe navigation operator (?.) might simplify this conditional a little?

jpreese:
Overall I think it's fine, just a couple ease of understanding changes.

System.IO.Abstractions.TestingHelpers.Tests/MockDirectoryTests.cs
fgreinacher: Yeah, the second file was there to ensure the destination directory exists. I agree there was too much going on, so I'll simplify this as suggested. I won't rename the file because the focus of this test is the special behavior of `MockDirectory.Move`.

System.IO.Abstractions.TestingHelpers.Tests/MockDirectoryTests.cs
fgreinacher: Good catch, will do

System.IO.Abstractions.TestingHelpers/MockDirectory.cs
fgreinacher: Sure, changed this line to use safe navigation operator and `Enum.HasFlag`. Reads a lot easier now.

fgreinacher:
Feedback addressed with e74649dedd39f01b1fe529d0b2692a65171efbaf

Issue #226, opened 2017-05-25
Ability to lock a MockFile for testing


gregmcguffey:
It would be really helpful to mark a MockFile as being locked, so locked file logic can easily be tested. This would related to commands like File.Delete, File.WriteAll*, ect. Or is this possible now?

tathamoddie:
This isn't currently implemented in the testing helpers.

The fake file system doesn't track if files have been opened exclusively or not:

https://github.com/tathamoddie/System.IO.Abstractions/blob/master/TestingHelpers/MockFile.cs#L361

Sounds like an interesting pull request though if you care to start coding something up! :)

erenes:
Hi guys, one of my customers reported an issue with file locking, so I am implementing it in my forked repo. Currently my approach is to add a FileShare property to the Mockfile that is checked when opening a stream or using one of the File methods.

I considered implementing it with a whole dictionary of streams that are using different FileAccess/FileShare parameters, but for my scenario that seems like a lot of wasted effort. 

To illustrate, this is one of the tests: 

        [Test]
        public void MockFile_Lock_FileShareNoneThrowsOnOpenRead()
        {
            string filepath = XFS.Path(@"c:\something\does\exist.txt");
            var filesystem = new MockFileSystem(new Dictionary<string, MockFileData>
            {
                { filepath, new MockFileData("I'm here") { CurrentFileShare = FileShare.None }}
            });

            Assert.Throws(typeof(IOException), () => filesystem.File.OpenRead(filepath));
        }

I am not entirely satisfied about the name "CurrentFileShare". Compared to writing a full File System simulation, I do think this way of marking locked files is easier to implement both in System.IO.Abstractions, as well as when writing unit tests. 

Any feedback / thoughts is welcome :)

fgreinacher:
This is released with https://github.com/System-IO-Abstractions/System.IO.Abstractions/releases/tag/v3.1.1

Issue #442, opened 2019-01-21
Please Enable TestingHelpers to throw Exceptions


Rainer1:
It quite some scenarios it is necessary to ensure by UnitTests, that the written code handles Exceptions correctly which are thrown by the file system.
For this reason it would be great to be able to define a Exception to be thrown, when a specific Method Is called.
Example Similar to many Mocking Frameworks:
`MockFileSystem.File.Delete.WhenCalled( (file) => throw new IOException(message))`

fgreinacher:
Duplicate of #168 - consider leaving your thumbs up there to show support :)

PR #445, opened 2019-02-19
Support for locked files

erenes:
Solves #226, but needs a review to see if I missed a scenario.

### DIFF
--- a/System.IO.Abstractions.TestingHelpers/MockFile.cs
+++ b/System.IO.Abstractions.TestingHelpers/MockFile.cs
@@ -177,6 +177,12 @@ namespace System.IO.Abstractions.TestingHelpers
             // but silently returns if deleting a non-existing file in an existing folder.
             VerifyDirectoryExists(path);
 
+            var file = mockFileDataAccessor.GetFile(path);
+            if (file != null && !file.AllowedFileShare.HasFlag(FileShare.Delete))
+            {
+                throw new IOException($"The process cannot access the file '{path}' because it is being used by another process.");
+            }
+
             mockFileDataAccessor.RemoveFile(path);
         }
 
--- a/System.IO.Abstractions.TestingHelpers/MockFile.cs
+++ b/System.IO.Abstractions.TestingHelpers/MockFile.cs
@@ -180,7 +180,7 @@ namespace System.IO.Abstractions.TestingHelpers
             var file = mockFileDataAccessor.GetFile(path);
             if (file != null && !file.AllowedFileShare.HasFlag(FileShare.Delete))
             {
-                throw new IOException($"The process cannot access the file '{path}' because it is being used by another process.");
+                throw CommonExceptions.ProcessCannotAccessFileInUse(path);
             }
 
             mockFileDataAccessor.RemoveFile(path);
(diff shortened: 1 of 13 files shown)


erenes:
Second commit should help with #442, if AllowedFileShare is set to None for a file, File.Delete now throws an IOException as in regular System.IO

fgreinacher:
Thanks for contributing this back!

I am a bit unsure about the approach of controlling the behavior solely through `MockFileData` - an alternative approach would be to (additionally) handle this by the regular APIs where you lock a file by passing `FileShare.None`. I see that it might involve some more implementation effort, but on the other side it would really be somehow unintuitive to support locking a file only via test APIs.

So my specific proposal would be to keep the property as convenience layer and additionally support locking via `FileShare` parameters passed to the various methods that somehow open files.

Other thoughts @jpreese @robkeim @tathamoddie?

erenes:
Hi Florian, thanks for the feedback! 

The only place where I expect this to require changes is in File.OpenInternal. When we open the FileStream, we already check the AllowedFileShare property to see if what we're doing is allowed. 

It's trivial to add a set of the current FileShare option if successful: 

            var mockFileData = mockFileDataAccessor.GetFile(path);
            mockFileData.CheckFileAccess(path, access);
            mockFileData.AllowedFileShare = share; // this would do it

            var length = mockFileData.Contents.Length;
            MockFileStream.StreamType streamType = MockFileStream.StreamType.WRITE;
            if (access == FileAccess.Read)

The tricky part is if you have several file streams open, and close one of them, what should the FileShare setting on the file then be afterwards? We will have to keep track of it somehow, and find out what the expected behavior of all combinations are. 

Example: 
What is the FileSharing mode after running the code below? Read, ReadWrite, reset to no lock at all? 

      if (File.Exists(path)) File.Delete(path);
      File.WriteAllText(path, "hello world");

      FileStream fs = File.Open(path, FileMode.Open, FileAccess.Read, FileShare.ReadWrite);
      FileStream fs2 = File.Open(path, FileMode.Open, FileAccess.ReadWrite, FileShare.Read);
      FileStream fs3 = File.Open(path, FileMode.Open, FileAccess.Read, FileShare.ReadWrite);
      fs.Close();
      fs.Dispose();

While not impossible to implement, I think it would add a lot of complexity to the code. I decided to keep it simple (for now) because for the purposes of a unit test, you probably want to make sure that your tested function handles externally locked files properly. 

I'm happy to be convinced otherwise, but I don't really see the need to verify that my tested code path has a contested file lock with itself in a unit test?

fgreinacher:
Thanks for the argumentation. You're right, keeping track of the intermediate states would add quite some complexity. 

So let's start with this simple solution and wait for feedback.

System.IO.Abstractions.TestingHelpers/MockFile.cs
```diff
@@ -354,7 +361,10 @@ public override void Move(string sourceFileName, string destFileName)
             {
                 throw CommonExceptions.FileNotFound(sourceFileName);
             }
-
+            if (!sourceFile.AllowedFileShare.HasFlag(FileShare.Delete))
+            {
+                throw new IOException("The process cannot access the file because it is being used by another process.");
```
fgreinacher: Please move this to `CommonExceptions`

erenes: Moved this exception, and also the one that includes the file name of the file in use to the CommonExceptions class.

fgreinacher:
Merged, thanks for your contribution!

Issue #440, opened 2019-01-12
MockFileSystem.FileInfo.FromFileName("some\\directory") has Length property set to '0'


Xerillio:
I noticed there's a discrepancy between `FileSystem` and `MockFileSystem` when returning a `FileInfoBase` for a directory path. Trying to access the `FileInfoBase.Length` property for an object returned by `FileSystem` results in a FileNotFoundException while the object from `MockFileSystem` is just 0.

It seems the `MockFileSystem` should also throw an exception to follow the .NET behaviour.

fgreinacher:
Thanks for reporting!

PatrykOlejniczak:
I think we can close this issue. It's fixed :)

fgreinacher:
Yeah, will close this once we release vNext!

fgreinacher:
This is released with https://github.com/System-IO-Abstractions/System.IO.Abstractions/releases/tag/v3.1.1

PR #448, opened 2019-02-24
Fix Length and Exists properties in MockFileInfo

PatrykOlejniczak:
Fix issue #440 , also when doing that issue notice Exists property is set to true when we set path to directory. (should be false as in normal FileInfo).


fgreinacher:
Looks great, thanks a lot for your contribution!

Issue #175, opened 2016-10-17
Consider API Adjustment to Use Interfaces Rather Than Abstract Classes


Mike-EEE:
Cool project. :)  Thank you for keeping it current.

Checking this out just now.  It appears there are files such as `DirectoryInfoBase` and `FileInfoBase`.  Has there been any thought towards making these interfaces instead, So `IDirectoryInfo` and `IFileInfo`?  That way we're really abstracting the goodness and don't have to pass implementations around.

Thank you for any consideration!

DavidArno:
Just found a link to this, looked at the code and saw the same thing and thought the same thing.

This could be a really good project, but it's difficult to take it seriously when it used abstract classes and inheritance, rather than interfaces. I'm happy to fix and do a PR if you'd like?

MCCoolMoDAue:
I looked at some classes, and mostly the abstract classes are unnecessary.
Introducing interfaces should be straight forward.

fgreinacher:
Implemented by #452

PR #452, opened 2019-03-06
Added IFileSystemWatcher, IFile and IDirectory interfaces

wexman:
I have added the IFileSystemWatcher, IFile and IDirectory interface for easier mocking...

### DIFF
--- a/System.IO.Abstractions/IDirectory.cs
+++ b/System.IO.Abstractions/IDirectory.cs
@@ -1,0 +1,58 @@
+﻿using System.Collections.Generic;
+using System.Security.AccessControl;
+
+namespace System.IO.Abstractions
+{
+    public interface IDirectory
+    {
+        IFileSystem FileSystem { get; }
+
+        DirectoryInfoBase CreateDirectory(string path);
+        void Delete(string path);
+        void Delete(string path, bool recursive);
+        IEnumerable<string> EnumerateDirectories(string path);
+        IEnumerable<string> EnumerateDirectories(string path, string searchPattern);
+        IEnumerable<string> EnumerateDirectories(string path, string searchPattern, SearchOption searchOption);
+        IEnumerable<string> EnumerateFiles(string path);
+        IEnumerable<string> EnumerateFiles(string path, string searchPattern);
+        IEnumerable<string> EnumerateFiles(string path, string searchPattern, SearchOption searchOption);
+        IEnumerable<string> EnumerateFileSystemEntries(string path);
+        IEnumerable<string> EnumerateFileSystemEntries(string path, string searchPattern);
+        IEnumerable<string> EnumerateFileSystemEntries(string path, string searchPattern, SearchOption searchOption);
+        bool Exists(string path);
+        DirectorySecurity GetAccessControl(string path);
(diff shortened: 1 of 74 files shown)


fgreinacher:
This would implement #175, though I'm not sure if a breaking API change like that would produce more negative than positive echo. 

@jpreese @robkeim What's your opinion?

wexman:
That might indeed be a breaking change, but a rather theoretical one, meaning it won't actually break because the interfaces have the same members as the abstract classes...

System.IO.Abstractions/IDirectory.cs
```diff
@@ -0,0 +1,58 @@
...
+        void SetLastAccessTime(string path, DateTime lastAccessTime);
+        void SetLastAccessTimeUtc(string path, DateTime lastAccessTimeUtc);
+        void SetLastWriteTime(string path, DateTime lastWriteTime);
+        void SetLastWriteTimeUtc(string path, DateTime lastWriteTimeUtc);
+
+#if NET40
+        DirectoryInfoBase CreateDirectory(string path, DirectorySecurity directorySecurity);
+        string[] GetLogicalDrives();
+#endif
+
```
robkeim: Minor: extra newline here

robkeim:
I also agree with the direction that #175 wants to go, how difficult do you think the breaking changes would be for people consuming the library to fix in their solutions @fgreinacher?

If you're using the `var` keyword the change shouldn't be breaking for local variables inside functions:
```
var fileSystemWatcherFactory = new FileSystemWatcherFactory();
var fileSystemWatcher = fileSystemWatcherFactory.CreateNew(); // No breaking change here changing from FileSystemWatcherBase to IFileSystemWatcher
```

For instances where there are breaking changes (explicitly typed local variables, function parameters, function return type, etc, is a simple replace of `XyzBase` to `IXyz` sufficient for all the types that introduced the interface? That's definitely a naive approach, but I can't think of situations off the top of my head where doing so wouldn't be sufficient or would introduce other issues.

If that's the case then I don't think we should be too worried about this breaking change.

wexman:
BTW, do you have any idea why the comments from my previous pull request show up here? Did i do something wrong?

fgreinacher:
> BTW, do you have any idea why the comments from my previous pull request show up here? Did i do something wrong?

(I assume comments == commits) That's just happening because we are updating your branch from `master`. It's fine :)

fgreinacher:
OK - let's do this. A couple of things that I'd ask you to do:

1. Update the version in `version.json` to 4.0 (as this is a breaking change)
1. Add `<inheritdoc>`comments to the new interfaces (just mirror what's done in the implementations) so that we get proper XML docs.
1. Think about whether it makes sense to add a short chapter to `README.md` that describes the new use case this change enables

wexman:
Oh, ok. But if we're going that route anyway, maybe we should do the same for all other cases as well? For example, IPath instead of PathBase, IFileInfo instead of FileInfoBase, IDirectoryInfo instead of DirectoryInfoBase and so on?

robkeim:
@wexman Yes I think it definitely makes sense to have a consistent API

wexman:
Ok, it's done...

fgreinacher:
Merged with some adaptions.  Thanks a lot for your work @wexman!

PR #456, opened 2019-03-11
 Added File Access check to MockFileStream

### LINKED ISSUE #455: AllowedFileShare not obeyed by MockFileStream
Version 3.1.1

`MockFileStream` does not seem to be obeying the `AllowedFileShare` property. See the example below, I would expect to see an exception throw when trying to create the FileStream. 

```csharp
using System;
using System.IO;
using System.Text;

namespace FileStreamTest
{
    class Program
    {
        static void Main(string[] args)
        {
            IDictionary<string, MockFileData> files  = new Dictionary<string, MockFileData>
            {
                { "C:\\test\\file.txt", new MockFileData("contents") { AllowedFileShare = FileShare.None } }
            };
            
            var fs = new MockFileSystem(files);
            
            // Does not throw exception
            var stream = fs.FileStream.Create("C:\\test\\file.txt", FileMode.Open, FileAccess.Read, FileShare.Read);
            using (stream)
            {
                using (var reader = new StreamReader(stream, Encoding.Default))
                {
                    string line;
                    while ((line = reader.ReadLine()) != null)
                    {
                        Console.Write(line);
                    }
                }
            }
           
            Console.Write(Environment.NewLine);
           
            // Throws System.IO.IOException as expected
            var text = fs.File.ReadAllText("C:\\test\\file.txt");
            Console.WriteLine(text);
        }
    }
}
```

roketworks:
Fixes #455 

Added access check to `MockFileStream`. Will always require Read flag, but if stream is write-able then additional access check will be completed.

### DIFF
--- a/System.IO.Abstractions.TestingHelpers.Tests/MockFileStreamTests.cs
+++ b/System.IO.Abstractions.TestingHelpers.Tests/MockFileStreamTests.cs
@@ -76,6 +76,46 @@
             Assert.IsFalse(stream.CanWrite);
             Assert.Throws<NotSupportedException>(() => stream.WriteByte(1));
         }
+        
+        [Test]
+        [TestCase(FileShare.None, MockFileStream.StreamType.READ)]
+        [TestCase(FileShare.None, MockFileStream.StreamType.WRITE)]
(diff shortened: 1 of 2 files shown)


fgreinacher:
Thanks for reporting and fixing!

Issue #461, opened 2019-04-02
Testing Helpers - FileAccess IOException does not have HResult that indicates file lock


roketworks:
The HResult value for the IOException thrown from the `CommonExceptions` for `ProcessCannotAccessFileInUse` does not indicate that the file is locked. 

See for error code reference: https://docs.microsoft.com/en-us/windows/desktop/debug/system-error-codes--0-499-

0x80070020 or 0x80070021 would indicate Sharing/Lock violations

fgreinacher:
This is fixed with https://www.nuget.org/packages/System.IO.Abstractions/4.1.6

PR #462, opened 2019-04-02
Added HResult value for ERROR_SHARING_VIOLATION to File in use exception

roketworks:
This resolves issue #461


fgreinacher:
Thanks! Please add unit tests and we‘re good to go!

roketworks:
> Thanks! Please add unit tests and we‘re good to go!

I've added to `MockFileLockTests`, so where `IOException` is returned it will check to make to make sure the `HResult` property is correct. I thought this made more sense than a single test that only tests the exception under 1 condition.

fgreinacher:
Perfect, thanks again!

Issue #460, opened 2019-04-02
MockFileStream.Create does not fail when directory does not exists


jbaehr:
When trying to `myFileSystem.FileStream.Create(@"someNotExistingDir\someFile")` I get a `DirectoryNotFoundException` when using a real file system (aka System.IO.FileStream), the mock file system does not reflect this behavior and silently creates the dir on the fly.

There has been a similar issue with `File.Open`, #227, that has been fixed. In System.IO.Abstractions.TestingHelpers v3.1.1 FileStream still suffers from this discrepancy between real and mock.

fgreinacher:
Good catch, thanks for reporting. PRs welcome :)

fgreinacher:
This is fixed with https://www.nuget.org/packages/System.IO.Abstractions/4.2.8. Thanks to @PatrykOlejniczak

PR #468, opened 2019-04-22
Throw DirectoryNotFoundException when try open FileStream for non-exiting path.

PatrykOlejniczak:
Refered to issue #460 

+ Throw DirectoryNotFoundException when try open FileStream for non-exiting path,
+ Added test for that case,
+ Fix existing test to pass after that change (directory init).


fgreinacher:
Thanks!

PR #470, opened 2019-05-09
Truncate existing files when writing via FileMode.Create/CreateNew

### LINKED ISSUE #465: MockFileStreamFactory.Create with FileMode.Create does not truncate an existing file
Found in version 3.0.10.

This is very similar, though distinct from, #357. Creating a stream with FileMode.Create is fine when the file does not exist, but does not truncate if the file exists. The following test should pass, but does not:

```
            const string FilePath = "File.txt";

            using(var stream = fileSystem.FileStream.Create(FilePath, System.IO.FileMode.Create, System.IO.FileAccess.Write))
            {
                using(var writer = new System.IO.StreamWriter(stream, System.Text.Encoding.UTF8, 4096, leaveOpen: true))
                {
                    writer.Write("1234567890");
                }
            }

            using(var stream = fileSystem.FileStream.Create(FilePath, System.IO.FileMode.Create, System.IO.FileAccess.Write))
            {
                using(var writer = new System.IO.StreamWriter(stream, System.Text.Encoding.UTF8, 4096, leaveOpen: true))
                {
                    writer.Write("AAAAA");
                }
            }

            string text = fileSystem.File.ReadAllText(FilePath);
            Assert.AreEqual("AAAAA", text); // text == "AAAAA67890"
```

Note that if you substitute FileMode.Truncate when opening the second stream, the test passes.

tomwilde:
Fixes #465

### DIFF
--- a/System.IO.Abstractions.TestingHelpers.Tests/MockFileStreamFactoryTests.cs
+++ b/System.IO.Abstractions.TestingHelpers.Tests/MockFileStreamFactoryTests.cs
@@ -46,6 +46,34 @@ namespace System.IO.Abstractions.TestingHelpers.Tests
 
         [Test]
         [TestCase(FileMode.Create)]
+        [TestCase(FileMode.CreateNew)]
+        public void TW___MockFileStreamFactory_CreateForAnExistingFile_ShouldTruncateExistingFile(FileMode fileMode)
+        {
+            var fileSystem = new MockFileSystem();
+            string FilePath = XFS.Path("C:\\File.txt");
+
+            using(var stream = fileSystem.FileStream.Create(FilePath, fileMode, System.IO.FileAccess.Write))
+            {
+                using(var writer = new System.IO.StreamWriter(stream, System.Text.Encoding.UTF8, 4096, leaveOpen: true))
+                {
+                    writer.Write("1234567890");
+                }
+            }
+
+            using(var stream = fileSystem.FileStream.Create(FilePath, fileMode, System.IO.FileAccess.Write))
+            {
+                using(var writer = new System.IO.StreamWriter(stream, System.Text.Encoding.UTF8, 4096, leaveOpen: true))
+                {
+                    writer.Write("AAAAA");
+                }
+            }
+
+            string text = fileSystem.File.ReadAllText(FilePath);
+            Assert.AreEqual("AAAAA", text); 
+        }
+
+        [Test]
+        [TestCase(FileMode.Create)]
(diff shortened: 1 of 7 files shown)


System.IO.Abstractions.TestingHelpers.Tests/MockFileStreamFactoryTests.cs
```diff
@@ -44,6 +44,34 @@ public void MockFileStreamFactory_CreateForNonExistingFile_ShouldReturnStream(Fi
...
+        [TestCase(FileMode.Create)]
+        [TestCase(FileMode.CreateNew)]
+        public void TW___MockFileStreamFactory_CreateForAnExistingFile_ShouldTruncateExistingFile(FileMode fileMode)
+        {
+            var fileSystem = new MockFileSystem();
+            string FilePath = XFS.Path("C:\\File.txt");
+
+            using(var stream = fileSystem.FileStream.Create(FilePath, fileMode, System.IO.FileAccess.Write))
+            {
+                using(var writer = new System.IO.StreamWriter(stream, System.Text.Encoding.UTF8, 4096, leaveOpen: true))
```
PatrykOlejniczak: There are problems in [appveyor ](https://ci.appveyor.com/project/tathamoddie/system-io-abstractions/builds/24463776/job/cu66awyi0m45so8d) tests because [NET40](https://docs.microsoft.com/pl-pl/dotnet/api/system.io.streamwriter.-ctor?view=netframework-4.0#System_IO_StreamWriter__ctor_System_String_System_Boolean_System_Text_Encoding_System_Int32_) didn't have constructor that support leaveOpen parameter.
To fix that you can make conditional scenario for `#if NET40` to get same react just don't create writer in using so [Dispose](https://stackoverflow.com/questions/2666888/is-there-any-way-to-close-a-streamwriter-without-closing-its-basestream) will not be run.

But please someone confirm that what i write make sense.

fgreinacher: I’d suggest just removing this argument, it does not seem to be necessary here. Same for the buffer size.

fgreinacher: Or even get rid of the StreamWriter and just write directly to the Stream.

tomwilde: Updated... :)

fgreinacher:
Thanks @tomwilde!

Issue #469, opened 2019-05-09
Potential MemLeak because IFileSystemWatcher does not implement IDisposable


BADF00D:
Hi,
I just stumble upon the fact, that IFileSystemWatcher is not IDisposable, but all its Implementation are. When I retrieve an instance from IFileSystemWatcherFactory, I have no hint that I have to dispose this instance. Which can lead to memory leaks. Given that FileSystemWatcher is a system resource, it might also be there after my program finished.

For me there are two solution to this problem:

1. Make IFileSystemWatcher IDisposable (my preferred solution).
2. Add a Release method to IFileSystemWatcherFactory, that handles the dispose call for me.

fgreinacher:
Thanks for reporting! I’d also prefer the first solution.

PR #479, opened 2019-05-31
Fixed Issue #469

BADF00D:
* Marked IFileSystemWatcher as IDisposable
* Removed IDisposable from FileSystemWatcherBase (because its now
  inherited)

I


fgreinacher:
Bumping the version to 5.0 because this is a breaking change for people implementing `IFileSystemWatcher`

fgreinacher:
Thanks!

PR #481, opened 2019-06-03
Implement Async File Methods

fedoranimus:
Please give any feedback - this may just be a starting point for implementing async methods.

Mocks are just passthroughs to sync methods.

Async methods are not available in .NET4 or .NET Standard, so added compilation contexts for .NET Core 2.0

### DIFF
--- a/System.IO.Abstractions.TestingHelpers/MockFile.cs
+++ b/System.IO.Abstractions.TestingHelpers/MockFile.cs
@@ -40,7 +40,7 @@ namespace System.IO.Abstractions.TestingHelpers
         }
 
 #if NETCOREAPP2_0
-        public async override Task AppendAllLinesAsync(string path, IEnumerable<string> contents, CancellationToken cancellationToken = default)
+        public async override Task AppendAllLinesAsync(string path, IEnumerable<string> contents, CancellationToken cancellationToken = default(CancellationToken))
         {
             mockFileDataAccessor.PathVerifier.IsLegalAbsoluteOrRelative(path, "path");
             VerifyValueIsNotNull(contents, "contents");
@@ -48,7 +48,7 @@ namespace System.IO.Abstractions.TestingHelpers
             await AppendAllLinesAsync(path, contents, MockFileData.DefaultEncoding, cancellationToken);
         }
 
-        public async override Task AppendAllLinesAsync(string path, IEnumerable<string> contents, Encoding encoding, CancellationToken cancellationToken = default)
+        public async override Task AppendAllLinesAsync(string path, IEnumerable<string> contents, Encoding encoding, CancellationToken cancellationToken = default(CancellationToken))
         {
             if (encoding == null)
             {
@@ -92,12 +92,12 @@ namespace System.IO.Abstractions.TestingHelpers
         }
 
 #if NETCOREAPP2_0
-        public async override Task AppendAllTextAsync(string path, string contents, CancellationToken cancellationToken = default)
+        public async override Task AppendAllTextAsync(string path, string contents, CancellationToken cancellationToken = default(CancellationToken))
         {
             AppendAllText(path, contents);
         }
 
-        public async override Task AppendAllTextAsync(string path, string contents, Encoding encoding, CancellationToken cancellationToken = default)
+        public async override Task AppendAllTextAsync(string path, string contents, Encoding encoding, CancellationToken cancellationToken = default(CancellationToken))
         {
             AppendAllText(path, contents, encoding);
         }
@@ -499,7 +499,7 @@ namespace System.IO.Abstractions.TestingHelpers
         }
 
 #if NETCOREAPP2_0
-        public async override Task<byte[]> ReadAllBytesAsync(string path, CancellationToken cancellationToken = default)
+        public async override Task<byte[]> ReadAllBytesAsync(string path, CancellationToken cancellationToken = default(CancellationToken))
         {
             mockFileDataAccessor.PathVerifier.IsLegalAbsoluteOrRelative(path, "path");
 
@@ -549,12 +549,12 @@ namespace System.IO.Abstractions.TestingHelpers
         }
 
 #if NETCOREAPP2_0
-        public async override Task<string[]> ReadAllLinesAsync(string path, CancellationToken cancellationToken = default)
+        public async override Task<string[]> ReadAllLinesAsync(string path, CancellationToken cancellationToken = default(CancellationToken))
         {
             return ReadAllLines(path);
         }
 
-        public async override Task<string[]> ReadAllLinesAsync(string path, Encoding encoding, CancellationToken cancellationToken = default)
+        public async override Task<string[]> ReadAllLinesAsync(string path, Encoding encoding, CancellationToken cancellationToken = default(CancellationToken))
         {
             return ReadAllLines(path, encoding);
         }
--- a/System.IO.Abstractions.TestingHelpers/MockFile.cs
+++ b/System.IO.Abstractions.TestingHelpers/MockFile.cs
@@ -40,25 +40,16 @@ namespace System.IO.Abstractions.TestingHelpers
         }
 
 #if NETCOREAPP2_0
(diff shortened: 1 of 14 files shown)


System.IO.Abstractions/IFile.cs
```diff
@@ -15,10 +20,22 @@ public interface IFile
         void AppendAllLines(string path, IEnumerable<string> contents);
         /// <inheritdoc cref="File.AppendAllLines(string,IEnumerable{string},Encoding)"/>
         void AppendAllLines(string path, IEnumerable<string> contents, Encoding encoding);
+#if NETCOREAPP2_0
+        /// <inheritdoc cref="File.AppendAllLinesAsync(string,IEnumerable{string},CancellationToken)"/>
+        Task AppendAllLinesAsync(string path, IEnumerable<string> contents, CancellationToken cancellationToken = default);
```
fgreinacher: CI is currently failing because the target-typed `default` literal is not available with the compiler we are using. Either change it to `default(CancellationToken)` or upgrade the language version in the csproj.

System.IO.Abstractions/System.IO.Abstractions.csproj
```diff
@@ -23,6 +23,12 @@
         <SignAssembly>True</SignAssembly>
     </PropertyGroup>
 
+    <ItemGroup Condition=" '$(TargetFramework)' == 'netcoreapp2.0'">
```
fgreinacher: Is this necessary? If yes you could just add the condition to the item group below.

fgreinacher:
Thanks for working on this! Glad to merge as soon as CI is happy!

System.IO.Abstractions.TestingHelpers/MockFile.cs
```diff
@@ -37,6 +39,28 @@ public override void AppendAllLines(string path, IEnumerable<string> contents, E
             AppendAllText(path, concatContents, encoding);
         }
 
+#if NETCOREAPP2_0
+        public async override Task AppendAllLinesAsync(string path, IEnumerable<string> contents, CancellationToken cancellationToken = default)
+        {
+            mockFileDataAccessor.PathVerifier.IsLegalAbsoluteOrRelative(path, "path");
+            VerifyValueIsNotNull(contents, "contents");
+
+            await AppendAllLinesAsync(path, contents, MockFileData.DefaultEncoding, cancellationToken);
```
robkeim: A couple of comments regarding the async implementation here:
1. Why are you awaiting this method instead of returning directly `return AppendAllLinesAsync(...);`?
2. For performance reasons any method that doesn't contain an `await` shouldn't be marked with the `async` keyword. Marking a method with the `async` keyword adds overhead and slows things down.

robkeim: This comment applies to the rest of the async methods in this PR as well, not just here.

System.IO.Abstractions.TestingHelpers/MockFile.cs
```diff
@@ -522,6 +584,18 @@ public override string ReadAllText(string path, Encoding encoding)
             return ReadAllTextInternal(path, encoding);
         }
 
+#if NETCOREAPP2_0
+        public async override Task<string> ReadAllTextAsync(string path, CancellationToken cancellationToken)
+        {
+            return ReadAllText(path);
```
robkeim: For these types of methods you can drop the `async` keyword and do a `return Task.FromResult(METHOD_CALL_HERE);` to save the overhead from the `async` keyword.

304NotModified:
This fixes https://github.com/System-IO-Abstractions/System.IO.Abstractions/issues/441, isn't?

robkeim:
@304NotModified correct

System.IO.Abstractions.TestingHelpers/MockFile.cs
fedoranimus: Good suggestion! I'll make that happen - how do you suggest handling the *Async methods which return an empty `Task` (as they pass through to a `void`)?

System.IO.Abstractions.TestingHelpers/MockFile.cs
fedoranimus: To answer your question, it was just oversight on my part. I'll clean those up.

System.IO.Abstractions/IFile.cs
fedoranimus: Updated to `default(CancellationToken)` and CI is fixed.

System.IO.Abstractions/System.IO.Abstractions.csproj
fedoranimus: Grouped into the other condition.

System.IO.Abstractions.TestingHelpers/MockFile.cs
fgreinacher: `Task.CompletedTask` is what you‘re for :)

fedoranimus:
Note: This PR doesn't have all of the wrappers implemented - working through that still.

fedoranimus:
I've implemented all the FileWrappers. Please review and let me know if you have any additional feedback. Thanks!

System.IO.Abstractions/FileWrapper.cs
```diff
@@ -21,6 +27,18 @@ public override void AppendAllLines(string path, IEnumerable<string> contents, E
             File.AppendAllLines(path, contents, encoding);
         }
 
+#if NETCOREAPP2_0
+        public override async Task AppendAllLinesAsync(string path, IEnumerable<string> contents, CancellationToken cancellationToken)
```
robkeim: Here you can remove the `async`/`await` and return a `Task.CompletedTask` as well (below too).

robkeim:
@fedoranimus looks like you missed removing the `async` keyword in one file... I went ahead and added a comment.

System.IO.Abstractions/FileWrapper.cs
fedoranimus: Gotcha, will update.

System.IO.Abstractions/FileWrapper.cs
```diff
@@ -214,6 +263,18 @@ public override string ReadAllText(string path, Encoding encoding)
             return File.ReadAllText(path, encoding);
         }
 
+#if NETCOREAPP2_0
+        public override async Task<string> ReadAllTextAsync(string path, CancellationToken cancellationToken)
+        {
+            return await File.ReadAllTextAsync(path, cancellationToken);
```
fedoranimus: @robkeim Should I leave the Read methods as is with `async/await`, or is there a difficult way you'd prefer for me to handle them?

robkeim: In this case you can remove the `async` and `await` keywords and it will compile since you're returning an overload of `ReadAllTextAsync` which also returns the same type `Task<string>`

System.IO.Abstractions/FileWrapper.cs
robkeim: Thanks!

System.IO.Abstractions/FileWrapper.cs
fedoranimus: Done.

fgreinacher:
Great, thanks also from my side! One last thing: Please update the base version in https://github.com/System-IO-Abstractions/System.IO.Abstractions/blob/master/version.json to `6.0` as we try to follow SemVer and this is a breaking change for implementers.

fedoranimus:
@fgreinacher Done.

christianrondeau:
I noticed ReadAllLinesAsync is not available with .NET Core 3 preview 7, looks like the precompiler condition trips up here. I wanted to ask here before opening an issue, maybe there's something I missed?

fgreinacher:
Is this a new API of .NET Core 3? If yes, thats fine because we don’t target .NET Core 3 at the moment. If no, it’s a bug. In both cases feel free to open an issue for that. Thanks!

christianrondeau:
No it's not, I believe the problem is that there's a precompiler #if specifically for netstandard 2.0

Issue #491, opened 2019-07-08
Cannot catch exception thrown by File.WriteAllBytesAsync


JerrettDavis:
When attempting to use `IFile.WriteAllBytesAsync`, all underlying exceptions from `File.WriteAllBytesAsync` are quietly disregarded. I can see the exceptions being thrown, but a `try/catch` fails to catch the exceptions. Instead, the program continues execution on the next line. I believe this is due to the `FileWrapper` not awaiting the response from `File.WriteAllBytes`, and instead it's always returning a Task.Completed task.

robkeim:
@JerrettDavis in order to catch exceptions thrown by async methods you need to await their results like this:
```
Task<Result> resultTask = someMethodThatReturnsAResultByThrowsAnException();
// No exception is thrown at this point
// Continue doing other work...
try
{
    Result result = await resultTask; // this will now throw an exception
}
catch (Exception e)
{
    // Handle your exception here
}
```

Here you can see a quick fiddle that I threw together showing this behavior:
https://dotnetfiddle.net/7jj6lJ

JerrettDavis:
@robkeim Ordinarily, that would work, but it doesn't appear to work properly for `WriteAllBytesAsync`.

Here is an example method:

```csharp
public async Task<string> SaveFileAsync(string path, byte[] file, string fileType, CancellationToken cancellationToken = default)
{
    var filename = _filenameProvider.GetFilename(path, fileType);
    var fullPath = _fileSystem.Path.Combine(path, filename)?.Replace('/', '\\');

    _fileSystem.Directory.CreateDirectory(path);
    try
    {
        await _fileSystem.File.WriteAllBytesAsync(fullPath, file, cancellationToken);
    }
    catch (Exception ex)
    {
        Console.WriteLine(ex);
    }

    return filename;
}
```

The `catch` block is never triggered, but

fgreinacher:
This is fixed with https://github.com/System-IO-Abstractions/System.IO.Abstractions/releases/tag/v6.0.15, should be on NuGet.org soon

JerrettDavis:
It's live on nuget, and the fix is confirmed working in my original use-case.

PR #492, opened 2019-07-08
Updated FileWrapper async calls to allow errors to bubble up.

JerrettDavis:
Await all asynchronous `FileWrapper` calls. Unawaited calls did not bubble up errors, per issue #491 .

### DIFF
--- a/System.IO.Abstractions/FileWrapper.cs
+++ b/System.IO.Abstractions/FileWrapper.cs
@@ -28,14 +28,14 @@ namespace System.IO.Abstractions
         }
 
 #if NETCOREAPP2_0
-        public override async Task AppendAllLinesAsync(string path, IEnumerable<string> contents, CancellationToken cancellationToken)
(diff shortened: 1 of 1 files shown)


robkeim:
@JerrettDavis I didn't see your PR until I pushed mine #493 :(

I'd prefer to avoid adding the async keyword to all of those methods because that has a performance impact. Instead we can return the underlying `Task` directly and let upstream callers decide how and when they want to handle it.

JerrettDavis:
No worries, I was mostly trying to get a quick proof of concept working just to verify I wasn't going crazy. Whatever works!

robkeim:
You're definitely not going crazy... thanks for taking the time to report this issue!

JerrettDavis:
I went ahead and corrected my PR, just in case.

robkeim:
Let's keep this PR, I've gone ahead and closed mine.

fgreinacher:
Thanks!

PR #493, opened 2019-07-08
Async methods return underlying Task instead of Task.CompletedTask

### LINKED ISSUE #491: Cannot catch exception thrown by File.WriteAllBytesAsync

robkeim:
Fixes #491


robkeim:
Duplicate of #492

Issue #499, opened 2019-08-14
Support File.ReadAllLinesAsync and friends on .NET Core 3 / netstandard2.1


christianrondeau:
Async methods were implemented in https://github.com/System-IO-Abstractions/System.IO.Abstractions/pull/481#issuecomment-521274716 but I think the precompiler directives or somehow the nuget binding doesn't get them.

Sample csproj:

```
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <TargetFramework>netstandard2.1</TargetFramework>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="System.IO.Abstractions" Version="6.0.21" />
  </ItemGroup>
</Project>
```

Sample code:

```
var lines = await new FileSystem().File.ReadAllLinesAsync(@"C:\file.txt")
```

I think the problem is simply that we're limited to NETCOREAPP2_0: https://github.com/fedoranimus/System.IO.Abstractions/blob/master/System.IO.Abstractions/FileBase.cs#L322 - you can probably instead use `NETSTANDARD2_0` but I won't be able to test that right away. I won't be able to provide a PR as I'm off until the end of next week, but if no one took this issue when I'm back all provide one.

fgreinacher:
Your sample csproj is targeting .NET Standard 2.1 and will therefore pull in the .NET Standard 2.0 version of System.IO.Abstractions.

Back in #481 we added support for the async overloads to a newly added .NET Core 2.0 target. To solve your issue we need to add an additional target for .NET Standard 2.1 in which we could light up the async overloads.

PR #502, opened 2019-08-28
Support File.ReadAllLinesAsync and friends on .NET Core 3 / netstandard2.1

christianrondeau:
Solves #499, but to do that I added a new netstandard2.1 target. As an example, if you look here, you'll see these async methods were there since netcoreapp2.0 but only available starting with netstandard2.1: https://docs.microsoft.com/en-us/dotnet/api/system.io.file.readallbytesasync?view=netstandard-2.1

I'm not fully sure if this will magically work for the nuget releases, nor if the build server can actually build it (it needs the now-stable preview 8 of .net core 3), but if you want you can merge this after the official .net core 3 release!

### DIFF
--- a/System.IO.Abstractions.TestingHelpers.Tests/FileSystemTests.cs
+++ b/System.IO.Abstractions.TestingHelpers.Tests/FileSystemTests.cs
@@ -20,7 +20,7 @@ namespace System.IO.Abstractions.TestingHelpers.Tests
         }
 #endif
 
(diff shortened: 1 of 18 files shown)


fgreinacher:
The Ubuntu image seems to have the correct .NET Core SDK now. Windows images don't seem to work yet. Let's just hang on :)

fgreinacher:
Thanks a lot for your contribution!

Issue #439, opened 2019-01-09
MockFileSystem.Directory.EnumerateFiles does not return files with no extension when Search Pattern is "*."


ptgullas:
When using a MockFileSystem, Directory.EnumerateFiles doesn't return files that don't have an extension, and the search pattern is "*." . It works with a real file system.

The below test fails:
``` c#
        [TestMethod]
        public void MockFileSystem_EnumerateFiles_ShouldReturnFilesWithNoExtensions() {
            var fs = new MockFileSystem();
            var mf = new MockFileData("my name is Luka");
            string filePath = @"c:\downloadfolder\ThisIsAFileWithNoExtension";
            fs.AddFile(filePath, mf);
            IEnumerable<string> files = fs.Directory.EnumerateFiles(@"c:\downloadfolder", "*.", SearchOption.TopDirectoryOnly);
            Assert.AreEqual(1, files.Count());
        }
```

fgreinacher:
Thanks for reporting!

fgreinacher:
Fixed with #516

PR #516, opened 2019-10-06
fix(439): Pattern "*." returns only files with no extension

KaliVi:
Add special case to Directory.GetFilesInternal to return only files without extensions when search pattern is "*."
This fixes bug #439

### DIFF
--- a/System.IO.Abstractions.TestingHelpers/MockDirectory.cs
+++ b/System.IO.Abstractions.TestingHelpers/MockDirectory.cs
@@ -237,6 +237,10 @@ namespace System.IO.Abstractions.TestingHelpers
             {
                 fileNamePattern = isUnix ? @"[^/]*?/?" : @"[^\\]*?\\?";
             }
+            else if (searchPattern == "*.")
+            {
+                fileNamePattern = isUnix ? @"[^/.]*?/?" : @"[^\\.]*?\\?";
+            }
             else
             {
                 fileNamePattern = Regex.Escape(searchPattern)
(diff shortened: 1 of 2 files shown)


System.IO.Abstractions.TestingHelpers/MockDirectory.cs
```diff
@@ -237,6 +237,10 @@ public override string[] GetFiles(string path, string searchPattern, SearchOptio
             {
                 fileNamePattern = isUnix ? @"[^/]*?/?" : @"[^\\]*?\\?";
             }
+            else if (searchPattern == "*.")
```
rpendleton: By checking for `*.`, `MockDirectory` still behaves differently than `System.Directory`. The system version interprets search filters ending in a period as a search for files that either lack an extension or that literally end with a period. Because of this, there are quite a few search filters that won't be fixed by this PR.

For example, `hello*.` should match any filename beginning with `hello` that doesn't have an extension (or ends with just a period). `h????*.` would be similar but would require the filename to be at least a 5 characters long. An even more complex example could be something like `h?*-w*.` which would match a filename like 'hello-world', 'hello-world.' or `hi there-world`.

KaliVi: @rpendleton
Thanks for reviewing!
I will think about the examples you provided.
Thanks!

KaliVi: Hi, I checked in a new version, which is much closer to how System.Directory behaves.

PR #521, opened 2019-10-28
Fix MockFileSystem serialization with empty PathVerifier

darl2ng:
test unit using fileSystem.GetFile(path).Contents) still work but not fileSystem.File.ReadAllBytes(path)) which uses PathVerifier.

### DIFF
--- a/System.IO.Abstractions.TestingHelpers.Tests/MockFileSystemSerializationTests.cs
+++ b/System.IO.Abstractions.TestingHelpers.Tests/MockFileSystemSerializationTests.cs
@@ -1,0 +1,95 @@
+namespace System.IO.Abstractions.TestingHelpers.Tests
+{
+    using NUnit.Framework;
+    using System.Reflection;
+    using System.Runtime.Serialization;
+    using System.Runtime.Serialization.Formatters.Binary;
+    using Text;
+    using XFS = MockUnixSupport;
+    [TestFixture]
+    class MockFileSystemSerializationTests
+    {
+
+        const string FS_FILENAME = @"test.fs";
+
+        [Test]
+        public void SerializationTexts()
+        {
+            // Arrange
+            string path = XFS.Path(@"c:\something\demo.txt");
+
+            var content = "Hello there!" + Environment.NewLine + "Second line!" + Environment.NewLine;
+
+            var fileSystem = new MockFileSystem();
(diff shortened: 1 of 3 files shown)


System.IO.Abstractions.TestingHelpers.Tests/MockFileSystemSerializationTests.cs
```diff
@@ -0,0 +1,94 @@
...
+
+            // Assert
+            Assert.AreEqual(
+                expected,
+                fileSystem.GetFile(path).Contents);
+            Assert.AreEqual(
+                content,
+                fileSystem.File.ReadAllBytes(path));
+        }
+        private void SaveFileSystem(MockFileSystem fileSystem)
```
fgreinacher: I'd suggest to serialize the file system to a `MemoryStream` and returning the stream from this method. This way we wouldn't need to touch the actual file system.

System.IO.Abstractions.TestingHelpers.Tests/MockFileSystemSerializationTests.cs
```diff
@@ -0,0 +1,94 @@
...
+        {
+            IFormatter formatter = new BinaryFormatter();
+            Stream stream = new FileStream(Path.GetTempPath() + FS_FILENAME, FileMode.Create, FileAccess.Write);
+            formatter.Serialize(stream, fileSystem);
+            stream.Close();
+        }
+        private IFileSystem LoadFileSystem()
```
fgreinacher: Based on the suggestion above this method would need to take the memory stream as parameter. It can then deserialize from that stream instead of touching the actual file system.

System.IO.Abstractions.TestingHelpers.Tests/MockFileSystemSerializationTests.cs
```diff
@@ -0,0 +1,94 @@
...
+        {
+            IFormatter formatter = new BinaryFormatter();
+            Stream stream = new FileStream(Path.GetTempPath() + FS_FILENAME, FileMode.Open, FileAccess.Read);
+            IFileSystem fileSystem = (MockFileSystem)formatter.Deserialize(stream);
+            stream.Close();
+            return fileSystem;
+        }
+        private void ClearFileSystem()
```
fgreinacher: This should not be necessary after the suggestions above are done.

Issue #334, opened 2018-07-18
Multi-target-framework support


rkoeninger:
There are a lot of `#if NET40 ... #endif` found in the code. These surround IO functions that are available in .NET Framework >= 4.0, but were not available in .NET Standard 1.0 < 2.0.

Some of these are now available in .NET Standard 2.0, but are not supported by the library, so the directive could be changed to `#if ! netstandard1.4` or whatever the exact naming would be.

Framework-specific methods could be made adjacent, or event moved to their own file using partial classes. For example, remove the 4 methods in `#if`'s from `MockFileInfo.cs` and add a file called `MockFileInfo.net40.cs` containing something like this:

```csharp
#if NET40

namespace System.IO.Abstractions.TestingHelpers
{
    public partial class MockFileInfo
    {
        public override void Decrypt() { ... }
        public override void Encrypt() { ... }
        public override FileInfoBase Replace(string x, string y) { ... }
        public override FileInfoBase Replace(string x, string y, bool z) { ... }
    }
}

#endif
```

Not sure what the name would be if it was for net40 and netstandard2.0, or not for netstandard1.4, however you represent it. Wouldn't want to have 2 separate files or blocks with the same thing in it.

Need to do a survey of methods and see what's supported in netstandard2.0 compared to netstandard1.4 and net40. Might be some in netstandard2.0 but not net40. [look at the diff files here](https://github.com/dotnet/standard/tree/master/docs/versions).

---

Don't know if this is an issue, but the libraries are build for `net40;netstandard1.4;netstandard2.0;`, but the `TestHelpers.Tests` project targets `net40;netcoreapp2.0`, leaving out Core 1.4. [Brief reading](https://stackoverflow.com/questions/45867239/advantages-of-netcoreapp2-0-vs-netstandard2-0-for-a-library-project) makes it sound like all 3 should target standard instead of core, since all 3 are libraries?

fgreinacher:
I don't like moving TFM-specific stuff to separate files because that quickly gets out-of-control, as you already noticed for the `!NETSTANDARD1.4` case.

What do you think of introducing "feature defines" in the project file and then using those in the sources?

It would look like something like this in the project file:
```xml
<PropertyGroup>
   <DefineConstants Condition="'$(TargetFramework)' != 'netstandard1.4'">$(DefineConstants);FEATURE_FILE_ENCRYPTION</DefineConstants>
</PropertyGroup>
```

And in the sources:
```csharp
[...]

PR #533, opened 2019-12-20
Enable File.Replace for all but netstandard1.4

spanglerco:
According to [the documentation](https://docs.microsoft.com/en-us/dotnet/api/system.io.file.replace?f1url=https%3A%2F%2Fmsdn.microsoft.com%2Fquery%2Fdev16.query%3FappId%3DDev16IDEF1%26l%3DEN-US%26k%3Dk%2520System.IO.File.Replace%2520%253Bk%2520DevLang-csharp%2520%26rd%3Dtrue%230877433039133023292&view=netcore-2.0#applies-to), File.Replace is a part of .NET Standard 2.0 and supported by Core 2.0 and later.

Related to #334 but does not attempt to address all methods that might be supported by .NET Standard/Core 2.0 and later.

This also enables the File.Replace unit tests in the same way. The tests pass on Windows 10 and Ubuntu on Windows.

### DIFF
--- a/System.IO.Abstractions.TestingHelpers.Tests/MockFileTests.cs
+++ b/System.IO.Abstractions.TestingHelpers.Tests/MockFileTests.cs
@@ -610,7 +610,6 @@ namespace System.IO.Abstractions.TestingHelpers.Tests
         }
 #endif
 
-#if NET40
         [Test]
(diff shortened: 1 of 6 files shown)


fgreinacher:
Thanks a lot!

PR #549, opened 2020-01-28
Add validation for illegal characters when creating MockDirectory

### LINKED ISSUE #546: CreateDirectory doesn't throw exception when the path contains illegal characters
When not using mocking, creating a directory with an illegal character (eg. '?') throws an exception.
With mocking, it simply succeeds.

Used method:
MockFileSystem.Directory.CreateDirectory("c:\\temp\\Invalid_?_");

BlueberryKing:
This fixes #546 

MockFileSystem did not check for illegal characters such as '_' and '?' when creating a directory and instead `CreateDirectory` just returned `null`.

We now check for illegal characters with `PathVerifier.HasIllegalCharacters()` and throw `CommonExceptions.IllegalCharactersInPath` accordingly.


fgreinacher:
Thanks a lot for your contribution!

PR #558, opened 2020-02-17
Reset MockFileStream position after flushing

### LINKED ISSUE #519: MockFileStream unexpected behavior when used via BinaryWriter
*Using the latest stable NuGet version (7.0.7)*

If I initialize a BinaryWriter with a MockFileStream (obtained e.g. by using MockFileSystem.File.Create), and then manipulate it via BinaryWriter.BaseStream, I get unexpected behavior. For instance, examine the following: 
```C#
using (var s = fileSystem.File.Create(@"C:\testfile.bin"))
using (var writer = new System.IO.BinaryWriter(s))
{
    writer.Write(new byte[400], 0, 400);
    writer.BaseStream.Seek(200, System.IO.SeekOrigin.Begin);
    Console.WriteLine(writer.BaseStream.Position);   // prints 400, expect 200
    Console.WriteLine(s.Position); // prints 400, expect 200
}
```
Presumably this is because BinaryWriter.BaseStream calls Flush on the underlying stream before returning it, though it's not immediately clear to me why this should be a problem. Avoiding accessing `BaseStream` by calling e.g. `writer.Seek` directly *does* avoid the problem, but the point is that MockFileStream does not replicate the behavior of a real FileStream in this regard:

```C#
using (var s = System.IO.File.Create(@"C:\testfile.bin")
using (var writer = new System.IO.BinaryWriter(s))
{
    writer.Write(new byte[400], 0, 400);
    writer.BaseStream.Seek(200,  System.IO.SeekOrigin.Begin);
    Console.WriteLine(writer.BaseStream.Position);   // prints 200 as expected
    Console.WriteLine(s.Position); // prints 200 as expected
}
```

hangy:
Resets the internal Stream position to the original position after reading the file's data to the `data` array.

Fixes #519

### DIFF
--- a/System.IO.Abstractions.TestingHelpers.Tests/MockFileStreamTests.cs
+++ b/System.IO.Abstractions.TestingHelpers.Tests/MockFileStreamTests.cs
@@ -164,5 +164,25 @@
             // Assert
             Assert.Throws<ObjectDisposedException>(() => stream.WriteByte(0));
         }
+
+        [Test]
+        public void MockFileStream_Flush_ShouldNotChangePosition()
+        {
+            // Arrange
+            var fileSystem = new MockFileSystem();
+            var path = XFS.Path("C:\\test");
+            fileSystem.AddFile(path, new MockFileData(new byte[0]));
+
+            using (var stream = fileSystem.FileInfo.FromFileName(path).OpenWrite())
+            {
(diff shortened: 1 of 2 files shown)


fgreinacher:
Thanks a lot!

PR #559, opened 2020-02-17
Copy metadata when a MockFile is moved

### LINKED ISSUE #557: MockFile.Move does not preserve file metadata when moving a file
MockFile.Move does not preserve file metadata when moving a file — instead, it simply creates a file with the same content in a new location, discarding file metadata such as created/modified timestamp.

**Expected:** A moved file's modification date is preserved
**Actual:** A moved file's modification date is always set to `2010-01-04 00:00:00 +0400`

### Repro

```csharp
using System;
using Xunit;

namespace Bug
{
	public class Repro
	{
		[Fact]
		public void WhenMovingFile_ShouldCopyMetadata()
		{
			// Arrange
			MockFileSystem sut = new MockFileSystem();

			DateTime writeTime = new DateTime(2020, 01, 01);
			sut.AddFile("file1.txt", new MockFileData("Test file contents") { LastWriteTime = writeTime });

			// Act
			sut.File.Move("file1.txt", "file2.txt");

			// Assert
			var destInfo = sut.FileInfo.FromFileName("file2.txt");
			Assert.Equal(writeTime, destInfo.LastWriteTime);
		}
	}
}
```

##### Output
```
Bug.Repro.WhenMovingFile_ShouldCopyMetadata

Assert.Equal() Failure
Expected: 2020-01-01T00:00:00.0000000
Actual:   2010-01-04T00:00:00.0000000
   at Bug.Repro.WhenMovingFile_ShouldCopyMetadata() in C:\Dev\MockFileSystemBug\Repro.cs:line 23
```

hangy:
Instead of copying only the data array, use the constructor that clones the `MockFile` properties, too.

Fixes #557

### DIFF
--- a/System.IO.Abstractions.TestingHelpers.Tests/MockFileMoveTests.cs
+++ b/System.IO.Abstractions.TestingHelpers.Tests/MockFileMoveTests.cs
@@ -362,5 +362,24 @@ namespace System.IO.Abstractions.TestingHelpers.Tests
 
             Assert.Throws<FileNotFoundException>(action);
         }
+
+        [Test]
+        public void MockFile_Move_ShouldRetainMetadata()
+        {
+            string sourceFilePath = XFS.Path(@"c:\something\demo.txt");
+            string sourceFileContent = "this is some content";
+            DateTimeOffset creationTime = DateTimeOffset.Now;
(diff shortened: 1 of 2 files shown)


fgreinacher:
Thanks a ton, also for this one 👍

PR #566, opened 2020-03-20
Support the FileStream constructor with FileOptions

vladimirlapacek:
Support the FileStream constructor with FileOptions parameter, which is present in implementations, but not in the IFileStreamFactory.


fgreinacher:
Cool, thanks! Could you please bump the major in https://github.com/System-IO-Abstractions/System.IO.Abstractions/blob/master/version.json#L3 to indicate this breaking change.

fgreinacher:
Thanks 👍

PR #569, opened 2020-04-01
Fixes #568: MockFile.Copy does not honor AccessFileShare on MockFileData

### LINKED ISSUE #568: MockFile.Copy does not honor AllowFileShare on MockFileData
If MockFileData has `AllowFileShare = Fileshare.None` various other operators (Move, OpenRead, etc) properly check and throw is FileShare is not set appropriately.  Copy should also honor this, and throw if FileShare does not have Read access.

pianomanjh:
- add `CheckFileAccess` on `MockFile.Copy`, so `IOException` is correctly thrown if Read access is not granted.
- added missing coverage on MockFile.Move related to `AccessFileShare`


fgreinacher:
Great, thank you also for this!

PR #570, opened 2020-04-01
Fixes #567: MockFileInfo.Append should create a new file if missing

### LINKED ISSUE #567: MockFileInfo.AppendText does not create file
When using MockFileInfo.AppendText to write text to a non-existing file, the method throws a FileNotFound exception. This works differently than the real FileInfo.AppendText method, which creates the file if it doesn't already exist. This is specified in .NET Core/Framework documentation: https://docs.microsoft.com/en-us/dotnet/api/system.io.fileinfo.appendtext?view=netcore-3.1

The line causing the issue is here: https://github.com/System-IO-Abstractions/System.IO.Abstractions/blob/28e5d7d106e561fc9056f4e670a96cc8c3aebcad/System.IO.Abstractions.TestingHelpers/MockFileInfo.cs#L155

Sample code:
```
public class Test
{
    public void AppendText_Test()
    {
        var file = fs.FileInfo.FromFileName("./Path/To/File.txt");

        using var sw = file.AppendText(); // This should create the file instead of throwing an exception
        sw.WriteLine("Text");
    }
}
```

pianomanjh:


fgreinacher:
👍 🚀

PR #571, opened 2020-04-01
Fixes #532: throw IOException: File already exists, if the file already exists when creating a FileSystem with FileMode.CreateNew

### LINKED ISSUE #532: FileStream.Create with FileMode.CreateNew doesn't throw exception when file already exists
When not using mocking, creating a filestream using FileMode.CreateNew throws an IOException when the target file already exists. However, the MockFileStream constructor behaves differently. When passing CreateNew as FileMode, it truncates the file. This causes issues for us in our unit tests, since the test scenario behaves differently than then real code.

pianomanjh:
- additional unreported fix: FileStream created with FileMode.Open or Truncate should throw an exception if the file is missing
- added a couple missing cases


fgreinacher:
Thanks a ton @pianomanjh!

PR #573, opened 2020-04-01
fixes #551, MockDirectory.Move should throw when target cannot be parsed

### LINKED ISSUE #551: MockDirectory.Move does not throw exception if (part of the) target's parent folder does not exist
MockDirectory.Move should throw a DirectoryNotFoundException if (part of) the parent folder of the target folder does not exist.

If you move from "c:\temp\exists\foldertomove" to "c:\temp\doesnotexist\foldertomove" and "c:\temp\doesnotexist" does not exist, an exception should be thrown.

In the code below, method MockDirectory_Move_ShouldMoveDirectories does not throw an exception on the Move, where Directory_Move_ShouldMoveDirectories does.

And what's even worse: the first assert in MockDirectory_Move_ShouldMoveDirectories succeeds, while the second assert fails. So according to the MockDirectory, the folder "c:\temp\doesnotexist" does not exist, but "c:\temp\doesnotexist\foldertomove" _does_ exist!?

```csharp
[TestClass]
public class MockDirectoryTests
{
    [TestMethod]
    public void MockDirectory_Move_ShouldMoveDirectories()
    {
        // Arange
        {
            {@"c:\temp\exists\foldertomove", new MockDirectoryData()}
        });

        string folderToMove = @"c:\temp\exists\foldertomove";
        string targetDirName = @"c:\temp\doesnotexist\foldertomove";

        // Act
        fileSystem.Directory.Move(folderToMove, targetDirName);

        // Assert
        Assert.IsTrue(fileSystem.Directory.Exists(@"c:\temp\doesnotexist\foldertomove"));
        Assert.IsTrue(fileSystem.Directory.Exists(@"c:\temp\doesnotexist"));
    }

    [TestMethod]
    public void Directory_Move_ShouldMoveDirectories()
    {
        // Arange
        Directory.CreateDirectory(@"c:\temp\exists\foldertomove");

        string folderToMove = @"c:\temp\exists\foldertomove";
        string targetDirName = @"c:\temp\doesnotexist\foldertomove";

        // Act
        // Move will throw a DirectoryNotFoundException: Could not find a part of the path
        Directory.Move(folderToMove, targetDirName);

        // Assert
        Assert.IsTrue(Directory.Exists(@"c:\temp\doesnotexist\foldertomove"));
    }
}
```

pianomanjh:
Based on .NET source, the Directory.Move method PInvokes Win32 MoveFileExW (tested on net 472), which has certain rules for what path shapes it can tolerate.  I wrote a comparable C# method that matched the cases I could think to test, confirming the behavior against `System.IO` behavior.


System.IO.Abstractions.TestingHelpers/Win32FileSystemBehavior.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace System.IO.Abstractions.TestingHelpers
+{
+    public static class Win32FileSystemBehavior
+    {
+        public static bool MoveFileCanParsePath(string fullPath)
```
fgreinacher, line 5:
```suggestion
        internal static bool MoveFileCanParsePath(string fullPath)
```

System.IO.Abstractions.TestingHelpers/MockDirectory.cs
```diff
@@ -415,6 +415,11 @@ public override void Move(string sourceDirName, string destDirName)
                 throw new DirectoryNotFoundException($"Could not find a part of the path '{sourceDirName}'.");
             }
 
+            if (!Win32FileSystemBehavior.MoveFileCanParsePath(fullDestPath))
```
fgreinacher, line 418: Did you compare behavior on Windows with behavior on Unix? Do we want to have the same behavior for both?

pianomanjh: I was thinking about this.  No I haven't manually tested against Unix, though I'm happy to before merging.
Really it seems the best approach for these kind of situations where we're attempting to exactly match the behavior of not just the OS, but the runtime, we could think about a test suite that runs these tests side by side with actual `FileSystem` to ensure the behavior is accurate on the runtime/OS combination.
This does mean introducing I/O into the unit test, but they should still run very fast.
Thoughts?

fgreinacher: IMO checking it manually is fine. We do not aim for 100% perfect parity with the real filesystem implementation, so I don't think we need to add such tests within this PR.

pianomanjh: I just manually ran the tests (`dotnet test`) against the real filesystem on both Windows, and Ubuntu (Windows Subsystem for Linux), and they pass.  (Except for the net461 tests on Linux, as expected, as it's unsupported there)
So we are good to go.

pianomanjh: also added a few more cases to catch some of the stranger pathing possible

System.IO.Abstractions.TestingHelpers/Win32FileSystemBehavior.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace System.IO.Abstractions.TestingHelpers
+{
+    public static class Win32FileSystemBehavior
```
fgreinacher, line 3:
```suggestion
    internal static class Win32FileSystemBehavior
```

System.IO.Abstractions.TestingHelpers/Win32FileSystemBehavior.cs
```diff
@@ -0,0 +1,17 @@
...
+    {
+        internal static bool MoveFileCanParsePath(string fullPath)
```
fgreinacher, line 5:
```suggestion
        public static bool MoveFileCanParsePath(string fullPath)
```

fgreinacher:
Thanks for all your contributions!

IvoMiller:
This appears to cause a bug. `MockFileSystem.Directory.Move` no longer works with paths with more than one element.

Reproduction:
`            var mockFileSystem = new MockFileSystem();
            mockFileSystem.AddDirectory(@"C:\Foo\Bar");

            mockFileSystem.Directory.Move(@"C:\Foo\Bar", @"C:\Foo\Baz");`

This produces:
System.IO.DirectoryNotFoundException : Could not find a part of the path.
   at System.IO.Abstractions.TestingHelpers.MockDirectory.Move(String sourceDirName, String destDirName)

fgreinacher:
> This appears to cause a bug. `MockFileSystem.Directory.Move` no longer works with paths with more than one element.
> 
> Reproduction:
> 
> ```
> var mockFileSystem = new MockFileSystem();
> mockFileSystem.AddDirectory(@"C:\Foo\Bar");
> 
> mockFileSystem.Directory.Move(@"C:\Foo\Bar", @"C:\Foo\Baz");
> ```
> 
> This produces:
> System.IO.DirectoryNotFoundException : Could not find a part of the path.
> at System.IO.Abstractions.TestingHelpers.MockDirectory.Move(String sourceDirName, String destDirName)

Sorry for the trouble. I'll revert this PR to get out a fixed version asap.

@pianomanjh Could you take a look at the problem and re-apply this PR once you fixed it?

fgreinacher:
@IvoMiller This change is reverted in https://www.nuget.org/packages/System.IO.Abstractions/10.0.8

IvoMiller:
@fgreinacher Thank you very much. I can confirm this solved the issue.

pianomanjh:
Thank you @IvoMiller, with my apologies.  I'll take a look tonight at what I've missed there, and be sure to get test cases in accordingly.

pianomanjh:
Hi there @IvoMiller @fgreinacher , I've had a chance to look at this, and the newly implemented behavior is correct.  The behavior change, in response to the bug, was to match MockFileSystem behavior to actual Filesystem behavior.  
Your repro above:
```
var mockFileSystem = new MockFileSystem();
mockFileSystem.AddDirectory(@"C:\Foo\Bar");

mockFileSystem.Directory.Move(@"C:\Foo\Bar", @"C:\Foo\Baz");
```
fails on a real file system with a `DirectoryNotFoundException`.

Let me know if you agree this PR should be re-merged.

fgreinacher:
I don’t get why this code should fail. This move operation is merely a rename that in my opinion is totally valid.

pianomanjh:
I couldn't agree more, but I doubt you'll get the Windows team to fix their kernel. :)  If the expectation is that the tests are accurate with respect to the OS behavior, then the change is warranted.  Feel free to double-check me against the file system you are using.  (I've only tested on Windows 10 19592, and Ubuntu 18.04.)

pianomanjh:
my apologies @fgreinacher I did not make an additional commit with the major version bump.  Do we want to do that?

fgreinacher:
I just opened #579 to reproduce the behavior.

PR #576, opened 2020-04-13
Fixed #464

### LINKED ISSUE #464: MockFile.ReadAllLines does not deal with BOM correctly
Or, to be more precise, not at all. To deal with file data, a `StreamReader` must be used. This is also what .NET internally does.

Consider the following example:

```
fs.AddDirectory("U:\\");
fs.File.WriteAllLines("U:\\Test.txt", new[] { "Hallo Welt"}, Encoding.UTF8);
var chars = fs.File.ReadAllLines("U:\\Test.txt", Encoding.UTF8).Single().ToArray();
```

`chars[0]` is now `65279`, which is the BOM. That’s because the encoding is used directly:

https://github.com/System-IO-Abstractions/System.IO.Abstractions/blob/452e3312776c14843806b14224ba82ae5d5ec137/System.IO.Abstractions.TestingHelpers/MockFile.cs#L496-L498

I’ll create a pull request later, if that’s okay.

---

Also, while I’m at it, might I ask why this library even bothers with trying to deal with text as a special case? This introduces unnecessary complexity and potential for differences in behavior with actual System.IO.

PeterDowdy:
Fixed https://github.com/System-IO-Abstractions/System.IO.Abstractions/issues/464 by using the test case provided. The change is small: StreamReader honours BOMs, but to preserve big-endian Unicode support you must pass an encoding in that case. Since encoding is always provided into that method, I pass it always.

### DIFF
--- a/System.IO.Abstractions.TestingHelpers/MockFile.cs
+++ b/System.IO.Abstractions.TestingHelpers/MockFile.cs
@@ -490,7 +490,9 @@ namespace System.IO.Abstractions.TestingHelpers
 
             using (var ms = new MemoryStream(mockFileDataAccessor.GetFile(path).Contents))
             using (var sr = new StreamReader(ms, encoding))
+           {
                 return sr.ReadToEnd().SplitLines();
+           }
         }
 
         public override string ReadAllText(string path)
--- a/System.IO.Abstractions.TestingHelpers/MockFile.cs
+++ b/System.IO.Abstractions.TestingHelpers/MockFile.cs
@@ -487,9 +487,10 @@ namespace System.IO.Abstractions.TestingHelpers
             }
 
             mockFileDataAccessor.GetFile(path).CheckFileAccess(path, FileAccess.Read);
-            return encoding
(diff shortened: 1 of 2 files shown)


System.IO.Abstractions.TestingHelpers/MockFile.cs
```diff
@@ -487,9 +487,10 @@ public override string[] ReadAllLines(string path, Encoding encoding)
             }
 
             mockFileDataAccessor.GetFile(path).CheckFileAccess(path, FileAccess.Read);
-            return encoding
-                .GetString(mockFileDataAccessor.GetFile(path).Contents)
-                .SplitLines();
+
+            using (var ms = new MemoryStream(mockFileDataAccessor.GetFile(path).Contents))
+            using (var sr = new StreamReader(ms, encoding))
+                return sr.ReadToEnd().SplitLines();
```
fgreinacher, line 493:
```suggestion
           {
               return sr.ReadToEnd().SplitLines();
           }
```

fgreinacher:
Nice, simple fix. Thank you!

PR #579, opened 2020-04-17
Fix/dir move again

fgreinacher:


System.IO.Abstractions.TestingHelpers.Tests/MockDirectoryTests.cs
```diff
@@ -1371,6 +1371,31 @@ public void Move_DirectoryExistsWithDifferentCase_DirectorySuccessfullyMoved()
...
+
+            var fileSystem = fileSystemType == FileSystemType.RealFileSystem
+                ? (IFileSystem)new FileSystem()
+                : new MockFileSystem();
+
+            fileSystem.Directory.CreateDirectory(sourceDir);
+
+            Assert.DoesNotThrow(() =>
+                fileSystem.Directory.Move(sourceDir, targetDir)
+            );
```
fgreinacher, line 1396: @pianomanjh This assertion fails when using the mock FS and passes when using the real FS on both Windows & Linux (see the PR build).

Did you test differently?

pianomanjh: I'm taking another look now to see what's up, I've added a matrix of source/target paths now to cover more combinations...

pianomanjh: yeah, I can confirm the regression.  I can add some commits here with the fix and much more coverage around the characteristics of the source folder.  It's quite tricky sadly.  Copying from `c:\foo\bar` to `c:\foo\baz` does work, but from `c:\foo\bar` to `c:\foo2\baz` does not.

pianomanjh: there is also different behavior based on whether or not your user principal created 'all' the folders in the source path, which is how my testing was flawed.  I won't address that situation with this work, and make the assumption that all folders are permissioned.

pianomanjh: I posted the fix here: https://github.com/System-IO-Abstractions/System.IO.Abstractions/pull/580

PR #580, opened 2020-04-20
On Directory.Move, first check to ensure the parent directory for the indicated target folder exists

pianomanjh:
The solution turns out to be much simpler than I had thought, which is that the target directory structure much exist before copying into a 'new' folder.


pianomanjh:
Let me know if you want me to clean up some of those test cases, it's not very clear specifically what each is uniquely testing.  However, they are all unique variations in source/target path segments.

fgreinacher:
> Let me know if you want me to clean up some of those test cases, it's not very clear specifically what each is uniquely testing. However, they are all unique variations in source/target path segments.

It's fine like that, no worries. Thanks for your endurance on this!

Issue #577, opened 2020-04-17
Add overloads to the CreateNew method of IFileSystemWatcherFactory


nathanpovo:
The [`FileSystemWatcher`](https://docs.microsoft.com/en-us/dotnet/api/system.io.filesystemwatcher.-ctor?view=netframework-4.8) constructor in `System.IO` has 3 overloads:

```csharp
FileSystemWatcher()
FileSystemWatcher(string path)
FileSystemWatcher(string path, string filter)
```

The `IFileSystemWatcherFactory` is missing the constructor which includes the filter.

Also, is there a reason why the 2 methods of `IFileSystemWatcherFactory` are implemented as separate methods instead of using an overload?

I can create a pull request to fix the issue.

fgreinacher:
> The IFileSystemWatcherFactory is missing the constructor which includes the filter.

Correct, seems like nobody needed it up to now 🤣 

> Also, is there a reason why the 2 methods of IFileSystemWatcherFactory are implemented as separate methods instead of using an overload?

I have no idea, that's really weird. I'd vote to obsolete `FromName` and go all in with `CreateNew` overloads.

> I can create a pull request to fix the issue.

Awesome, go for it 🚀

PR #584, opened 2020-04-21
Add overloads the CreateNew method of FileSystemWatcherFactory 

nathanpovo:
Fixes issue #577.


nathanpovo:
Should the `FromPath` method 

https://github.com/System-IO-Abstractions/System.IO.Abstractions/blob/7337796a9ee7c37a25cbc869630f0b2b7cd9812b/System.IO.Abstractions/FileSystemWatcherFactory.cs#L16-L19

be marked as obsolete or removed?

Issue #588, opened 2020-04-30
MockFileSystem sets Temp path to actual file system with no option to change this


IrisClasson:
Apologies if this is intended behavior!

We've been using System.IO.Abstractions at work so we can increase our test coverage to include working with files on disk and the TestingHelpers let's us test more thoroughly that using mocks. 

However, I came across a problem when it comes to working with files that are written to the temp directory. 

The MockFileSystem implementation sets the IPath property in the constructor to a MockFile class that inherits from the MockWrapper class, that calls System.IOPath.GetTempPath(); The Path property has no setter, and I don't see any ways that this part can be mocked, nor can the temp path be changes. The MockFileSystem seems to mix a virtual and an actual system, and therefore there are several ways this can cause unnecessary exceptions.

Here is an example:

```
var mockFileSystem = new MockFileSystem();

var tempPath = mockFileSystem.Path.GetTempPath();

// Returns false
var exists = mockFileSystem.Directory.Exists(tempPath);

```

In our case, we write to the temp path, and the tests verify they can read from the same path. The Read* methods throw, as they verify the directory exists. The directory doesn't exist. 

My workaround has been to get the temp path and add it as a directory to the MockFileSystem

```
var mockFileSystem = new MockFileSystem();

var tempPath = mockFileSystem.Path.GetTempPath();
mockFileSystem.AddDirectory(tempPath);

// Returns true
var exists = mockFileSystem.Directory.Exists(tempPath);
```
Should there a better way to do this (maybe there is and I've missed it)?

Thank you for the hard work btw, the libraries have been immensely useful! Would love to contribute to the project if there is anything up for grabs. Maybe docs?

fgreinacher:
Hey @IrisClasson 👋

Thanks for the detailed issue! This is for sure not the intended behavior, it just seems that nobody ran into it up to now 😄 

We had a similar situation with the `Directory.GetCurrentDirectory` some time ago (see #350) and improved it by hard-coding the initial current directory to `/` and making sure it is created (see #372).

I think we should go for a similar solution for the temp directory:

- `MockFileSystem` should take care of creating the temp directory (maybe something like `/tmp`), similar like it's done for the current directory
- `MockFileSystem.Path.GetTempPath()` should return the path of this directory

What do you think?

fgreinacher:
(I think this would be a good starter if you like to contribute something 🥇)

IrisClasson:
[...]

PR #590, opened 2020-05-04
Feature/588 default virtual temp folder

IrisClasson:
As discussed in the issue #588 I've added a default temp directory to MockFileSystem (ping @fgreinacher )

I'm not 100% sure with my solution to use the previous GetTempPath as a fallback if the path passed in to the MockPath is null or empty (or the original constructor is called). 

Let me know what you think, I'm torn but decided on the fallback as not to add a breaking change.

### DIFF
--- a/System.IO.Abstractions.TestingHelpers/MockFileSystem.cs
+++ b/System.IO.Abstractions.TestingHelpers/MockFileSystem.cs
@@ -11,7 +11,7 @@ namespace System.IO.Abstractions.TestingHelpers
     public class MockFileSystem : IFileSystem, IMockFileDataAccessor
     {
         private const string DEFAULT_CURRENT_DIRECTORY = @"C:\";
-        private const string DEFAULT_TEMP_DIRECTORY = @"C:\temp";
+        private const string TEMP_DIRECTORY = @"C:\temp";
 
         private readonly IDictionary<string, MockFileData> files;
         private readonly PathVerifier pathVerifier;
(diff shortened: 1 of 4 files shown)


System.IO.Abstractions.TestingHelpers/MockFileSystem.cs
```diff
@@ -24,11 +25,13 @@ public MockFileSystem(IDictionary<string, MockFileData> files, string currentDir
                 currentDirectory = XFS.Path(DEFAULT_CURRENT_DIRECTORY);
             }
 
+            var defaultTempDirectory = XFS.Path(DEFAULT_TEMP_DIRECTORY);
```
fgreinacher, line 28: Just a minor nit: I’d leave out the „default“ prefix here as there is no way to change this value form the outside.

System.IO.Abstractions.TestingHelpers/MockPath.cs
```diff
@@ -128,13 +132,18 @@ private string[] GetSegments(params string[] paths)
...
             string fullPath = mockFileDataAccessor.Path.Combine(tempDir, fileName);
 
             mockFileDataAccessor.AddFile(fullPath, new MockFileData(string.Empty));
 
             return fullPath;
         }
+
+        public override string GetTempPath()
+        {
+            return !string.IsNullOrEmpty(defaultTempDirectory) ? defaultTempDirectory : base.GetTempPath();
```
fgreinacher, line 146: Maybe move this logic to the constructor? Seems like making this decision once should be enough 😁

System.IO.Abstractions.TestingHelpers/MockPath.cs
```diff
@@ -11,10 +11,14 @@ namespace System.IO.Abstractions.TestingHelpers
     public class MockPath : PathWrapper
     {
         private readonly IMockFileDataAccessor mockFileDataAccessor;
+        private readonly string defaultTempDirectory;
 
-        public MockPath(IMockFileDataAccessor mockFileDataAccessor) : base(mockFileDataAccessor?.FileSystem)
+        public MockPath(IMockFileDataAccessor mockFileDataAccessor) : this(mockFileDataAccessor,string.Empty) { }
```
fgreinacher, line 16: I also slightly prefer this solution to the breaking change 👍

IrisClasson:
Thanks for the feedback and suggestions @fgreinacher ! I've pushed the changes

fgreinacher:
Thanks again!

PR #602, opened 2020-06-15
Make constructor in PathVerifier public

### LINKED ISSUE #601: PathVerifier only has internal constructor
Looking to implement my own IMockFileDataAccessor and this requires exposing a PathVerifier property, but this class has no public constructor and is therefore impossible to implement. 

Related to issue #597 as this interface should really expose an IPathVerifier but a quick fix is to make the constructor public.

JamesDanielsDDN:
Closes #601


fgreinacher:
Thanks a lot @JamesDanielsDDN!

PR #607, opened 2020-06-16
Reject non-rooted current directory

### LINKED ISSUE #592: MockFileSystem with currentDirectory should return only one currentDirectory
```
{
    { @"myfile.txt", new MockFileData("A") },
}, @"current");
```
makes
`fileSystem.AllFiles.Single()`
return 
`"current\current\current\myfile.txt"`
which is two too many.  
I also believe there should be a prepending `\`.

----

A remedy is to prepend "current" with a slash like so"
```
{
    { @"myfile.txt", new MockFileData("A") },
}, @"\current");
```
to give the result
`"\current\myfile.txt"`

----

Although I present a work around it might be considered a but as it is imho an unexpected behaviour.

fgreinacher:
Fixes #592


fgreinacher:
@jpreese Not sure whether you're still interested in this? If no just say so :)

fgreinacher:
I think I'm now alone here, so I'll merge this as-is :)

PR #638, opened 2020-09-10
Fixes #635 Directory.GetParent(..) behaviour for root directories

### LINKED ISSUE #635: Different Behaviour for Directory.GetParent(..) with root folder as argument
**Describe the bug**
There seems to be a difference between the call to `System.IO.Abstractions.Directory.GetParent(..)` and the normal `System.IO.Directory.GetParent(..)`. Because i think that this package tries to mirror the "to be expected" return values and exceptions i opened up this issue.

**To Reproduce**
We are using the version v12.1.1 of System.IO.Abstractions in our project.

We had a loop to find a specific subfolder starting from an executable file name going up to root.
E.g. if the file is **C:\foo\bar\myapp.exe** we wanted to find **C:\foo\\.peter**. The important part of that loop is our call to
`currentPath = Directory.GetParent(currentPath)`. Afterwards we would check `currentPath == null` to see if we "stepped out of root", so to say.

For example `Directory.GetParent("C:\\") == null`, so our loop would stop there.

Now with `_filesystem.Directory.GetParent("C:\\")` we get an `ArgumentNullException`, even though we never had to handle this exception before.

**Expected behavior**
I think the behaviour of the two methods should be the same?

**Additional context**
If it was a special decission to differ the behaviour then this whole bug can be closed pretty quickly. I did not find any info (quickly) that made me sure it was intended or not, so i figured i would just report it like this. 

After taking a quick look, the problem seems to be 
https://github.com/System-IO-Abstractions/System.IO.Abstractions/blob/a786e3bf4f10f5842925a6d2d7a6eb22edbc15b1/src/System.IO.Abstractions/DirectoryWrapper.cs#L139
where the `null`-Result from the original `Directory.GetParent(..)` just gets passed into the `DirectoryInfoWrapper`-ctor - which throws on `null` arguments:
https://github.com/System-IO-Abstractions/System.IO.Abstractions/blob/a786e3bf4f10f5842925a6d2d7a6eb22edbc15b1/src/System.IO.Abstractions/DirectoryInfoWrapper.cs#L14

I want to finish with thanking you a lot for this package! It made it real simple to test some heavy file accessing in our projects and made us be able to cover our logic with unittests way quicker! Thank you.

IloSophiep:
See issue #635 .

I hope my changes are properly done.
Would i need to adjust any versioning info, because this changes API behaviour..?

### DIFF
--- a/tests/System.IO.Abstractions.Tests/DirectoryWrapperTests.cs
+++ b/tests/System.IO.Abstractions.Tests/DirectoryWrapperTests.cs
@@ -10,27 +10,44 @@ namespace System.IO.Abstractions.Tests
         {
             // Arrange
             var wrapperFilesystem = new FileSystem();
+            var root = wrapperFilesystem.Directory.GetDirectoryRoot(".");
 
             // Act
-            var result = wrapperFilesystem.Directory.GetParent(@"C:\");
+            var result = wrapperFilesystem.Directory.GetParent(root);
 
             // Assert
             Assert.IsNull(result);
         }
 
         [Test]
-        public void GetParent_ForSimpleDirectory_ShouldReturnParent()
+        public void GetParent_ForSimpleSubfolderPath_ShouldReturnRoot()
         {
             // Arrange
-            var expectedBaseDirectory = @"C:\directory1";
-
             var wrapperFilesystem = new FileSystem();
+            var root = wrapperFilesystem.Directory.GetDirectoryRoot(".");
+            var subfolder = wrapperFilesystem.Path.Combine(root, "some-folder");
 
             // Act
-            var result = wrapperFilesystem.Directory.GetParent($@"{expectedBaseDirectory}\file1.txt");
+            var result = wrapperFilesystem.Directory.GetParent(subfolder);
 
             // Assert
-            Assert.AreEqual(expectedBaseDirectory, result.FullName);
+            Assert.AreEqual(root, result.FullName);
+        }
+
+        [Test]
+        public void GetParent_ForSimpleFilePath_ShouldReturnSubfolder()
+        {
+            // Arrange
+            var wrapperFilesystem = new FileSystem();
+            var root = wrapperFilesystem.Directory.GetDirectoryRoot(".");
+            var subfolder = wrapperFilesystem.Path.Combine(root, "some-folder");
+            var file = wrapperFilesystem.Path.Combine(subfolder, "some-file.txt");
+
+            // Act
+            var result = wrapperFilesystem.Directory.GetParent(file);
+
+            // Assert
+            Assert.AreEqual(subfolder, result.FullName);
         }
     }
 }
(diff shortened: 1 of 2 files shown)


tests/System.IO.Abstractions.Tests/DirectoryWrapperTests.cs
```diff
@@ -0,0 +1,36 @@
...
+    public class DirectoryWrapperTests
+    {
+        [Test]
+        public void GetParent_ForRootDirectory_ShouldReturnNull()
+        {
+            // Arrange
+            var wrapperFilesystem = new FileSystem();
+
+            // Act
+            var result = wrapperFilesystem.Directory.GetParent(@"C:\");
```
fgreinacher, line 15: This does not work on Linux because there is no `C:` drive. See e.g. https://github.com/System-IO-Abstractions/System.IO.Abstractions/pull/638/checks?check_run_id=1098272021#step:6:3572

To make this generic you could use `Directory.GetDirectoryRoot`, something like this:

```csharp
var root = wrapperFilesystem.Directory.GetDirectoryRoot(".");
var result = wrapperFilesystem.Directory.GetParent(root);
```

tests/System.IO.Abstractions.Tests/DirectoryWrapperTests.cs
```diff
@@ -0,0 +1,36 @@
...
+
+            // Assert
+            Assert.IsNull(result);
+        }
+
+        [Test]
+        public void GetParent_ForSimpleDirectory_ShouldReturnParent()
+        {
+            // Arrange
+            var expectedBaseDirectory = @"C:\directory1";
```
fgreinacher, line 25: Same comment as above :)

fgreinacher (changes requested):
Thanks a lot for your contribution @IloSophiep. Productive code change looks great, but tests are currently failing on Linux & macOS. See my comments 👍

IloSophiep:
I seem to have broken Unix tests... I'm not quite sure where to start trying to fix them though. I've seen the attribute `[WindowsOnly(WindowsSpecifics.Drives)]` (or other reasons) - but that seems to not be available in the test project i used (and i'm not sure if that's even appropriate?).

Would Unix need their own test with a path like `~/foo/bar` becoming `~/foo` and a test with some "root path" like input?

Any hints for me where to start looking now?

fgreinacher:
> I seem to have broken *nix tests... I'm not quite sure where to start trying to fix them though. I've seen the attribute `[WindowsOnly(WindowsSpecifics.Drives)]` (or other reasons) - but that seems to not be available in the test project i used (and i'm not sure if that's even appropriate?).
> 
> Would *nix need their own test with a path like `~/foo/bar` becoming `~/foo` and a test with some "root path" like input?
> 
> Any hints for me where to start looking now?

Yeah, I also noticed this. See [my comment above](https://github.com/System-IO-Abstractions/System.IO.Abstractions/pull/638#discussion_r487256863).

fgreinacher:
> Edit: Just saw your comments (great timing by me...). Will look into that, thank you!

Really funny, obviously both of us chose Friday nights for OSS work 🤣 

This is merged and will be avaiable on NuGet.org soon. Thanks again 👍

PR #644, opened 2020-09-24
fix Path.Join

### LINKED ISSUE #643: Path.Join methods don't honor Path1
When using the FileSystem object. The PathWrapper object methods, according to source, for the Path.Join methods use 

```cs
public override string Join(ReadOnlySpan<char> path1, ReadOnlySpan<char> path2) =>
            Path.Join(path2, path2);
```

instead of:

```cs
 public override string Join(ReadOnlySpan<char> path1, ReadOnlySpan<char> path2) =>
            Path.Join(path1, path2);
```

**To Reproduce**
1. Go to 'https://github.com/System-IO-Abstractions/System.IO.Abstractions/blob/main/src/System.IO.Abstractions/PathWrapper.cs'

2. Scroll down to 'Line 135'

**Expected behavior**

```cs
 public override string Join(ReadOnlySpan<char> path1, ReadOnlySpan<char> path2) =>
            Path.Join(path1, path2);
```

Change the methods to use Path1, instead of Path2

CaringDev:
use all path segments instead of second one twice, fixes #643

### DIFF
--- a/src/System.IO.Abstractions/PathWrapper.cs
+++ b/src/System.IO.Abstractions/PathWrapper.cs
@@ -133,19 +133,19 @@
 #if FEATURE_PATH_JOIN_WITH_SPAN
         /// <inheritdoc />
         public override string Join(ReadOnlySpan<char> path1, ReadOnlySpan<char> path2) =>
-            Path.Join(path2, path2);
+            Path.Join(path1, path2);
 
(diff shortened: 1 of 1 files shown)


fgreinacher (approved):
Thanks a lot for taking care!

PR #645, opened 2020-09-27
Added possibility to fake file size

simmarn:
This PR introduces a way to fake large files without caring about the file content. Useful if you want to test a function which shall limit directory size or file size.

### DIFF
--- a/src/System.IO.Abstractions.TestingHelpers/MockFileData.cs
+++ b/src/System.IO.Abstractions.TestingHelpers/MockFileData.cs
@@ -10,6 +10,8 @@ namespace System.IO.Abstractions.TestingHelpers
     [Serializable]
     public class MockFileData
     {
+        public const long LengthIsNotFaked = -1;
+
         /// <summary>
         /// The default encoding.
         /// </summary>
@@ -37,6 +39,7 @@ namespace System.IO.Abstractions.TestingHelpers
         /// </summary>
         [NonSerialized]
         private FileSecurity accessControl;
+        private long length = LengthIsNotFaked;
 
         /// <summary>
         /// Gets a value indicating whether the <see cref="MockFileData"/> is a directory or not.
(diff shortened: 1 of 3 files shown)


src/System.IO.Abstractions.TestingHelpers/MockFileData.cs
```diff
@@ -10,6 +10,8 @@ namespace System.IO.Abstractions.TestingHelpers
     [Serializable]
     public class MockFileData
     {
+        public const long LengthIsNotFaked = -1;
```
fgreinacher, line 13: Codacy found an issue: [Change this constant to a 'static' read-only property.](https://app.codacy.com/gh/System-IO-Abstractions/System.IO.Abstractions/pullRequest?prid=6242633)

simmarn: OK. Fixed.

fgreinacher:
Thanks for your contribution @simmarn!

I am very sorry for giving that response to a first-time contributor, but I will not accept this change, for the following reasons:

* The idea of the mock file system is to act as light-weight in-memory filesystem (and yes, I know, the name does not reflect that 🙁) for quick "integration-style" testing. This change works against that idea.
* I don't consider the need for faking file length much different from faking any other file property. I don't want to handle follow-up requests like "Cool I can now fake file length, please also give me an option to fake XYZ".
* I think you can achieve what you want to using a mock framework to fake return values of the relevant methods/properties. See the last code snippet in [the readme](https://github.com/System-IO-Abstractions/System.IO.Abstractions/blob/main/README.md).

Sorry again, but I am the only maintainer here and don't want to introduce "work-multplying" features like this one 😃

simmarn:
> Thanks for your contribution @simmarn!
> 
> I am very sorry for giving that response to a first-time contributor, but I will not accept this change, for the following reasons:
> 
>     * The idea of the mock file system is to act as light-weight in-memory filesystem (and yes, I know, the name does not reflect that slightly_frowning_face) for quick "integration-style" testing. This change works against that idea.
> 
>     * I don't consider the need for faking file length much different from faking any other file property. I don't want to handle follow-up requests like "Cool I can now fake file length, please also give me an option to fake XYZ".
> 
>     * I think you can achieve what you want to using a mock framework to fake return values of the relevant methods/properties. See the last code snippet in [the readme](https://github.com/System-IO-Abstractions/System.IO.Abstractions/blob/main/README.md).
> 
> 
> Sorry again, but I am the only maintainer here and don't want to introduce "work-multplying" features like this one smiley

OK. Understood. I have made a mock using FakeItEasy, but since there were a bunch of methods to fake I thought it would be simpler to just add file length in the constructor.

Issue #651, opened 2020-10-09
Not all matches are returned from DirectoryInfo.EnumerateFileSystemInfo


P-Storm:
**Describe the bug**
Not all matches are returned from DirectoryInfo.EnumerateFileSystemInfo. It looks like that the RecurseSubdirectories is ignored.

**To Reproduce**

```c#
        [Test]
        public void Delete_Me()
        {
            var fileSystem = new MockFileSystem(new Dictionary<string, MockFileData>()
            {
                { @"c:\bla.txt", null },
                { @"c:\Test\test.txt", null },
                { @"c:\Test\more-test.png", null },
                { @"c:\Test\A\deep-nested.png", null },
            });

            var path = @"C:\";
            var searchPattern = "*";
            var enumerateFileSystemEntries = fileSystem.Directory
                .EnumerateFileSystemEntries(path, searchPattern, SearchOption.AllDirectories)
                .OrderBy(r => r);

            var directoryInfo = fileSystem.DirectoryInfo.FromDirectoryName(path);
            var fileSystemInfos = directoryInfo.EnumerateFileSystemInfos(searchPattern, new EnumerationOptions()
                {
                    MatchType = MatchType.Win32,
                    RecurseSubdirectories = true,
                    IgnoreInaccessible = true,
                    ReturnSpecialDirectories = false,
                    AttributesToSkip = FileAttributes.Hidden,
                    MatchCasing = MatchCasing.PlatformDefault,
                })
                .Select(r => r.FullName);

            CollectionAssert.AreEquivalent(enumerateFileSystemEntries, fileSystemInfos);
        }
```

```

  Expected: equivalent to < "c:\bla.txt", "C:\temp", "c:\Test", "c:\Test\A", "c:\Test\A\deep-nested.png", "c:\Test\more-test.png", "c:\Test\test.txt" >
  But was:  < "c:\Test", "C:\temp", "c:\bla.txt" >
  Missing (4): < "c:\Test\A", "c:\Test\A\deep-nested.png", "c:\Test\more-test.png", "c:\Test\test.txt" >
```

**Expected behavior**
I would expect that EnumerateFileSystemEntries and EnumerateFileSystemInfos resulted in the same list. It is doing that for the .net core implementation.

fgreinacher:
Fixed via #652

PR #652, opened 2020-10-09
EnumerationOptions With recursion

P-Storm:
See issue #651

It looks like the code is not checking the `EnumerationOptions.RecurseSubdirectories`.

This fixes my problem with the recursion, but I see more problems in the future with the following settings in:
* MatchType = MatchType.Win32,
* IgnoreInaccessible
* ReturnSpecialDirectories
* AttributesToSkip
* MatchCasing

Will creating a seperate issue for these.


fgreinacher:
Great finding, thanks a lot! Also for the additional issues!

Issue #653, opened 2020-10-09
Add more attributes of EnumerationOptions


P-Storm:
**Describe the bug**
See #652 .

I see more problems in the future with the following settings in:
* MatchType = MatchType.Win32,
* IgnoreInaccessible
* ReturnSpecialDirectories
* AttributesToSkip
* MatchCasing

**To Reproduce**
Use one of the top usings to see if that works.

**Additional context**
I know that `EnumerationOptions` is one of the newest addition of the code, but it is incomplete.

fgreinacher:
With #655 we now explicitly throw `NotSupportedException`s when encountering unsupported `EnumerationOptions` property values.

Chaosflo:
Hey,
When I was testing a class, I came over this problem.
Its already 2 years after, oh man time flies

Whats the status of this?

PR #655, opened 2020-10-12
refactor: throw exception when encountering unsupported EnumerationOptions

fgreinacher:
Addresses #653


fgreinacher:
@P-Storm FYI, this at least makes it obvious which properties are supported and which are not. We can implement the individual properties one-by-one.

PR #665, opened 2020-10-31
fix: do not move unrelated files with same prefix

### LINKED ISSUE #664: TestingHelpers: Directory.Move operates on files with a related name
**Describe the bug**
Windows user here. I suspect this is a bug in the testing helpers?

Directory.Move operates on files with a related name.:
```
- source (dir)
    - dummy (dir)
        - content.txt (file)
    - dummy.txt (file)
- destination (dir)
```

Moving /source/dummy to /destination/dummy will produce the following:
```
- source (dir)
- destination (dir)
    - dummy (dir)
        - content.txt (file)
    - dummy.txt (file)
```

What I expected:
```
- source (dir)
    - dummy.txt (file) <-- not moved
- destination (dir)
    - dummy (dir)
        - content.txt (file)
```

**To Reproduce**
```
[Test]
public void MockDirectory_Move_ShouldOnlyMoveDirAndFilesWithinDir()
{
    {
        // Directory to move and its content:
        {XFS.Path(@"c:\source\dummy"), new MockDirectoryData()},
        {XFS.Path(@"c:\source\dummy\content.txt"), new MockFileData(new byte[] {0})},

        // Unrelated file, not in directory to move, but with a similar name:
        {XFS.Path(@"c:\source\dummy.txt"), new MockFileData(new byte[] {0})},

        // Destination of directory move operation:
        {XFS.Path(@"c:\destination"), new MockDirectoryData()},
    });

    fileSystem.Directory.Move(@"c:\source\dummy", @"c:\destination\dummy");

    Assert.That(fileSystem.FileExists(@"c:\source\dummy.txt"), Is.True);
}
```

Very cool library :thumbsup:

meisenring:
My attempt to fix #664

I'm not very familiar with expected behavior in unix file systems but I guess it covers those as well?

Apologies if this is prematurely.


src/System.IO.Abstractions.TestingHelpers/MockFileSystem.cs
```diff
@@ -236,10 +236,12 @@ public void MoveDirectory(string sourcePath, string destPath)
             sourcePath = FixPath(sourcePath);
             destPath = FixPath(destPath);
 
+            var sourcePathSequence = sourcePath.Split(new[] {Path.DirectorySeparatorChar}, StringSplitOptions.RemoveEmptyEntries);
```
siprbaum, line 239: Should this also split on Path.AltDirectorySeparatorChar ?
E.g. the underlying question is, what happens when you do `Directory.Move("c:/foo/bar", "c:/baz");`
If it is supported on the real filessystem the mockFilessystem should also support it.

src/System.IO.Abstractions.TestingHelpers/MockFileSystem.cs
```diff
@@ -249,6 +251,25 @@ public void MoveDirectory(string sourcePath, string destPath)
                     files.Remove(path);
                 }
             }
+
+            bool PathStartsWith(string path, string[] minMatch)
+            {
+                var pathSequence = path.Split(new[] {Path.DirectorySeparatorChar}, StringSplitOptions.RemoveEmptyEntries);
```
siprbaum, line 257: Same here, should this also split on Path.AltDirectorySeparatorChar ?

src/System.IO.Abstractions.TestingHelpers/MockFileSystem.cs
meisenring: `FixPath` on the lines above (236 and 237) ensures only Path.DirectorySeparatorChar is in the path (Path.AltDirectorySeparatorChar is replaced).

Should it split on Path.AltDirectorySeparatorChar anyway?

siprbaum: Oh, I wasn't aware of that. In that case, it is not needed.
What threw me off is that I couldn't find any tests which ensure that paths with  Path.AltDirectorySeparatorChar are actually handeled correctly, at least I couldn't find any when glancing over the code. 
Pls don't take this the wrong way, this is not a request for you to add those tests!
It is more or less a question if those tests for Path.AltDirectorySeparatorChar exist at all, e.g. at a central place or how else it is ensured that this works

tests/System.IO.Abstractions.TestingHelpers.Tests/MockDirectoryTests.cs
```diff
@@ -1461,6 +1461,27 @@ public void MockDirectory_Move_ShouldMoveDirectoryWithReadOnlySubDirectory()
...
+            {
+                {XFS.Path(@"c:\source\dummy"), new MockDirectoryData()},
+                {XFS.Path(@"c:\source\dummy\content.txt"), new MockFileData(new byte[] {0})},
+                {XFS.Path(@"c:\source\dummy.txt"), new MockFileData(new byte[] {0})},
+                {XFS.Path(@"c:\source\dummy2"), new MockDirectoryData()},
+                {XFS.Path(@"c:\destination"), new MockDirectoryData()},
+            });
+
+            // Act
+            fileSystem.Directory.Move(@"c:\source\dummy", @"c:\destination\dummy");
```
fgreinacher, line 1478:
```suggestion
            fileSystem.Directory.Move(XFS.Path(@"c:\source\dummy"), XFS.Path(@"c:\destination\dummy"));
```

tests/System.IO.Abstractions.TestingHelpers.Tests/MockDirectoryTests.cs
```diff
@@ -1461,6 +1461,27 @@ public void MockDirectory_Move_ShouldMoveDirectoryWithReadOnlySubDirectory()
...
+
+            // Assert
+            Assert.That(fileSystem.FileExists(@"c:\source\dummy.txt"), Is.True);
```
fgreinacher, line 1481:
```suggestion
            Assert.That(fileSystem.FileExists(XFS.Path(@"c:\source\dummy.txt")), Is.True);
```

tests/System.IO.Abstractions.TestingHelpers.Tests/MockDirectoryTests.cs
```diff
@@ -1461,6 +1461,27 @@ public void MockDirectory_Move_ShouldMoveDirectoryWithReadOnlySubDirectory()
...
+            Assert.That(fileSystem.Directory.Exists(@"c:\source\dummy2"), Is.True);
```
fgreinacher, line 1482:
```suggestion
            Assert.That(fileSystem.Directory.Exists(XFS.Path(@"c:\source\dummy2")), Is.True);
```

fgreinacher:
Thanks for your contribution! Changes look alright to me at a first look, happy to do a deeper review when tests are passing.

PR #668, opened 2020-11-10
Updates DirectoryWrapper.SetLastWriteTime to use SetLastWriteTime instead of SetLastAccessTime.

jeramyRR:
Submitting a fix for a probable copy/paste error.


fgreinacher:
Oops, thanks a lot!

PR #695, opened 2021-02-10
feat: increase MockFileSystem constructor speed

jflepp:
The `MockFileSystem` constructor builds the internal `IDictionary<string, MockFileData>` from a given one. While doing so, its performance decreased significantly with large data sets.

This PR splits the internal dictionary into one containing the files and one containing the directories. Some of our internal test durations were decreased by a factor >10 with these changes.

We would be happy if we could get this speedup out of the box.

### DIFF
--- a/src/System.IO.Abstractions.TestingHelpers/MockFileSystem.cs
+++ b/src/System.IO.Abstractions.TestingHelpers/MockFileSystem.cs
@@ -253,21 +253,21 @@ namespace System.IO.Abstractions.TestingHelpers
 
             lock (@lock)
             {
-                MoveFiles(files, sourcePathSequence, sourcePath, destPath);
-                MoveFiles(directories, sourcePathSequence, sourcePath, destPath);
+                Move(files, sourcePathSequence, sourcePath, destPath);
+                Move(directories, sourcePathSequence, sourcePath, destPath);
             }
 
-            void MoveFiles<T>(IDictionary<string, T> dictionary, string[] sourcePathSequence, string sourcePath, string destPath)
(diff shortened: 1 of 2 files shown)


fgreinacher:
Thanks a lot for your contribution! Can you share your test cases with numbers before and after this change? I have some concerns that this improves your specific use cases, but degrades performance for hot paths, mostly due to the double lookups.

jflepp:
About the numbers: We implemented our use-case in the test `MockFileSystem_Constructor_InitializedQuicklyWithLargeAmountOfData`. On our system, the test takes about 14 minutes without the changes, with the changes about 4 seconds.

The double lookups do not affect how the code scales, it's just a constant that got a bit larger, whereas the adjustments we made changed the time complexity of the algorithm from super- to sub-linear.

fgreinacher:
> About the numbers: We implemented our use-case in the test `MockFileSystem_Constructor_InitializedQuicklyWithLargeAmountOfData`. On our system, the test takes about 14 minutes without the changes, with the changes about 4 seconds.
> 
> The double lookups do not affect how the code scales, it's just a constant that got a bit larger, whereas the adjustments we made changed the time complexity of the algorithm from super- to sub-linear.

Oh sorry, I somehow missed the new test case. Thanks a lot, I'll have a deeper look next week 🤝

schibu007:
After discussion with @jflepp, I created PR #698 as an alternative approach. @fgreinacher Please let us know, in which direction you prefer to go.

fgreinacher:
Closing in favor of #698

PR #698, opened 2021-02-15
feat: speed up MockFileSystem constructor

schibu007:
This PR is an alternative to the PR #695. It has a slightly better performance and does not change the concept of having all paths in one collection.
I created this PR after discussion with @jflepp.

PR #700, opened 2021-02-17
fix: file move after directory move doesn't work correctly

jflepp:
This bug was introduced by #698


fgreinacher:
Thanks for taking care 👍

PR #701, opened 2021-02-17
fix: move on Unix systems doesn't work correctly

jflepp:
While fixing #700, we encountered another bug: When moving a directory that contains a file with the same name, the `MoveDirectory` method not only renamed the directory but also the file. This appears to be a UNIX problem only.


jflepp:
Note: This will probably create a merge conflict with #700. We would prefer to have #700  merged first and will then resolve the conflicts

fgreinacher:
@jflepp #700 is merged, I'll merge this after the conflicts are resolved 🤝

jflepp:
Merged changes from Main.

@fgreinacher Thanks for your fast reviews! :smile:

PR #714, opened 2021-03-09
fix: Pass `FileAccess` to open function when FileMode is `Create` or `CreateNew`

### LINKED ISSUE #713: FileAccess parameter in File.Open() is ignored when the FileMode is Create or CreateNew.
**Describe the bug**

FileAccess parameter in File.Open() is ignored when the FileMode is Create or CreateNew. Bug can be observed by checking the `CanRead` property of the returned `Stream` with `FileAccess.ReadWrite`.

**To Reproduce**
Below code snippet compares actual behavior to the mock:
```cs
var tempFile = Path.GetTempFileName();
using var fs = File.Open(tempFile, FileMode.Create);
Console.WriteLine(fs.CanRead);

var mock = new MockFileSystem();
var mockFile = mock.Path.GetTempFileName();
using var mockFs = mock.File.Open(mockFile, FileMode.Create);
Console.WriteLine(mockFs.CanRead);
```

Output:
```
True
False
```

**Expected behavior**

Mock behavior should match the .NET type.

willson556:
Resolves #713.

### DIFF
--- a/src/System.IO.Abstractions.TestingHelpers/MockFile.cs
+++ b/src/System.IO.Abstractions.TestingHelpers/MockFile.cs
@@ -126,9 +126,9 @@ namespace System.IO.Abstractions.TestingHelpers
             Create(path, bufferSize, FileOptions.None);
 
         public override Stream Create(string path, int bufferSize, FileOptions options) =>
-            CreateInternal(path, options);
+            CreateInternal(path, FileAccess.Write, options);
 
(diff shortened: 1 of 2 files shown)


fgreinacher:
Looking great, thanks a lot for your contribution @willson556 👍

willson556:
Thank you for processing it so quickly!

PR #715, opened 2021-03-13
fix: Updated MockDirectoryInfo.Name to be consistent with DirectoryInfo.Name

### LINKED ISSUE #697: DirectoryInfo.Name returns different result for root path between mock and real file system
**Describe the bug**
When you call DirectoryInfo.Name, the result is an empty string for a root path in the mocked version. The real file system returns "C:\\". 

**To Reproduce**
I'm using version 13.2.10 and the TestingHelpers project from NuGet. These unit tests succeed, which should not happen:

 ``` csharp
[Test]
public void DirectoryName_DifferentForRoot()
{
    var mock = new MockFileSystem();
    var real = new FileSystem();

    var nameInMock = mock.DirectoryInfo.FromDirectoryName("C:\\").Name;
    var nameInReal = real.DirectoryInfo.FromDirectoryName("C:\\").Name;

    Assert.AreEqual("", nameInMock);
    Assert.AreEqual("C:\\", nameInReal);
}

[Test]
public void DirectoryName_IdenticalForFile()
{
    var mock = new MockFileSystem();
    var real = new FileSystem();
    
    var nameInMock = mock.DirectoryInfo.FromDirectoryName("C:\\test.sldprt").Name;
    var nameInReal = real.DirectoryInfo.FromDirectoryName("C:\\test.sldprt").Name;

    Assert.AreEqual("test.sldprt", nameInMock);
    Assert.AreEqual("test.sldprt", nameInReal);
}
```

**Expected behavior**
I'd expect the directory name to always be the same for mocked and real file systems.

**Additional context**
Link to the docs: https://docs.microsoft.com/en-us/dotnet/api/system.io.directoryinfo.name?view=net-5.0

JordanBowker:
Fixes #697. When constructed with only the root; MockDirectoryInfo.Name now returns the the root value instead of an empty string. Making it consistent with DirectoryInfo.Name.

### DIFF
--- a/src/System.IO.Abstractions.TestingHelpers/MockDirectoryInfo.cs
+++ b/src/System.IO.Abstractions.TestingHelpers/MockDirectoryInfo.cs
@@ -120,7 +120,11 @@ namespace System.IO.Abstractions.TestingHelpers
 
         public override string Name
         {
-            get { return new MockPath(mockFileDataAccessor).GetFileName(directoryPath.TrimEnd(mockFileDataAccessor.Path.DirectorySeparatorChar)); }
+            get
+            {
+                var mockPath = new MockPath(mockFileDataAccessor);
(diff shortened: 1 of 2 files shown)


fgreinacher:
Thanks!

PR #719, opened 2021-04-01
feat: Made a change to auto implemented properties on MockDriveInfo to actu…

YamSager:
…ally set the values

### DIFF
--- a/src/System.IO.Abstractions.TestingHelpers/MockDriveInfo.cs
+++ b/src/System.IO.Abstractions.TestingHelpers/MockDriveInfo.cs
@@ -38,9 +38,9 @@
             IsReady = true;
         }
 
-        public new long AvailableFreeSpace { get; set; }
-        public new string DriveFormat { get; set; }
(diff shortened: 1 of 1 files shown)


YamSager:
In regards to #274, users should be able to set the DriveInfo if need be. As of now, if the code under test checks the available free space, the test will fail and there is no workaround. This change will allow users to set the value and for the value to take hold to spoof the free space, drive type etc. if needed.

fgreinacher:
Sorry @YamSager, I somehow missed this PR.

I still have the same opinion as in https://github.com/System-IO-Abstractions/System.IO.Abstractions/issues/274#issuecomment-770445316:

> Hm, I wouldn't consider this a bug, but rather a bad API. The properties at MockDriveInfo shouldn't be set-able.

I'd rather like to solve this at the `MockFileSystem` level, where we already allow people to customize the behavior of the mocks.

Could you create an issue that describes your use case? We could then discuss solution options there.

Issue #562, opened 2020-02-27
Lack of Path.TrimEndingDirectorySeparator on .NET Core 3.1


Stuv7CB:
Lack of Path.TrimEndingDirectorySeparator on .NET Core 3.1

https://docs.microsoft.com/en-us/dotnet/api/system.io.path.trimendingdirectoryseparator?view=netcore-3.1

fgreinacher:
We don't target .NET Core for maintainability reasons. Feel free to add this via an extension method on `IPath`.

PR #777, opened 2021-12-12
feat: add missing path APIs

### LINKED ISSUE #774: 'await' cannot be used in an expression containing the type 'System.ReadOnlySpan<char>'
**Describe the bug**
When utilizing System.IO.Abstractions for Path.Join, the following code snippet does not compile on .NET 6, however it does when using System.IO Path.Join.

`
var path = System.IO.Abstractions.FileSystem.Path.Join("/config/covers", await _unitOfWork.ChapterRepository.GetChapterCoverImageAsync(chapterId));

public async Task<string> GetChapterCoverImageAsync(int chapterId)
    {

        return await _context.Chapter
            .Where(c => c.Id == chapterId)
            .Select(c => c.CoverImage)
            .AsNoTracking()
            .SingleOrDefaultAsync();
    }
`

**Expected behavior**
I expect that the Abstraction works identically to System.IO and it builds and functions correctly. 

**Additional context**
I am using .NET 6 and System.IO.Abstractions 14.0.3

fgreinacher:
This adds support for all `System.IO.Path` members in .NET 5 and 6.

BREAKING CHANGE: This is a breaking change for people implementing `System.IO.Abstractions.IPath` or deriving from `System.IO.Abstractions.PathBase` because new members have been added to their API surface.

Fixes #774


12emerta:
Sorry if I should open a new issue rather than comment here (I can open a more descriptive issue if needed).

The Path.Join methods which take strings are also shipped with Core 3.0 and Core 3.1, not just .Net 5 and 6 (these methods were put behind conditional compilation constant "FEATURE_PATH_JOIN_WITH_PARAMS" in this PR).

```
string Join(params string[] paths);
string Join(string path1, string path2);
string Join(string path1, string path2, string path3);
```

fgreinacher:
> Sorry if I should open a new issue rather than comment here (I can open a more descriptive issue if needed).
> 
> The Path.Join methods which take strings are also shipped with Core 3.0 and Core 3.1, not just .Net 5 and 6 (these methods were put behind conditional compilation constant "FEATURE_PATH_JOIN_WITH_PARAMS" in this PR).
> 
> ```
> string Join(params string[] paths);
> string Join(string path1, string path2);
> string Join(string path1, string path2, string path3);
> ```

@12emerta This package [does not target .NET Core 3.1](https://github.com/TestableIO/System.IO.Abstractions/blob/main/src/System.IO.Abstractions/System.IO.Abstractions.csproj#L6). If these APIs are not available in .NET Standard we won' be able to expose them (see https://github.com/TestableIO/System.IO.Abstractions/issues/562 and https://github.com/TestableIO/System.IO.Abstractions/pull/540 for some more reasoning).

12emerta:
> > Sorry if I should open a new issue rather than comment here (I can open a more descriptive issue if needed).
> > The Path.Join methods which take strings are also shipped with Core 3.0 and Core 3.1, not just .Net 5 and 6 (these methods were put behind conditional compilation constant "FEATURE_PATH_JOIN_WITH_PARAMS" in this PR).
> > ```
> > string Join(params string[] paths);
> > string Join(string path1, string path2);
> > string Join(string path1, string path2, string path3);
> > ```
> 
> @12emerta This package [does not target .NET Core 3.1](https://github.com/TestableIO/System.IO.Abstractions/blob/main/src/System.IO.Abstractions/System.IO.Abstractions.csproj#L6). If these APIs are not available in .NET Standard we won' be able to expose them (see #562 and #540 for some more background).

Thanks for the quick reply! I didn't realize Core 3.x wasn't supported.  Will try and do some inheritance magic as a workaround.

PR #791, opened 2022-01-08
fix: make Mock{File,Directory}Info cache file attributes

### LINKED ISSUE #661: Discrepancy between DirectoryInfo(Base).Exists in real file system and MockFileSystem DirectoryInfoBase.Exists
**Describe the bug**
System.IO.DirectoryInfo.Exists seems to store the "exists" information at the time of creation. MockFileSystem seems to use up-to-date information.

**To Reproduce**
~~~
        public void MockFileSystemDirectoryInfoExists_AfterCreate_ShouldBeFileSystemDirectoryInfoExists()
        {
            var folder = Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString());
            var di = new DirectoryInfo(folder);
            Debug.Assert(!di.Exists, "!di.Exists");
            di.Create();

            var mfs = new MockFileSystem();
            var di_mfs = mfs.DirectoryInfo.FromDirectoryName(folder);
            Debug.Assert(!di_mfs.Exists, "!di_mfs.Exists");
            di_mfs.Create();

            di_mfs.Exists.Should().Be(di.Exists);
        }
~~~

**Expected behavior**
Mock File System should behave like the real file system.

**Additional context**
.NET 4.6.2; System.IO.Abstractions 3.0.2

fgreinacher:
This makes them behave more similar to the real file system implementation

Fixes #661

### DIFF
--- a/src/System.IO.Abstractions.TestingHelpers/MockDirectoryInfo.cs
+++ b/src/System.IO.Abstractions.TestingHelpers/MockDirectoryInfo.cs
@@ -13,6 +13,8 @@ namespace System.IO.Abstractions.TestingHelpers
         private readonly IMockFileDataAccessor mockFileDataAccessor;
         private readonly string directoryPath;
         private readonly string originalPath;
+        private MockFileData cachedMockFileData;
+        private bool refreshOnNextRead;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="MockDirectoryInfo"/> class.
@@ -33,6 +35,7 @@ namespace System.IO.Abstractions.TestingHelpers
                 directoryPath = directoryPath.TrimEnd(' ');
             }
             this.directoryPath = directoryPath;
+            Refresh();
         }
 
         /// <inheritdoc />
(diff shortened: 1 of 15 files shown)


fgreinacher:
@brianmcbrayer Wanna have a look at this? (I warned you in advance 😀)

BrianMcBrayer:
Yeah! I'll take a look today!

src/System.IO.Abstractions.TestingHelpers/MockDirectoryInfo.cs
```diff
@@ -380,11 +383,17 @@ public override IDirectoryInfo Root
 
         private MockFileData GetMockFileDataForRead()
         {
-            return mockFileDataAccessor.GetFile(directoryPath) ?? MockFileData.NullObject;
+            if (refreshOnNextRead)
```
BrianMcBrayer, line 386: In a multithreaded scenario, is this thread-safe? I think it probably would be, but it might not be as deterministic as the file system. Now that I've said that, the file system is often not too deterministic, so it's probably okay.

BrianMcBrayer (approved):
Looks really good and clean.

I had one comment/question about multithreading, but I don't think it's an issue.

src/System.IO.Abstractions.TestingHelpers/MockDirectoryInfo.cs
fgreinacher: Good hint! I'd say it's fine - worst case we'd refresh more than once.

fgreinacher:
> Looks really good and clean.
> 
> I had one comment/question about multithreading, but I don't think it's an issue.

Thanks a lot for the quick review!

Issue #779, opened 2021-12-14
IFile.Create has different return type than File.Create


Heldenkrieger01:
**Describe the bug**
The `IFile.Create()` method returns a `Stream`, but the "original" `File.Create()` returns a `FileStream`. This is important, because the `FileStream` contains the `Name`-property with the files name.

**Expected behavior**
`IFile.Create()` returns a `FileStream`.

fgreinacher:
This is more or less by design, because `FileStream` relies on the real file system.

We could return a custom subclass of `Stream` that adds the `Name` property.

What do you think about this?

Heldenkrieger01:
That would be great, especially for people like me, that move to Abstractions with existing code.

fgreinacher:
Cool, so let's make this as ready and discuss details in a PR, once somebody finds the time to work on this 👍

vbreuss:
I think this issue can be closed, as the `FileSystemStream` was implemented in #906 ...
Do you agree, @Heldenkrieger01 ?

fgreinacher:
I'm closing this. Feel free to reopen if needed!

PR #793, opened 2022-01-08
feat: Add `Name` to `IFile.Create` (and others)

BrianMcBrayer:
Added draft implementation of a `FileStream`-like class It is not plumbed in at all because I want to discuss the approach first.

@fgreinacher is this kind of what you were thinking in #779 ? I figured if this is correct, I can go through and update all the related `Stream` references to return `FileStream` with the `Name` property properly set.

### DIFF
--- a/src/System.IO.Abstractions/FileStream.cs
+++ b/src/System.IO.Abstractions/FileStream.cs
@@ -1,0 +1,58 @@
+﻿namespace System.IO.Abstractions
+{
+    /// <summary>
+    /// A <see cref="Stream"/> that has properties similar to <see cref="System.IO.FileStream"/>
+    /// </summary>
+    public class FileStream : Stream
+    {
+        private readonly Stream _wrappedStream;
+
+        /// <summary>
+        /// Constructor for <see cref="FileStream"/>
+        /// </summary>
+        /// <param name="wrappedStream">The underlying wrapped <see cref="Stream"/></param>
+        /// <param name="name">The name of the underlying stream object</param>
+        public FileStream(Stream wrappedStream, string name)
+        {
+            _wrappedStream = wrappedStream;
+
+            Name = name;
+        }
+
+        /// <summary>
+        /// Name of the underlying stream object.
+        /// See <see cref="System.IO.FileStream.Name"/>
+        /// </summary>
+        public string Name { get; }
+
+        /// <inheritdoc />
+        public override bool CanRead => _wrappedStream.CanRead;
+
+        /// <inheritdoc />
+        public override bool CanSeek => _wrappedStream.CanSeek;
+
+        /// <inheritdoc />
+        public override bool CanWrite => _wrappedStream.CanWrite;
+
+        /// <inheritdoc />
+        public override long Length => _wrappedStream.Length;
+
+        /// <inheritdoc />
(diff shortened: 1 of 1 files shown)


BrianMcBrayer:
Just for reference, there are a few other fields that I am not sure I want to tackle now that we could eventually add here:
- `Handle` (obsolete)
- `IsAsync` - Gets a value that indicates whether the FileStream was opened asynchronously or synchronously
- `SafeFileHandle` - Gets a SafeFileHandle object that represents the operating system file handle for the file that the current FileStream object encapsulates

(see https://docs.microsoft.com/en-us/dotnet/api/system.io.filestream?view=net-6.0#properties)

fgreinacher:
> @fgreinacher is this kind of what you were thinking in #779 ? I figured if this is correct, I can go through and update all the related `Stream` references to return `FileStream` with the `Name` property properly set.

Yes, that's what I had In mind, thanks for taking care!

src/System.IO.Abstractions/FileStream.cs
```diff
@@ -0,0 +1,58 @@
+﻿namespace System.IO.Abstractions
+{
+    /// <summary>
+    /// A <see cref="Stream"/> that has properties similar to <see cref="System.IO.FileStream"/>
+    /// </summary>
+    public class FileStream : Stream
```
fgreinacher, line 6: I think we should use another name for this to prevent conflicts with `System.IO.FileStream`.

I don't have a good idea, maybe `FileStreamBase` to mirror other types here?

fgreinacher:
> Just for reference, there are a few other fields that I am not sure I want to tackle now that we could eventually add here:
> 
> * `Handle` (obsolete)
> * `IsAsync` - Gets a value that indicates whether the FileStream was opened asynchronously or synchronously
> * `SafeFileHandle` - Gets a SafeFileHandle object that represents the operating system file handle for the file that the current FileStream object encapsulates
> 
> (see https://docs.microsoft.com/en-us/dotnet/api/system.io.filestream?view=net-6.0#properties)

I think we can ignore `Handle` and `SafeFileHandle` as these would be impossible to mock in a useful way. I'm unsure about `IsAsync` - do you have an idea what this is commonly used for?

BrianMcBrayer:
> > Just for reference, there are a few other fields that I am not sure I want to tackle now that we could eventually add here:
> > 
> > * `Handle` (obsolete)
> > * `IsAsync` - Gets a value that indicates whether the FileStream was opened asynchronously or synchronously
> > * `SafeFileHandle` - Gets a SafeFileHandle object that represents the operating system file handle for the file that the current FileStream object encapsulates
> > 
> > (see https://docs.microsoft.com/en-us/dotnet/api/system.io.filestream?view=net-6.0#properties)
> 
> I think we can ignore `Handle` and `SafeFileHandle` as these would be impossible to mock in a useful way. I'm unsure about `IsAsync` - do you have an idea what this is commonly used for?

`IsAsync` is something that can be specified when opening a file that will hint to the system whether IO should be on a `ThreadPool` thread or not. I'm not very familiar with it.

I think I should add it anyway, since it's just a pass-through property and it's readonly. I'll do that and the other things in the review today, and then I'll start wiring this in.

BrianMcBrayer:
I agree. I feel bad I did not get around to getting this PR in when it was needed 😬

PR #790, opened 2022-01-07
feat: Add `LinkTarget` to `FileSystemInfo` for .NET 6 targets

BrianMcBrayer:
## Overview

In .NET 6, additional symlinks support was added to `FileSystemInfo`. This PR adds at least one of those properties here, the `LinkTarget` property.

See https://github.com/TestableIO/System.IO.Abstractions/issues/789

We need to discuss the mock portion of this though as I'm not sure how this codebase expects that.

PR #795, opened 2022-01-12
fix: add `LinkTarget` to `IFileSystemInfo`

BrianMcBrayer:
In https://github.com/TestableIO/System.IO.Abstractions/pull/790 we added `LinkTarget` to `FileSystemInfoBase`, but I forgot to also add it to the interface (which was the whole point in the first place).

So this is embarrassing, but also it will allow `LinkTarget` to be used and mocked.


BrianMcBrayer:
@fgreinacher I missed this during my last PR. Sorry!

BrianMcBrayer:
On a related note, I do not think the tests are testing using .net6, or else they would have blown up (I think). Taking a look.

fgreinacher:
Thanks @BrianMcBrayer!

The tests are indeed running against .NET 6, but they don't assert anything the the interfaces and base classes are equivalent. Might be something to improve 🤔

PR #802, opened 2022-02-01
fix: initialize streams returned by MockFile.Create with ReadWrite access

### LINKED ISSUE #801: MockFileSystem.File.Create(string) returns a stream with CanRead false
`MockFileStream`, when returned as a result of a `MockFileSystem.File.Create(string path)` call, incorrectly has `CanRead` set to false.
This is different than `System.IO.File.Create(string path)`, which correctly sets `CanRead` to true.  

`CanRead` being false, can result in problems when used with code that expects true, for example, `new BinaryReader(Stream stream)`, which checks the `CanRead` property in it's ctor, and throws when not true.

pianomanjh:
Fixes #801


fgreinacher:
Thanks a lot for you contribution @pianomanjh!

PR #807, opened 2022-02-03
fix: make MockFileInfo.Delete honor FileAccess

### LINKED ISSUE #794: Mock FileInfo.Delete does not consider MockFileData.AllowedFileShare property
**Describe the bug**
Mock FileInfo.Delete does not consider the MockFileData.AllowedFileShare property during the delete operation. Similar to #747, but on the FileInfo interface.

**To Reproduce**
The following test reproduces the error:
``` CSharp
 [Test]
        public void MockFileInfo_Delete_ShouldThrowIfFileAccessShareHasNoWriteOrDeleteAccess()
        {
            // Arrange
            var fileSystem = new MockFileSystem(new Dictionary<string, MockFileData>(StringComparer.OrdinalIgnoreCase)
            {
                { @"c:\bar\foo.txt", new MockFileData("some demo content") { AllowedFileShare = System.IO.FileShare.None } },
                { @"c:\bar\baz.txt", new MockFileData("some demo content") { AllowedFileShare = System.IO.FileShare.None } },
            });

            // Throws exception, as expected
            Assert.Throws(typeof(System.IO.IOException), () => fileSystem.File.Delete(@"c:\bar\baz.txt"));

            var fi = fileSystem.FileInfo.FromFileName(@"c:\bar\foo.txt");

            // Does not throw exception, but should
            Assert.Throws(typeof(System.IO.IOException), () => fi.Delete());
        }
```

**Expected behavior**
As the AllowedFileShare  is set to NONE, an exception should be throw when the delete is attempted.

Thanks!

pianomanjh:
- update `MockFileInfo` to proxy `Delete` to `MockFile.Delete` (as is done for most other operations in `MockFileInfo`) Fixes #794 
- normalize the use of `MockFile` for static file operations in `MockFileInfo`

### DIFF
--- a/src/System.IO.Abstractions.TestingHelpers/MockFileInfo.cs
+++ b/src/System.IO.Abstractions.TestingHelpers/MockFileInfo.cs
@@ -12,6 +12,7 @@ namespace System.IO.Abstractions.TestingHelpers
         private string path;
         private readonly string originalPath;
         private MockFileData cachedMockFileData;
+        private MockFile mockFile;
         private bool refreshOnNextRead;
 
         /// <inheritdoc />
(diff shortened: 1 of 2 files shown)


fgreinacher:
Awesome, thanks for fixing and also for making it consistent 👍

Issue #810, opened 2022-02-09
File.Exists throws an ArgumentException when empty string paths are used


masood-haloratech:
**Describe the bug**
MockFileSystem throws an System.ArgumentException ("The path is not of a legal form") when a empty string is passed to the File.Exists() method.

System.IO.File.Exists(string.Empty) returns false.

**To Reproduce**
```
        [Test]
        public void MockFile_Exists_ShouldReturnFalseForEmptyString()
        {
            // Arrange
            var fileSystem = new MockFileSystem();

            // Act
            var result = fileSystem.File.Exists(string.Empty);

            // Assert
            Assert.IsFalse(result);
        }
```

**Expected behavior**
fileSystem.File.Exists() should return false and behave like System.IO.File.Exists()

fgreinacher:
Thanks for reporting!

Seems we don't respect https://docs.microsoft.com/en-us/dotnet/api/system.io.file.exists?view=net-6.0#remarks.

PR welcome 😀

PR #811, opened 2022-02-10
fix: make MockFile.Exists handle empty string

masood-haloratech:
PR created in relation to #810

Added check for path.Length in MockFile and added corresponding unit test.

I am a bit new to contributing so please let me know if I have missed anything :)

### DIFF
--- a/src/System.IO.Abstractions.TestingHelpers/MockFile.cs
+++ b/src/System.IO.Abstractions.TestingHelpers/MockFile.cs
@@ -208,6 +208,7 @@ namespace System.IO.Abstractions.TestingHelpers
                 return false;
             }
 
+            //Not handling exceptions here so that mock behaviour is as similar as possible to System.IO.File.Exists (See #810)
             try
             {
                 mockFileDataAccessor.PathVerifier.IsLegalAbsoluteOrRelative(path, nameof(path));
--- a/src/System.IO.Abstractions.TestingHelpers/MockFile.cs
+++ b/src/System.IO.Abstractions.TestingHelpers/MockFile.cs
@@ -203,13 +203,23 @@ namespace System.IO.Abstractions.TestingHelpers
                 return false;
             }
 
-            if (path.Length == 0)
+            if (path.Trim() == string.Empty)
             {
                 return false;
             }
 
-            var file = mockFileDataAccessor.GetFile(path);
-            return file != null && !file.IsDirectory;
+            try
+            {
+                mockFileDataAccessor.PathVerifier.IsLegalAbsoluteOrRelative(path, nameof(path));
+
+                var file = mockFileDataAccessor.GetFile(path);
+                return file != null && !file.IsDirectory;
+            }
+            catch (ArgumentException) { }
+            catch (IOException) { }
+            catch (UnauthorizedAccessException) { }
+
+            return false;
         }
 
         /// <inheritdoc />
--- a/src/System.IO.Abstractions.TestingHelpers/MockFile.cs
+++ b/src/System.IO.Abstractions.TestingHelpers/MockFile.cs
@@ -216,6 +216,7 @@ namespace System.IO.Abstractions.TestingHelpers
                 return file != null && !file.IsDirectory;
             }
             catch (ArgumentException) { }
+            catch (NotSupportedException) { }
             catch (IOException) { }
             catch (UnauthorizedAccessException) { }
 
--- a/src/System.IO.Abstractions.TestingHelpers/MockFile.cs
+++ b/src/System.IO.Abstractions.TestingHelpers/MockFile.cs
@@ -203,6 +203,11 @@ namespace System.IO.Abstractions.TestingHelpers
                 return false;
             }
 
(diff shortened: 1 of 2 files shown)


src/System.IO.Abstractions.TestingHelpers/MockFile.cs
```diff
@@ -203,6 +203,11 @@ public override bool Exists(string path)
                 return false;
             }
 
+            if (path.Length == 0)
+            {
+                return false;
+            }
+
```
siprbaum, line 210: I am wondering if the scope of this PR should be extended, because now this handles `path == path.Empty`, but
according to https://docs.microsoft.com/en-us/dotnet/api/system.io.file.exists?view=net-6.0 it should also handle other cases:

> This method also returns false if path is null, an invalid path, or a zero-length string. If the caller does not have
> sufficient permissions to read the specified file, no exception is thrown and the method returns false regardless of
> the existence of path.

Perhaps something like
```
try {
    mockFileDataAccessor.PathVerifier.IsLegalAbsoluteOrRelative(path, nameof(path));
} catch (...) {
    return null;
}
```

Actually i don't like throwing and catching the exception, but the `PathVerifier.IsLegalAbsoluteOrRelative()` logic could be refactored or (extracted into a seperate method) were we can check before throwing, which could then be used here. But I believe for making my point the conceptual suggestion above is good enough.

HINT: Take this info with a grain of salt, as I am not a contributor to this project and have just seen this PR and wondered about the other illegal usecases and then went to look up what else is also not allowed besides an empty path. 
But all these cases should be checked with the real `File.Exists` behaviour first.

fgreinacher: Yes, I agree that we should handle the other cases in this PR.

I'd also be prefer to not use exception handling for flow control.

@masood-haloratech Feel free to ping me (`@fgreinacher`) anytime if you're stuck. I'll try to help as good as I can!

masood-haloratech: I am not a big fan of using exceptions for flow control either but looking at the code from the dotnet runtime it seems that is what they have done too. 

Link to code [here](https://github.com/dotnet/runtime/blob/1ba0394d71a4ea6bee7f6b28a22d666b7b56f913/src/libraries/System.Private.CoreLib/src/System/IO/File.cs#L88) (Hopefully I am looking at the correct code)

```
        public static bool Exists([NotNullWhen(true)] string? path)
        {
            try
            {
                if (path == null)
                    return false;
                if (path.Length == 0)
                    return false;

                path = Path.GetFullPath(path);

                // After normalizing, check whether path ends in directory separator.
                // Otherwise, FillAttributeInfo removes it and we may return a false positive.
                // GetFullPath should never return null
                Debug.Assert(path != null, "File.Exists: GetFullPath returned null");
                if (path.Length > 0 && PathInternal.IsDirectorySeparator(path[path.Length - 1]))
                {
                    return false;
                }

                return FileSystem.FileExists(path);
            }
            catch (ArgumentException) { }
            catch (IOException) { }
            catch (UnauthorizedAccessException) { }

            return false;
        }
```
Also discussions [here](https://github.com/dotnet/runtime/issues/24937) indicate that any exceptions should return false too.

@fgreinacher I can have a go at refactoring `PathVerifier.IsLegalAbsoluteOrRelative()` but there is a chance that it may still not behave the same. Wouldn't it be better trying to stick as close as possible with the dotnet runtime as above.

fgreinacher: Thanks for looking at the .NET sources. Yes sure, then let's go for a similar approach, probably would be close to what  @siprbaum suggested above.

masood-haloratech: I wasn't too sure how you would like to test invalid paths as there could be so many combinations so at the moment I have just added one. I had a look at writing a test for UnauthorizedAccessException but it looks like `MockFileSystem` does not do any access control checks.

src/System.IO.Abstractions.TestingHelpers/MockFile.cs
```diff
@@ -203,8 +203,24 @@ public override bool Exists(string path)
...
+            try
+            {
+                mockFileDataAccessor.PathVerifier.IsLegalAbsoluteOrRelative(path, nameof(path));
+
+                var file = mockFileDataAccessor.GetFile(path);
+                return file != null && !file.IsDirectory;
+            }
+            catch (ArgumentException) { }
```
fgreinacher, line 218: @masood-haloratech Please add clarifying comments in the empty catch blocks so that it's clear why we don't handle the exceptions.

src/System.IO.Abstractions.TestingHelpers/MockFile.cs
fgreinacher: That's totally fine!

fgreinacher (approved):
LGTM, thanks a ton!

Issue #169, opened 2016-08-31
[MockFileSystem] Directory discover could return the directory self


zvirja:
Consider the following code:

``` csharp
      var mfs = new MockFileSystem();
      mfs.Directory.CreateDirectory("dir");
      var directoriesRelative = mfs.Directory.GetDirectories("dir");
      var directoriesAbs = mfs.Directory.GetDirectories(mfs.Path.GetFullPath("dir"));
```

You will see that `directoriesRelative` contains the `dir` folder itself, while `directoriesAbs` is empty as expected.

This is a critical bug because it forces to rewrite production code to cover it later by tests.

Likely, in order to fix the issue the following two lines should be reordered: [MockDirectory.cs](https://github.com/tathamoddie/System.IO.Abstractions/blob/master/TestingHelpers/MockDirectory.cs#L180).

MCCoolMoDAue:
Hi Zvirja, I am looking into your issue.
Can you elaborate which lines you mean?
only one line is highighted

Zvirja:
Good catch :) I believe that [these two lines](https://github.com/tathamoddie/System.IO.Abstractions/blob/master/TestingHelpers/MockDirectory.cs#L180-L181) should be reordered among themselves.

Will be really thankful if this bug is fixed.

tathamoddie:
Fixed in #258, shipped to NuGet as 2.1.0.174

PR #816, opened 2022-02-24
fix: GetDirectories returns relative paths if applicable

### LINKED ISSUE #815: MockFileSystem.Directory.GetDirectories returns absolute paths instead of relative paths
**Describe the bug**
Calling `MockFileSystem.Directory.GetDirectories(@"some\path")` returns absolute paths instead of relative paths

**To Reproduce**
In this code
```c#
var mockFs = new MockFileSystem();
mockFs.Directory.CreateDirectory(@"foo\bar");
var output = mockFs.Directory.GetDirectories("foo");
```
`output` is an array with a single string: `C:\foo\bar`, which is incorrect.

**Expected behavior**
In the same example, `output` should be an array with a single string `foo\bar`

**Additional context**
Here's a more comprehensive code snippet that shows:
- how the System.IO.Directory class behaves
- that FileSystem.Directory behaves correctly
- that MockFileSystem.Directory behaves incorrectly

```c#
[TestMethod]
public void Directory_GetDirectories_Should_Return_Relative_Paths()
{
    //This shows how the system classes behave : it returns one relative directory
    Directory.CreateDirectory(@"foo\bar");
    var dirs1 = Directory.GetDirectories("foo");
    CollectionAssert.AreEqual(new[] { @"foo\bar" }, dirs1); //passes

    //This shows that the real FileSystem behaves correctly, same as system classes
    var fs = new FileSystem();
    var dirs2 = fs.Directory.GetDirectories("foo");
    CollectionAssert.AreEqual(dirs1, dirs2); //passes

    //This shows that the MockFileSystem behaves incorrectly : it returns one absolute directory path
    var mockFs = new MockFileSystem();
    mockFs.Directory.CreateDirectory(@"foo\bar");
    var dirs3 = mockFs.Directory.GetDirectories("foo");
    Console.WriteLine(dirs3[0]); // C:\foo\bar
    CollectionAssert.AreEqual(dirs1, dirs3, "dirs3 should equal dirs1"); //fails
}
```

tuan-tu-tran:
This should fix #815 

If the path orginally given is relative, we return relative paths as well.
For instance in a folder structure like:
```
.
./foo
./foo/bar
```
where current directory is `/foobar`

Initially, the underlying method returns absolute paths i.e. calling `GetFilesInternal("/foobar/foo")` returns `/foobar/foo/bar`.
We then compute the last component of the full path i.e. `bar`, which we combine with the path originally requested:
- `foo -> foo/bar`
- `./foo -> ./foo/bar`
- `../foobar/foo -> ../foobar/foo/bar`

This works better than trying to get the absolute path relative to the current directory

Also, we add a new relevant test and adapt an existing incorrect test: 
It verifies that the returned paths are absolute, even though the path given is relative:
[MockDirectory_GetDirectories_RelativeDirectory_WithChildren_ShouldReturnChildDirectories(string relativeDirPath)](https://github.com/TestableIO/System.IO.Abstractions/blob/ced2f2b8e2a251cd87a0ee1049d19c947d0edfe3/tests/System.IO.Abstractions.TestingHelpers.Tests/MockDirectoryTests.cs#L1214)
That test was added by #258 for #169

### DIFF
--- a/src/System.IO.Abstractions.TestingHelpers/MockDirectory.cs
+++ b/src/System.IO.Abstractions.TestingHelpers/MockDirectory.cs
@@ -567,11 +567,13 @@ namespace System.IO.Abstractions.TestingHelpers
             return EnumerateDirectories(path, searchPattern, SearchOption.TopDirectoryOnly);
         }
 
+
         /// <inheritdoc />
         public override IEnumerable<string> EnumerateDirectories(string path, string searchPattern, SearchOption searchOption)
         {
             mockFileDataAccessor.PathVerifier.IsLegalAbsoluteOrRelative(path, "path");
             var isOriginalPathRooted = mockFileDataAccessor.Path.IsPathRooted(path);
+            string originalPath = path;
             path = path.TrimSlashes();
             path = mockFileDataAccessor.Path.GetFullPath(path);
             var directories = GetFilesInternal(mockFileDataAccessor.AllDirectories, path, searchPattern, searchOption)
@@ -579,7 +581,10 @@ namespace System.IO.Abstractions.TestingHelpers
             if (!isOriginalPathRooted)
             {
                 //if the path was orginally relative, return relative paths: #815
-                directories = directories.Select(p => p.Substring(currentDirectory.Length + 1));
+                directories = directories
+                    .Select(p => p.Split(Path.DirectorySeparatorChar).Last()) //first get the last component of the path
+                    .Select(p => Path.Combine(originalPath, p)) //we combine the last component with the original base path
+                ;
             }
             return directories;
         }
--- a/src/System.IO.Abstractions.TestingHelpers/MockDirectory.cs
+++ b/src/System.IO.Abstractions.TestingHelpers/MockDirectory.cs
@@ -571,10 +571,17 @@ namespace System.IO.Abstractions.TestingHelpers
         public override IEnumerable<string> EnumerateDirectories(string path, string searchPattern, SearchOption searchOption)
         {
             mockFileDataAccessor.PathVerifier.IsLegalAbsoluteOrRelative(path, "path");
+            var isOriginalPathRooted = mockFileDataAccessor.Path.IsPathRooted(path);
             path = path.TrimSlashes();
             path = mockFileDataAccessor.Path.GetFullPath(path);
-            return GetFilesInternal(mockFileDataAccessor.AllDirectories, path, searchPattern, searchOption)
+            var directories = GetFilesInternal(mockFileDataAccessor.AllDirectories, path, searchPattern, searchOption)
                 .Where(p => !mockFileDataAccessor.StringOperations.Equals(p, path));
+            if (!isOriginalPathRooted)
+            {
+                //if the path was orginally relative, return relative paths: #815
+                directories = directories.Select(p => p.Substring(currentDirectory.Length + 1));
+            }
+            return directories;
         }
 
 #if FEATURE_ENUMERATION_OPTIONS
--- a/src/System.IO.Abstractions.TestingHelpers/MockDirectory.cs
+++ b/src/System.IO.Abstractions.TestingHelpers/MockDirectory.cs
@@ -567,7 +567,6 @@ namespace System.IO.Abstractions.TestingHelpers
             return EnumerateDirectories(path, searchPattern, SearchOption.TopDirectoryOnly);
         }
 
-
         /// <inheritdoc />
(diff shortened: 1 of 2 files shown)


tuan-tu-tran:
Seriously, please check the way I built back the relative paths: I used a substring but if you see a better way, please let me know.

src/System.IO.Abstractions.TestingHelpers/MockDirectory.cs
```diff
@@ -571,10 +571,17 @@ public override IEnumerable<string> EnumerateDirectories(string path, string sea
...
+            var isOriginalPathRooted = mockFileDataAccessor.Path.IsPathRooted(path);
             path = path.TrimSlashes();
             path = mockFileDataAccessor.Path.GetFullPath(path);
-            return GetFilesInternal(mockFileDataAccessor.AllDirectories, path, searchPattern, searchOption)
+            var directories = GetFilesInternal(mockFileDataAccessor.AllDirectories, path, searchPattern, searchOption)
                 .Where(p => !mockFileDataAccessor.StringOperations.Equals(p, path));
+            if (!isOriginalPathRooted)
+            {
+                //if the path was orginally relative, return relative paths: #815
+                directories = directories.Select(p => p.Substring(currentDirectory.Length + 1));
```
tuan-tu-tran, line 582: To rebuild the relative path I tried using `mockFileDataAccessor.Path.GetRelativePath` but it's not available in all frameworks. So I used a substring.

fgreinacher:
Sorry @tuan-tu-tran, I forgot about this PR. Will have a look by next week!

tuan-tu-tran:
Sure! No worries!

tests/System.IO.Abstractions.TestingHelpers.Tests/MockDirectoryTests.cs
```diff
@@ -1244,6 +1244,26 @@ public void MockDirectory_GetDirectories_AbsoluteWithNoSubDirectories_ShouldRetu
             Assert.That(actualResult, Is.Empty);
         }
 
+        [TestCase(@"Folder\SubFolder")]
```
fgreinacher, line 1247: I think we're still missing some test cases here. I just played around with a tiny test program:

```csharp

var directories = Directory.GetDirectories(args[0], args[1]);
foreach (var directory in directories)
{
    Console.WriteLine(directory);
}
```

Here are some scenarios that I guess don't work with the current approach:

image

tuan-tu-tran: I've added the test cases but in the test about relative paths above

src/System.IO.Abstractions.TestingHelpers/MockDirectory.cs
tuan-tu-tran: I've re-implemented this part by using `Path.Combine` with the original path argument and the "last path component"
I think using a logic similar to `GetRelativePath` won't do the trick.

For instance, if we're in `/foo` where there's one subfolder `/foo/bar`
If we ask to enumerate `.`, the internal enumeration will return `/foo/bar` and getting the path relative to current folder will simply return `bar` not `./bar`.
Same if we ask `../foo`, using relative path would return `bar` and not `../foo/bar`

Basically I take the last part component `bar` which I recombine with whatever was initially given.

tests/System.IO.Abstractions.TestingHelpers.Tests/MockDirectoryTests.cs
```diff
@@ -1224,7 +1228,7 @@ public void MockDirectory_GetDirectories_RelativeDirectory_WithChildren_ShouldRe
 
             // Assert
             CollectionAssert.AreEqual(
-                new[] { XFS.Path(currentDirectory + @"\" + relativeDirPath + @"\child") },
+                new[] { XFS.Path(relativeDirPath + @"\child") },
```
fgreinacher, line 1231: I think the current implementation does not work for deeper directory structures, e.g. if you have a `child/grandchild` directory here. Could you please add some tests for such scenarios?

src/System.IO.Abstractions.TestingHelpers/MockDirectory.cs
fgreinacher: I don't believe that this is the generally correct approach, see my comment in the test.

I'm wondering whether it would be easier to only fix this for target frameworks where Path.GetRelativePath is available (we are multi-targeting here and could use conditional compilation)

fgreinacher:
https://github.com/TestableIO/System.IO.Abstractions/issues/815 has been resolved, so I assume we can close this. Feel free to reopen if needed.

PR #828, opened 2022-03-22
fix: MockFileInfo.Exists no longer returns stale cached data

### LINKED ISSUE #822: MockFileSystem created file don't exist
**Describe the bug**
From version 16.0.8 of **System.IO.Abstractions.TestingHelpers** when I create new file in **MockFileSystem** property **Exist** return false.

**To Reproduce**
```cs
IFileSystem fileSystem = new MockFileSystem();
IFileInfo fileInfo = fileSystem.FileInfo.FromFileName("testFile.txt");
using (Stream fs = fileInfo.Create()) 
{
    fs.WriteByte(0);
}
Console.WriteLine($"File exist: {fileInfo.Exists}");
```
Version 16.0.7 return **True**
Version 16.0.8 return **False**

**Expected behavior**
Property Exist should return **True**

siprbaum:
In 5216e0d ("fix: make Mock{File,Directory}Info cache file attributes
(#791)", 2022-01-09) a cache of the MockFileData was introduced, to mimic
the behaviour that a FileInfo object will not read the data from disk
but instead also remembers its internal state, which can be updated by
calling its `Refresh` method. This internal (or cached state) is
invalidated when calling the methods `Create`, `Delete` and `MoveTo`
methods, as can be seen by the `Invalidate` calls in
https://github.com/dotnet/runtime/blob/8766a1cf1ed6c3b69cef50154139699b72fb52c5/src/libraries/System.Private.CoreLib/src/System/IO/FileInfo.cs

In the PR #791 it was overlooked to also invalidate the cached data
in these methods. This commit fixes it by invalidating the cached data
only for `Create`, `CreateText` and `Delete`.

The two `MoveTo` variants are not changed as I could not manage to find
any behaviour which is broken without that invalidation.
Most likely reason is that the  MockFileData is updated to represent the
destination file ofter the move, so invalidating is not needed.

Fixes #822

### DIFF
--- a/src/System.IO.Abstractions.TestingHelpers/MockFileInfo.cs
+++ b/src/System.IO.Abstractions.TestingHelpers/MockFileInfo.cs
@@ -28,6 +28,7 @@ namespace System.IO.Abstractions.TestingHelpers
         /// <inheritdoc />
         public override void Delete()
         {
+            refreshOnNextRead = true;
             mockFile.Delete(path);
         }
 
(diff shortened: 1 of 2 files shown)


fgreinacher:
Thanks a ton 👍

rcdailey:
This same issue exists for `IDirectoryInfo`. I'm going to open a pull request with similar changes for that class.

PR #834, opened 2022-04-06
fix: correct MockFileInfo  and MockDirectoryInfo for non existing files/directories

### LINKED ISSUE #829: not existing MockFileInfo Attributes
I am not sure if that is a bug because I don't know how much this library aims to simulate behaviours of real file system so I wanted to ask about attibutes. 
In microsoft [docs](https://docs.microsoft.com/en-us/dotnet/api/system.io.filesysteminfo.attributes?view=net-6.0#remarks) it is said that if file does not exist then attributes is set to `(FileAttributes)(-1)`. [FileNotFoundException](https://docs.microsoft.com/en-us/dotnet/api/system.io.filenotfoundexception?view=net-6.0) or [DirectoryNotFoundException](https://docs.microsoft.com/en-us/dotnet/api/system.io.directorynotfoundexception?view=net-6.0) can only be thrown when setting the value.

But this library is throwing `FileNotFoundException` on reading attributes instead of returning `-1` value.

1. Mock FileSystem with TestHelpers
2. use `fs.FileInfo.FromFileName(@"x:\not-existing\path")`
3. Try to check `Attributes` property

**Expected behavior**
Value `-1` should be returned

### LINKED ISSUE #830: IFileInfo.Attributes are set for unexpected files
**Describe the bug**
This happens with the `MockFileSystem`.
Setting a file to `ReadOnly` also sets another file to `ReadOnly`.

**To Reproduce**
Add following test to `MockFileInfoTests.cs` and execute.
```csharp
        [Test]
        public void MockFileInfo_Attributes_ShouldSetAttributesOnlyForExpectedFile()
        {
            // Arrange
            var firstFilePath = XFS.Path(@"A:\file_01.txt");
            var secondFilePath = XFS.Path(@"A:\file_02.txt");

            fileSystem.AddFile(firstFilePath, MockFileData.NullObject);
            fileSystem.AddFile(secondFilePath, MockFileData.NullObject);

            var firstFileInfo = new MockFileInfo(fileSystem, firstFilePath);
            var secondFileInfo = new MockFileInfo(fileSystem, secondFilePath);

            //ensure the obvious
            //Assert.True(fileSystem.FileInfo.FromFileName(secondFilePath) is MockFileInfo);

            Assert.False(fileSystem.FileInfo.FromFileName(firstFilePath).Attributes.HasFlag(System.IO.FileAttributes.ReadOnly));
            Assert.False(fileSystem.FileInfo.FromFileName(secondFilePath).Attributes.HasFlag(System.IO.FileAttributes.ReadOnly));

            Assert.False(firstFileInfo.Attributes.HasFlag(System.IO.FileAttributes.ReadOnly));
            Assert.False(secondFileInfo.Attributes.HasFlag(System.IO.FileAttributes.ReadOnly));

            // Act
            fileSystem.FileInfo.FromFileName(secondFilePath).Attributes = System.IO.FileAttributes.ReadOnly;
            //secondFileInfo.Attributes = System.IO.FileAttributes.ReadOnly; //as expected sets attributes for instance only, not for the fileSystem

            // Assert
            Assert.False(firstFileInfo.Attributes.HasFlag(System.IO.FileAttributes.ReadOnly));
            Assert.False(secondFileInfo.Attributes.HasFlag(System.IO.FileAttributes.ReadOnly));//expected, because different instance

            Assert.False(fileSystem.FileInfo.FromFileName(firstFilePath).Attributes.HasFlag(System.IO.FileAttributes.ReadOnly));//<-- error, should not throw
            Assert.True(fileSystem.FileInfo.FromFileName(secondFilePath).Attributes.HasFlag(System.IO.FileAttributes.ReadOnly));
        }
```

**Expected behavior**
Penultimate line should not throw.

**Additional context**
version tested: 16.1.25

siprbaum:
fixes #829 
closes #830

### DIFF
--- a/version.json
+++ b/version.json
@@ -1,6 +1,6 @@
 {
   "$schema": "https://raw.githubusercontent.com/AArnott/Nerdbank.GitVersioning/master/src/NerdBank.GitVersioning/version.schema.json",
-  "version": "16.1",
+  "version": "17.0",
   "assemblyVersion": {
     "precision": "major"
   },


fgreinacher:
@siprbaum I agree that we should make `MockFileData.NullObject` internal to prevent misunderstandings. The changes look great! Please do the [version bump](https://github.com/TestableIO/System.IO.Abstractions/blob/main/version.json) as part of this PR and rebase against `main`.

siprbaum:
Rebased and added a commit on top which inreases the major version number.
Please tell if this should be squashed into the existing commit, e.g. https://github.com/TestableIO/System.IO.Abstractions/pull/834/commits/c927c6b99a0a1e16b8071b005dd9f97b35df84fe

It is not clear if increasing the version per PR is fine or if we want to do it per commit, as I believe keeping the history
of this series seems like a worthwhile goal, especially as there is a lot of information in the commits!

PR #852, opened 2022-05-20
ci: deploy only fixes and features

fgreinacher:
I'd like to reduce the release noise a bit. There are more advanced options like `semantic-release` but they would require quite some rework. So I went for this simplistic approach 😄


rkm (approved):

0xced:
Thank you for this! 🙏

PR #856, opened 2022-05-28
fix: MockFile - clone the bytes in read/write allbytes methods

jamesmoore:
### What is this doing?
* When reading from the MockFile's content with the ReadAllBytes method return a copy of the content, not a reference to the content itself.
* When writing to the MockFile content update it with a copy of the new content, not a reference to the new content.
### Why are we doing this?
These Read/WriteAllBytes should be result in two instances of the file content - one on the mock filesystem and the in-memory instance.
### How was it tested?
Unit tests to ensure there are two independent instances of the content after each operation.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,7 +84,13 @@ jobs:
           path: packages/*.*
   deploy:
     name: Deploy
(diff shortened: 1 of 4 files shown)


siprbaum (approved):
Looks good to me.

fgreinacher:
Thanks @jamesmoore 👍

PR #860, opened 2022-06-29
feat: Add support for default values on stream creation

Birnsen:
For some tests we need the current `Datetime` and normal `FileAttributes` in order for the system under tests to work properly. This PR adds a new property to the `MockFileSystem` which initializes the `MockFileData` with the current `Datetime` and `FileAttributes` if needed.

### DIFF
--- a/src/System.IO.Abstractions.TestingHelpers/MockFileData.cs
+++ b/src/System.IO.Abstractions.TestingHelpers/MockFileData.cs
@@ -27,6 +27,14 @@ namespace System.IO.Abstractions.TestingHelpers
             Attributes = (FileAttributes)(-1),
         };
 
+        internal static readonly MockFileData NewObject = new MockFileData(string.Empty)
+        {
+            CreationTime = DateTime.Now,
+            LastWriteTime = DateTime.Now,
+            LastAccessTime = DateTime.Now,
+            Attributes = FileAttributes.Normal,
+        };
+
         /// <summary>
         /// Gets the default date time offset.
         /// E.g. for not existing files.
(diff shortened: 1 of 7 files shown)


src/System.IO.Abstractions.TestingHelpers/MockFileData.cs
```diff
@@ -27,6 +27,14 @@ public class MockFileData
             Attributes = (FileAttributes)(-1),
         };
 
+        internal static readonly MockFileData NewObject = new MockFileData(string.Empty)
+        {
+            CreationTime = DateTime.Now,
+            LastWriteTime = DateTime.Now,
+            LastAccessTime = DateTime.Now,
+            Attributes = FileAttributes.Normal,
```
fgreinacher, line 35: That already works now, doesn't it?

Birnsen: Yes you are right! It looks like I had a false assumption based on a test I made. If you do it this way `Attributes` is indead -1
```
var fi = fileSystem.FileInfo.FromFileName(file); //File does not exist
using var stream = fi.Create(); //File exists
var attr = fileSystem.File.GetAttributes(file); //FileAttributes.Normal
var attr2 = fi.Attributes; // -1
```
which is correct I assume. With a `FileInfo.Refresh()` before accessing the `Attributes` it works as expected.

Birnsen:
I realized that this "solution" is not the best one and look for a better one, where the timestamps are set at the proper locations.

Thanks for your time. This PR can be closed.

PR #861, opened 2022-06-29
Add adjusted timestamps on copied MockFiles

xtopaz:
For tests with MockFiles we would like to have the same behaviour as on a real 
file system regarding the timestamps on the copied file.

### DIFF
--- a/src/System.IO.Abstractions.TestingHelpers/MockFile.cs
+++ b/src/System.IO.Abstractions.TestingHelpers/MockFile.cs
@@ -125,7 +125,9 @@ namespace System.IO.Abstractions.TestingHelpers
 
             var sourceFileData = mockFileDataAccessor.GetFile(sourceFileName);
             sourceFileData.CheckFileAccess(sourceFileName, FileAccess.Read);
-            mockFileDataAccessor.AddFile(destFileName, new MockFileData(sourceFileData));
+            var destFileData = new MockFileData(sourceFileData);
+            destFileData.CreationTime = destFileData.LastAccessTime = DateTime.Now;
+            mockFileDataAccessor.AddFile(destFileName, destFileData);
         }
 
         /// <inheritdoc />
--- a/tests/System.IO.Abstractions.TestingHelpers.Tests/MockFileCopyTests.cs
+++ b/tests/System.IO.Abstractions.TestingHelpers.Tests/MockFileCopyTests.cs
@@ -28,6 +28,23 @@ namespace System.IO.Abstractions.TestingHelpers.Tests
         }
 
         [Test]
+        public void MockFile_Copy_ShouldAdjustTimestampsOnDestination()
+        {
+            var sourceFileName = XFS.Path(@"c:\source\demo.txt");
+            var destFileName = XFS.Path(@"c:\source\demo_copy.txt");
+
+            var mockFileSystem = new MockFileSystem();
+            mockFileSystem.AddFile(sourceFileName, "Original");
+            mockFileSystem.File.Copy(sourceFileName, destFileName);
+
+            var sourceFileInfo = mockFileSystem.FileInfo.FromFileName(sourceFileName);
+            var destFileInfo = mockFileSystem.FileInfo.FromFileName(destFileName);
+            Assert.AreEqual(sourceFileInfo.LastWriteTime, destFileInfo.LastWriteTime);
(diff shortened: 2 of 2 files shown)


fgreinacher:
This looks very similar to #860

xtopaz:
Unfortunately after my last commit https://github.com/TestableIO/System.IO.Abstractions/pull/861/commits/920527b3e6588f70a574052c91775a2b41cc83c7 an issue cropped up in the static code analysis which claims that "'existingContents' is null on at least one execution path." in src/System.IO.Abstractions.TestingHelpers/MockFileStream.cs#L44. 
@fgreinacher, since this is obviously a false positive, could you please set it to ignore?

src/System.IO.Abstractions.TestingHelpers/MockFileStream.cs
```diff
@@ -31,7 +31,8 @@ public class MockFileStream : MemoryStream
                     throw CommonExceptions.FileAlreadyExists(path);
                 }
 
-                var fileData = mockFileDataAccessor.GetFile(path);
+                fileData = mockFileDataAccessor.GetFile(path);
+                fileData.LastAccessTime = DateTime.Now;
                 fileData.CheckFileAccess(path, access);
```
siprbaum, line 36: This change caught my eye, as it changes the LastAccessTime before the check if we even can access the file.
Have you checked what the real filesystem does in case of the code path covered in `fileData.CheckFileAccess(...)` ?

To make it obvious, it is not clear to me if `fileData.LastAccessTime = DateTime.Now;` should becomme before or after the line `fileData.CheckFileAccess(path, access);`

src/System.IO.Abstractions.TestingHelpers/MockFileStream.cs
```diff
@@ -71,6 +74,20 @@ public class MockFileStream : MemoryStream
...
+            return base.Read(buffer, offset, count);
+        }
+
+        /// <inheritdoc />
+        public override void Write(byte[] buffer, int offset, int count)
+        {
+            fileData.LastWriteTime = fileData.LastAccessTime = DateTime.Now;
+            base.Write(buffer, offset, count);
+        }
+
```
siprbaum, line 90: I have to admit that right now I do not know if the real FileStream implementation is updating the LastAccessTime or
the LastWriteTime on any Read/Write. I could also imagine that it kind of bulk reads / writes and only on `Flush` it will
do the updates.

But as I said, I have not checked. Some tests or comments or links to the documentation from you would help to clarify that change.

siprbaum (commented):
See inline comments, mainly clarifications needed.

src/System.IO.Abstractions.TestingHelpers/MockFileStream.cs
xtopaz: Thank you for your thoughtful look! 
Obviously the check `fileData.CheckFileAccess(path, access);` should be done before a timestamp is updated. I also realised that the LastAccessTime timestamp should not be updated in the constructor, but only when the file is actually accessed. 

This has been fixed in d5f5d1e

src/System.IO.Abstractions.TestingHelpers/MockFileStream.cs
xtopaz: In fact, I too have not examined the `FileStream` implementation in that regard.  However, I have noticed that the timestamps are not always updated immediately after the respective operation, but at the latest when the OS has processed the flush.

In contrast, with the `MockFileStream` implementation, I made the assumption for simplicity that the IO operations are executed immediately and the data is persisted instantaneously.

fgreinacher: Thanks for checking! I agree that we can go for the simple solution here. I think there is no exact guarantee of when these timestamps are set even in the real FS (depends on OS).

fgreinacher:
> Unfortunately after my last commit [920527b](https://github.com/TestableIO/System.IO.Abstractions/commit/920527b3e6588f70a574052c91775a2b41cc83c7) an issue cropped up in the static code analysis which claims that "'existingContents' is null on at least one execution path." in src/System.IO.Abstractions.TestingHelpers/MockFileStream.cs#L44. @fgreinacher, since this is obviously a false positive, could you please set it to ignore?

Sure, done! Thanks for having a look at this!

fgreinacher:
Thanks a lot for your contribution @xtopaz.

Also thanks @siprbaum for helping with the review!

PR #862, opened 2022-06-30
fix: improve cache invalidation in MockDirectoryInfo

rcdailey:
When invoking the `Create()` and `Delete()` methods on
`MockDirectoryInfo`, the `Exists` property should intrinsically invoke
`Refresh()` the next time it is called. This is necessary to prevent
returning stale state about the directory after those operations.

Additional test cases have been added for `MoveTo()` as well, to ensure
it also not returning stale state.

For context, this change was motivated by PR #828 where similar issues
were solved in `MockFileInfo`.

### DIFF
--- a/tests/System.IO.Abstractions.TestingHelpers.Tests/MockDirectoryInfoTests.cs
+++ b/tests/System.IO.Abstractions.TestingHelpers.Tests/MockDirectoryInfoTests.cs
@@ -1,5 +1,6 @@
 ﻿using System.Collections.Generic;
 using System.Linq;
+using System.Security.AccessControl;
 using NUnit.Framework;
 
 namespace System.IO.Abstractions.TestingHelpers.Tests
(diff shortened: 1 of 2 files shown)


rcdailey:
I'm not going to add the conventional commit prefix as I prefer the repository maintainers to squash-merge my PR and apply the appropriate convention. This is not out of laziness, but rather out of respect: The maintainers will be more familiar with how they wish to deliver my changes.

tests/System.IO.Abstractions.TestingHelpers.Tests/MockDirectoryInfoTests.cs
```diff
@@ -444,6 +445,76 @@ public void MockDirectoryInfo_Exists_ShouldUpdateCachedDataOnRefresh()
...
+
+            // Act
+            directoryInfo.Create();
+
+            // Assert
+            Assert.IsTrue(directoryInfo.Exists);
+        }
+
+        [Test]
+        public void Directory_exists_after_creation_with_security()
```
fgreinacher, line 463: This can only run on Windows. See https://github.com/TestableIO/System.IO.Abstractions/blob/d7ba4dc9a47ec054730aba16fb1995ad77809cea/tests/System.IO.Abstractions.TestingHelpers.Tests/MockDirectoryInfoAccessControlTests.cs#L10 for how we handle this 🙃

fgreinacher (changes requested):
Thanks a lot @rcdailey! We have a failing test on non-Windows that needs adaptation, otherwise good to go!

tests/System.IO.Abstractions.TestingHelpers.Tests/MockDirectoryInfoTests.cs
rcdailey: I'm confused. Are `MockDirectoryInfo` / `MockFileInfo` not platform-agnostic?

fgreinacher: The actual "magic" is [`WindowsOnlyAttribute`](https://github.com/TestableIO/System.IO.Abstractions/blob/d7ba4dc9a47ec054730aba16fb1995ad77809cea/tests/System.IO.Abstractions.TestingHelpers.Tests/WindowsOnlyAttribute.cs) which prevents the test/fixture they it is applied to from running on non-Windows operating systems. `WindowsSpecifics.AccessControlLists` is just a marker to declare (for us humans :smile:) why this specific test is Windows-only.

rcdailey:
@fgreinacher Let me know if there's anything else I need to do. Thank you once again for taking the time to review my PR.

fgreinacher (approved):
Thanks for your work on this!

Issue #849, opened 2022-05-18
Recursively delete a directory with readonly file results in inconsistent state


vbreuss:
**Describe the bug**
Recursively deleting a directory with read only file results in an inconsistent state of the `MockFileSystem`:
- The deletion results (as expected) in an `UnauthorizedAccessException`
- The file is still present (`File.Exists` returns `true`)
- The directory however is removed (`Directory.Exists` returns `false`)

**To Reproduce**
Consider the following unit test (using Xunit and FluentAssertions):
```csharp
    [Fact]
    public void DeleteDirectoryRecursive_WithReadOnlyFile_ShouldThrowExceptionAndNotDeleteDirectory()
    {
      var sut = new MockFileSystem();
      sut.Directory.CreateDirectory("foo");
      sut.File.WriteAllText("foo/bar.txt", "xyz");
      sut.File.SetAttributes("foo/bar.txt", FileAttributes.ReadOnly);
      
      var exception = Record.Exception(() =>
      {
        sut.Directory.Delete("foo", true);
      });

      exception.Should().BeAssignableTo<UnauthorizedAccessException>();
      sut.File.Exists("foo/bar.txt").Should().BeTrue();
      sut.Directory.Exists("foo").Should().BeTrue();
    }
```
This test fails at the last line, as the directory is deleted, although the file still exists, resulting in an inconsistent state of the `MockFileSystem`

**Expected behavior**
The unit test should be successful.

fgreinacher:
Thanks for reporting! This would probably needs to be fixed in https://github.com/TestableIO/System.IO.Abstractions/blob/main/src/System.IO.Abstractions.TestingHelpers/MockDirectory.cs#L118.

VictorLlanir:
Hello, @fgreinacher, I would like to work on this issue.

fgreinacher:
> Hello, @fgreinacher, I would like to work on this issue.

Thanks!

VictorLlanir:
You're welcome, @fgreinacher! I've already opened a PR (#863) for this issue.

PR #863, opened 2022-07-03
#849 - Implementing directory validation for ReadOnly files

VictorLlanir:
#849 - Implementing directory validation for ReadOnly files when trying to delete directory recursively

On the RemoveFile method, there was no validation to check if there is a ReadOnly file inside the directory berfore trying to remove. This PR implements this validation.

### DIFF
--- a/src/System.IO.Abstractions.TestingHelpers/MockFileSystem.cs
+++ b/src/System.IO.Abstractions.TestingHelpers/MockFileSystem.cs
@@ -176,8 +176,7 @@ namespace System.IO.Abstractions.TestingHelpers
 
             lock (files)
             {
-                if (FileExists(fixedPath) &&
-                    (GetFile(fixedPath).Attributes & FileAttributes.ReadOnly) == FileAttributes.ReadOnly)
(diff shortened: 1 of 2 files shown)


src/System.IO.Abstractions.TestingHelpers/MockFileSystem.cs
```diff
@@ -396,6 +403,14 @@ private bool DirectoryExistsWithoutFixingPath(string path)
             }
         }
 
+        private bool FileIsReadOnly(string path)
+        {
+            lock (files)
```
fgreinacher, line 408: Can we get of the `lock` here? Looks unnecessary to me because we already lock at the call site.

fgreinacher (commented):
Thanks a lot @VictorLlanir, I have a small question/suggestion, let me know what you think.

VictorLlanir:
You're right, @fgreinacher. I'm going to fix this and update this PR.

fgreinacher (approved):
Thanks, looking great now!

PR #864, opened 2022-07-06
fix: #736 - MoveTo implementation for MockDirectoryInfo under TestingHelpers

### LINKED ISSUE #736: Using the MoveTo function to move a mocked directory doesn't update the DirectoryInfo information
**Describe the bug**
In a unit test which consists on moving directory using the MoveTo function from IDirectoryInfo doesn't update the IDirectoryInfo entries and the moved directory info keeps the same initial values. However using the same function on the debug and release modes( FileSystem) does update the fields and properties of the DirectorytInfo

**To Reproduce**
1. Mock a directory 
2. Initialize DirectoryInfo
3. Move the directory to a new destination
4. Compare the path of the current directory and the destination path.

**Expected behavior**
The moved mocked directory is updated to new state (new path).

afroze9:
Fix for issue #736.
Made changes to the `MockDirectoryInfo` under `System.IO.Abstractions.TestingHelpers`.
`MoveTo` now updates the `directoryName` when moving the directory to the destination folder.

This should also fix the following methods (if they are called after `MoveTo` is called):
* Delete
* GetAccessControl
* GetDirectories
* SetAccessControl
* GetMockFileDataForWrite

### DIFF
--- a/src/System.IO.Abstractions.TestingHelpers/MockDirectoryInfo.cs
+++ b/src/System.IO.Abstractions.TestingHelpers/MockDirectoryInfo.cs
@@ -11,7 +11,7 @@ namespace System.IO.Abstractions.TestingHelpers
     public class MockDirectoryInfo : DirectoryInfoBase
     {
         private readonly IMockFileDataAccessor mockFileDataAccessor;
-        private readonly string directoryPath;
+        private string directoryPath;
         private readonly string originalPath;
         private MockFileData cachedMockFileData;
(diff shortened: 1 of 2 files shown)


fgreinacher (changes requested):
Thanks a lot for your contribution @afroze9! Please have a look at the failing tests and ping me once they are fixed 🙇

fgreinacher:
Superseded by #972

/close

Issue #872, opened 2022-08-09
Set Creation, LastAccess and LastWrite time for files


vbreuss:
**Is your feature request related to a problem? Please describe.**
I want to test a functionality that relies on the "GetLastWriteTimeUtc" method to detect when a file was last changed.
This property in `MockFileSystem` only returns a fixed value "2010-02-03 20:00:00"

**Describe the solution you'd like**
`MockFileSystem` should set the creation, last access and last write times and return the correct values.
Ideally you would also provide an optional callback or property to set, so that I can even mock the time that is stored.

**Describe alternatives you've considered**
None, as the times are not even tracked anywhere internally.

**Additional context**
I would like e.g. the following unit test to pass:
```csharp
  [Theory]
  [WithAutoData]
  public void GetCreationTimeUtc_ShouldBeSet(string file)
  {
    var now = DateTime.UtcNow;
    var sut = new MockFileSystem();
    sut.File.WriteAllText("foo.txt", "xyz");

    var result = sut.File.GetCreationTimeUtc("foo.txt");

    result.Kind.Should().Be(DateTimeKind.Utc);
    result.Should().BeOnOrAfter(now);
  }
```

fgreinacher:
We already have the partially, see https://github.com/TestableIO/System.IO.Abstractions/pull/861.

Makes sense to implement this consistently for all operations 👍

vbreuss:
I would give it a try...

Issue #839, opened 2022-04-29
Incorrect LastAccessTime & LastWriteTime for directory


CzBuCHi:
**Describe the bug**
Mock file system is not storing LastAccessTime & LastWriteTime properties for directories

**To Reproduce**
```
var fileSystem = new MockFileSystem();
var info = fileSystem.DirectoryInfo.FromDirectoryName(@"C:\\Parent\\Dir");
info.Create();
info.LastWriteTime = new DateTime(2001, 2, 3);

var c= fileSystem.DirectoryInfo.FromDirectoryName("C:\\Parent");
foreach (var info in parent.EnumerateDirectories()) {
   var thisShouldBeTrue = info.LastAccessTime  == new DateTime(2001, 2, 3);
}
```
**Expected behavior**
variable `thisShouldBeTrue` should be true.

**Additional context**
class `MockFileSystem` should have method `AddDirectory(string path, MockDirectoryInfo mock)` so i can set directory attributes same way as in method `AddFile`. `MockDirectoryInfo` is derived from `MockFileInfo` but with no content (only default ctor) and with attribute FileAttributes.Directory.

fgreinacher:
Sounds useful, maybe @siprbaum you also have an opinion here 😀

siprbaum:
First, let me adress thispart:

> **Additional context** class `MockFileSystem` should have method `AddDirectory(string path, MockDirectoryInfo mock)` so i can set directory attributes same way as in method `AddFile`. `MockDirectoryInfo` is derived from `MockFileInfo` but with no content (only default ctor) and with attribute FileAttributes.Directory.

I think you meant `MockDirectoryInfo` -> `MockDirectoryData`  and  `MockFileInfo` ->  `MockFileData`. 
Adding a method 
```
AddDirectory(string path, MockDirectoryData mock)
                          ^^^^ ...Data not ...Info
```
could be useful, I believe.


Concerning the automatic LastAccessTime update when enumerating the  directroy:
> ```
> var fileSystem = new MockFileSystem();
> var info = fileSystem.DirectoryInfo.FromDirectoryName(@"C:\\Parent\\Dir");
> info.Create();
> info.LastWriteTime = new DateTime(2001, 2, 3);
> 
> var c= fileSystem.DirectoryInfo.FromDirectoryName("C:\\Parent");
> foreach (var info in parent.EnumerateDirectories()) {
[...]

PR #875, opened 2022-08-10
feat: set Creation, LastAccess and LastWrite time for files

vbreuss:
Implements #872 
Adjust `CreationTime`, `LastAccessTime` and `LastWriteTime` when interacting with files in the `MockFileSystem`.

Provides a means to mock the used DateTime by calling e.g.
```csharp
var fixedTime = new DateTime(2022, 01, 01);
var fileSystem = new MockFileSystem().MockTime(() => fixedTime);
// All times will now be set to the fixedTime.
```

Implementes the following logic in MockFile:
*All these cases are covered by unit tests in MockFileAdjustTimesTest*
- When creating files
   CreationTime, LastAccessTime and LastWriteTime are set to the current time
- When changing files (`WriteAllText`, `WriteAllBytes`, `AppendAllText`, `OpenWrite`, `Open`)
   LastAccessTime and LastWriteTime are set to the current time
   CreationTime is left unchanged
- When reading files (`ReadAllText`, `ReadAllLines`, `ReadAllBytes`, `OpenRead`)
   LastAccessTime is set to the current time
   CreationTime and LastWriteTime are left unchanged
- When setting attributes or ACLs (`SetAttributes`, `SetAccessControl`)
   LastAccessTime is set to the current time
   CreationTime and LastWriteTime are left unchanged
- When moving files (`Move`)
   LastAccessTime is set to the current time
   CreationTime and LastWriteTime are left unchanged
- When copying files (`Copy`)
   CreationTime and LastAccessTime of the copied file in the destination directory is set to the current time
   LastWriteTime and all times of the source file are left unchanged


vbreuss:
Please note, that I changed the Interface `IMockFileDataAccessor`.
This interface looks like it is intended for internal use, but as it is public this could be considered a breaking change...

I am not sure what I need to do...

vbreuss:
If this is an accepted approach, I could also try to have a look at #839, which deals with the times of directories. This would probably be quite similar...

fgreinacher:
Sorry @vbreuss, I'm a bit busy with other things, will try to have a look by the end of this week!

src/System.IO.Abstractions.TestingHelpers/MockFileSystem.cs
```diff
@@ -124,6 +140,28 @@ private string GetPathWithCorrectDirectoryCapitalization(string fullPath)
             return fullPath.TrimSlashes();
         }
 
+        /// <inheritdoc />
+        public MockFileData AdjustTimes(MockFileData fileData, TimeAdjustments timeAdjustments)
+        {
+            var now = (dateTimeProvider ?? defaultDateTimeProvider)();
```
siprbaum, line 146: Why is this check needed here again? The dateTimeProvider is initialized to the defaultDateTimeProvider, so the below should be fine.
```suggestion
            var now = dateTimeProvider();
```

src/System.IO.Abstractions.TestingHelpers/MockFileSystem.cs
```diff
@@ -124,6 +140,28 @@ private string GetPathWithCorrectDirectoryCapitalization(string fullPath)
...
+            if ((timeAdjustments & TimeAdjustments.CreationTime) != TimeAdjustments.None)
```
siprbaum, line 147: Not wrong, but I'd find the following more idiomatic
```suggestion
            if ((timeAdjustments & TimeAdjustments.CreationTime) != TimeAdjustments.CreationTime)
```

src/System.IO.Abstractions.TestingHelpers/MockFileSystem.cs
```diff
@@ -124,6 +140,28 @@ private string GetPathWithCorrectDirectoryCapitalization(string fullPath)
...
+            {
+                fileData.CreationTime = now;
+            }
+
+            if ((timeAdjustments & TimeAdjustments.LastAccessTime) != TimeAdjustments.None)
```
siprbaum, line 152: Same here
```suggestion
            if ((timeAdjustments & TimeAdjustments.LastAccessTime) != TimeAdjustments.LastAccessTime)
```

src/System.IO.Abstractions.TestingHelpers/MockFileSystem.cs
```diff
@@ -124,6 +140,28 @@ private string GetPathWithCorrectDirectoryCapitalization(string fullPath)
...
+            {
+                fileData.LastAccessTime = now;
+            }
+
+            if ((timeAdjustments & TimeAdjustments.LastWriteTime) != TimeAdjustments.None)
```
siprbaum, line 157: Same here 
```suggestion
            if ((timeAdjustments & TimeAdjustments.LastWriteTime) != TimeAdjustments.LastWriteTime)
```

src/System.IO.Abstractions.TestingHelpers/IMockFileDataAccessor.cs
```diff
@@ -8,6 +8,14 @@ namespace System.IO.Abstractions.TestingHelpers
     /// </summary>
     public interface IMockFileDataAccessor : IFileSystem
     {
+        /// <summary>
+        /// Adjust the times of the <paramref name="fileData"/>.
+        /// </summary>
+        /// <param name="fileData">The <see cref="MockFileData"/> for which the times should be adjusted.</param>
+        /// <param name="timeAdjustments">The adjustments to make on the <see cref="MockFileData"/>.</param>
+        /// <returns>The adjusted file.</returns>
+        MockFileData AdjustTimes(MockFileData fileData, TimeAdjustments timeAdjustments);
```
siprbaum, line 17: The interface is a little confusing. The method in place modifies the given parameter `fileData` and also returns it.
At first when I read some usages of this method further in the changes, I was wondering why sometimes the return value is used and sometimes not. At least this confused me and I needed to check what was going on

src/System.IO.Abstractions.TestingHelpers/MockFileSystem.cs
vbreuss: It is not initialized, when serialized/deserialized. I am not sure how to work around this any other way 😟...

vbreuss: According to [this thread on StackOverflow](https://stackoverflow.com/a/9419943) I could add the following method to initialize the callback:

```csharp
void OnDeserializing(System.Runtime.Serialization.StreamingContext c);
```

I will try this approach together with the other review changes...

src/System.IO.Abstractions.TestingHelpers/MockFileSystem.cs
```diff
@@ -15,6 +15,9 @@ public class MockFileSystem : FileSystemBase, IMockFileDataAccessor
 
         private readonly IDictionary<string, FileSystemEntry> files;
         private readonly PathVerifier pathVerifier;
+        [NonSerialized]
+        private Func<DateTime> dateTimeProvider = defaultDateTimeProvider;
+        private static Func<DateTime> defaultDateTimeProvider = () => DateTime.Now;
```
fgreinacher, line 20: Is it really `DateTime.Now` (vs. `UtcNow`) for the real file system?

src/System.IO.Abstractions.TestingHelpers/MockFileSystem.cs
```diff
@@ -88,6 +91,19 @@ public MockFileSystem(IDictionary<string, MockFileData> files, string currentDir
...
         public PathVerifier PathVerifier => pathVerifier;
 
+        /// <summary>
+        /// Replaces the time provider with a mocked instance. This allows to influence the used time in tests.
+        /// <para />
+        /// If not set, the default implementation returns <see cref="DateTime.Now"/>.
+        /// </summary>
+        /// <param name="dateTimeProvider">The function that returns the current <see cref="DateTime"/>.</param>
+        /// <returns></returns>
+        public MockFileSystem MockTime(Func<DateTime> dateTimeProvider)
```
fgreinacher, line 101: This would introduce a new pattern for altering behavior of the mock file system. Could we maybe just add a new optional constructor parameter to pass the provider?

src/System.IO.Abstractions.TestingHelpers/MockFileSystem.cs
fgreinacher:
```suggestion
            if (timeAdjustments.HasFlag(TimeAdjustments.CreationTime))
```

fgreinacher (commented):
Looks good so far, I made some suggestions.

I'm leaving for summer vacation soon, feel free to finish the review with @siprbaum and ping me for merge!

src/System.IO.Abstractions.TestingHelpers/MockFileSystem.cs
vbreuss: If I add it only as a constructor parameter, I can't change it later on, as I used it in the unit tests.
I could instead make it a public writable property, but don't think, that this is fundamentally a better approach...

src/System.IO.Abstractions.TestingHelpers/MockFileSystem.cs
vbreuss: Just checked and the approach with the `OnDeserializing` method works. Will change it accordingly...

src/System.IO.Abstractions.TestingHelpers/MockFileSystem.cs
vbreuss: Will replace with `HasFlag` in next update

src/System.IO.Abstractions.TestingHelpers/MockFileSystem.cs
vbreuss: Will replace with `HasFlag` in next update

src/System.IO.Abstractions.TestingHelpers/MockFileSystem.cs
vbreuss: Will replace with `HasFlag` in next update

src/System.IO.Abstractions.TestingHelpers/MockFileSystem.cs
vbreuss: It is stored as local time in [MockFileInfo.cs](https://github.com/TestableIO/System.IO.Abstractions/blob/main/src/System.IO.Abstractions.TestingHelpers/MockFileInfo.cs), e.g.:

```csharp
        /// <inheritdoc />
        public override DateTime CreationTime
        {
            get
            {
                var mockFileData = GetMockFileDataForRead();
                return mockFileData.CreationTime.DateTime;
            }
            set
            {
                var mockFileData = GetMockFileDataForWrite();
                mockFileData.CreationTime = value;
            }
        }
```

vbreuss: I adjusted the code to consistently store it in UTC.

During this change I noticed, that some existing unit tests were wrong with regards to the `DateTimeKind` (e.g. in [MockFileInfo_LastWriteTimeUtc_ShouldSetLastWriteTimeUtcOfFileInMemoryFileSystem](https://github.com/vbreuss/System.IO.Abstractions/blob/f1f3eda8477e3a2b5a969cca17b5435065aef4f8/tests/System.IO.Abstractions.TestingHelpers.Tests/MockFileInfoTests.cs#L434) `LastWriteTime` instead of `LastWriteTimeUtc` is read and written)

vbreuss: I think this change might make #879 obsolete?
Or rather the behaviour is, that `DateTimeKind.Unspecified` is now consistently interpreted as local time...

vbreuss:
@siprbaum : I merged the latest changes, so that this branch is up to date.
All comments are either fixed and resolved, or I left a comment explaining, why I implemented it this way. Could you please have a look again at this pull request?

src/System.IO.Abstractions.TestingHelpers/MockFile.cs
```diff
@@ -539,6 +541,12 @@ public override Stream Open(string path, FileMode mode, FileAccess access)
 
             var mockFileData = mockFileDataAccessor.GetFile(path);
             mockFileData.CheckFileAccess(path, access);
+            var timeAdjustments = TimeAdjustments.LastAccessTime;
+            if (access != FileAccess.Read)
```
siprbaum, line 545: Perhaps use the explicit check if we are dealing with any write, e.g. `ReadWrite` or `Write`

```suggestion
            if (access.HasFlag(FileAcess.Write))
```

src/System.IO.Abstractions.TestingHelpers/MockFileData.cs
```diff
@@ -126,17 +126,32 @@ public string TextContents
         /// <summary>
         /// Gets or sets the date and time the <see cref="MockFileData"/> was created.
         /// </summary>
-        public DateTimeOffset CreationTime { get; set; } = new DateTimeOffset(2010, 01, 02, 00, 00, 00, TimeSpan.FromHours(4));
```
siprbaum, line 129: Isn't this a user visible behavior change, now that for this and all the other xxxTime Properties below are no longer initialized?

src/System.IO.Abstractions.TestingHelpers/MockFileStream.cs
```diff
@@ -60,7 +61,8 @@ public class MockFileStream : MemoryStream
                 }
 
                 fileData = new MockFileData(new byte[] { });
-                fileData.CreationTime = fileData.LastWriteTime = fileData.LastAccessTime = DateTime.Now;
+                mockFileDataAccessor.AdjustTimes(fileData,
+                    TimeAdjustments.CreationTime | TimeAdjustments.LastAccessTime);
```
siprbaum, line 65: Is it intentional that we no longer update the `LastWriteTime` ?
I have to admit I haven't checked what the correct behavior is, but just noticed that this is different from what the code did before.

src/System.IO.Abstractions.TestingHelpers/MockFileData.cs
vbreuss: You are right, I should probably initialize them in the the constructor.
But are the "Dummy" times really the desired behavior or should I not rather initialize them to either `DateTime.UtcNow` or `DefaultDateTimeOffset`?

src/System.IO.Abstractions.TestingHelpers/MockFileStream.cs
vbreuss: No, this was not intentional, but correct all the same :-).
However, when I compared the behaviour to the actual behaviour of the real file system, I got some interesting results in the other case (that the file already exists).
The correct logic there would depend both on the `mode` and the `access`:
```csharp
        private TimeAdjustments GetTimeAdjustmentsForFileStreamWhenFileExists(FileMode mode, FileAccess access)
        {
            switch (mode)
            {
                case FileMode.Append:
                case FileMode.CreateNew:
                    if (access != FileAccess.Write)
                    {
                        return TimeAdjustments.LastAccessTime;
                    }
                    return TimeAdjustments.None;
                case FileMode.Create:
                case FileMode.Truncate:
                    if (access != FileAccess.Read)
                    {
                        return TimeAdjustments.LastAccessTime | TimeAdjustments.LastWriteTime;
                    }
                    return TimeAdjustments.LastAccessTime;
                case FileMode.Open:
                case FileMode.OpenOrCreate:
                default:
                    return TimeAdjustments.None;
            }
        }
```

vbreuss: For reference:
This was my system test to compare the behaviour:
Should/Can I include such tests? Would it make sense to include this kind of tests e.g. in a separate test suite?

```csharp
using NUnit.Framework;
using System.Threading;
using System.Linq;

namespace System.IO.Abstractions.TestingHelpers.Tests
{
    [TestFixture]
    public class SystemTests
    {
        public static IEnumerable<object[]> GetSystemTestParameters()
        {
            foreach (var fileMode in Enum.GetValues(typeof(FileMode)).Cast<FileMode>())
            {
                foreach (var fileAccess in Enum.GetValues(typeof(FileAccess)).Cast<FileAccess>())
                {
                    yield return new object[] { true, fileMode, fileAccess };
                    yield return new object[] { false, fileMode, fileAccess };
                }
            }
        }

        [Test, TestCaseSource(nameof(GetSystemTestParameters))]
        public void Stream_Create_ShouldBehaveSameAsRealFileSystem(bool fileExists, FileMode fileMode, FileAccess fileAccess)
        {
            string fileName = @$"c:\test\existing_{fileExists}_{fileMode}_{fileAccess}.txt";
            if (fileExists)
            {
                File.WriteAllText(fileName, "foo");
            }
            else
            {
                File.Delete(fileName);
            }

            var time_before = DateTime.UtcNow.AddDays(-5);
            var time_after = time_before.AddSeconds(1);
            var fs = new MockFileSystem().MockTime(() => time_before);
            fs.Directory.CreateDirectory(@$"c:\test");
            if (fileExists)
            {
                fs.File.WriteAllText(fileName, "foo");
            }

            var realfilesystem_before_accessed = File.GetLastAccessTimeUtc(fileName);
            var realfilesystem_before_written = File.GetLastWriteTimeUtc(fileName);
            var realfilesystem_before_created = File.GetCreationTimeUtc(fileName);
            Thread.Sleep(1000);
            try
            {
                _ = new FileStream(fileName, fileMode, fileAccess);
            }
            catch (Exception ex)
            {
                fs.MockTime(() => time_after);
                Assert.Throws(ex.GetType(), () =>
                {
                    _ = fs.FileStream.Create(fileName, fileMode, fileAccess);
                });
                return;
            }

            var realfilesystem_after_accessed = File.GetLastAccessTimeUtc(fileName);
            var realfilesystem_after_written = File.GetLastWriteTimeUtc(fileName);
            var realfilesystem_after_created = File.GetCreationTimeUtc(fileName);

            bool isRealCreatedChanged = realfilesystem_before_created != realfilesystem_after_created;
            bool isRealAccessedChanged = realfilesystem_before_accessed != realfilesystem_after_accessed;
            bool isRealWrittenChanged = realfilesystem_before_written != realfilesystem_after_written;

            var mockfilesystem_before_accessed = fs.File.GetLastAccessTimeUtc(fileName);
            var mockfilesystem_before_written = fs.File.GetLastWriteTimeUtc(fileName);
            var mockfilesystem_before_created = fs.File.GetCreationTimeUtc(fileName);
            fs.MockTime(() => time_after);
            _ = fs.FileStream.Create(fileName, fileMode, fileAccess);
            var mockfilesystem_after_accessed = fs.File.GetLastAccessTimeUtc(fileName);
            var mockfilesystem_after_written = fs.File.GetLastWriteTimeUtc(fileName);
            var mockfilesystem_after_created = fs.File.GetCreationTimeUtc(fileName);

            if (isRealCreatedChanged)
            {
                Assert.That(mockfilesystem_before_created, Is.Not.EqualTo(mockfilesystem_after_created), message: "Created");
            }
            else
            {
                Assert.That(mockfilesystem_before_created, Is.EqualTo(mockfilesystem_after_created), message: "Created");
            }
            if (isRealAccessedChanged)
            {
                Assert.That(mockfilesystem_before_accessed, Is.Not.EqualTo(mockfilesystem_after_accessed), message: "Accessed");
            }
            else
            {
                Assert.That(mockfilesystem_before_accessed, Is.EqualTo(mockfilesystem_after_accessed), message: "Accessed");
            }
            if (isRealWrittenChanged)
            {
                Assert.That(mockfilesystem_before_written, Is.Not.EqualTo(mockfilesystem_after_written), message: "Written");
            }
            else
            {
                Assert.That(mockfilesystem_before_written, Is.EqualTo(mockfilesystem_after_written), message: "Written");
            }
        }
    }
}
```

I realized, that in the following cases the `MockFileSystem` does not throw any or the correct exception in comparison with the real file system:
image
But for this I will create a separate issue and not fix it with this PR!

src/System.IO.Abstractions.TestingHelpers/IMockFileDataAccessor.cs
vbreuss: Is keeping the interface as is, OK for you, @siprbaum ?

src/System.IO.Abstractions.TestingHelpers/MockFileData.cs
vbreuss: This is only relevant when directly creating the `MockFileData` (when using the methods from the `IFileSystem`, the time is adjusted correctly). Therefore I suggest to change the constructor to initialize them with `DateTime.UtcNow` as this is the same behaviour of the real file system, even if this is a user visible behaviour change (relying on explicit dummy-values seems not a good approach for me in any case):
```csharp
        /// <summary>
        /// Initializes a new instance of the <see cref="MockFileData"/> class with an empty content.
        /// </summary>
        private MockFileData()
        {
            var now = DateTime.UtcNow;
            LastWriteTime = now;
            LastAccessTime = now;
            CreationTime = now;
        }
```

src/System.IO.Abstractions.TestingHelpers/MockFile.cs
vbreuss: Implemented as suggested

vbreuss:
@siprbaum : Thanks for the review!
I implemented the changes and answered your findings.
Can you please have a look again, and resolve the comments, that you agree with?

Regarding the user detectable behaviour change:
As we are now setting the times correctly, the behaviour changes when previously in the tests you would (incorrectly) rely upon unchanged times for the files.
However I am still unsure, if I should increment the version when considering my [previous comment in this pull request](https://github.com/TestableIO/System.IO.Abstractions/pull/875#issuecomment-1212374146), regarding the change in the interface `IMockFileDataAccessor`.
What do you think?

src/System.IO.Abstractions.TestingHelpers/MockFileStream.cs
fgreinacher: Generally I like the idea of having functional parity tests, but I think it would be quite hard to maintain and overlap a lot with the existing unit tests. Also hard for contributors - would we expect them to add unit and system tests for each feature/fix? What's the criteria for when to pick what?

vbreuss: I agree that it would be difficult to maintain. That's why I didn't include it yet, but it helped me during the implementation.
Sometimes this kind of tests might be beneficial but I would keep them separate from unit tests and optional.

src/System.IO.Abstractions.TestingHelpers/MockFileData.cs
siprbaum: Looks reasonable, but as stated elsewhere, the other ctors also need to be handled

src/System.IO.Abstractions.TestingHelpers/MockFileStream.cs
siprbaum: Good arguments. Nevertheless, I fear that this time handling is quite difficult to get right, especially as the documentation is quite lacking when the time values are updated :-(

siprbaum (changes requested):

src/System.IO.Abstractions.TestingHelpers/IMockFileDataAccessor.cs
siprbaum: Yes, it is

src/System.IO.Abstractions.TestingHelpers/MockFileData.cs
```diff
@@ -49,7 +49,10 @@ public class MockFileData
         /// </summary>
         private MockFileData()
         {
-            // empty
+            var now = DateTime.UtcNow;
+            LastWriteTime = now;
+            LastAccessTime = now;
+            CreationTime = now;
```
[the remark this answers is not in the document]
vbreuss: The other constructors call `this()` (if I remember correctly), but will re-check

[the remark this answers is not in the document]
vbreuss: I overlooked one constructor and added the call to `this()` accordingly...

vbreuss:
@fgreinacher : I merged the latest `main` branch and answered or fixed all review comments.
Can you please have a look?

src/System.IO.Abstractions.TestingHelpers/MockDirectoryInfo.cs
```diff
@@ -114,29 +114,29 @@ public override string FullName
         /// <inheritdoc />
         public override DateTime LastAccessTime
         {
-            get { return GetMockFileDataForRead().LastAccessTime.DateTime; }
+            get { return GetMockFileDataForRead().LastAccessTime.LocalDateTime; }
```
fgreinacher, line 117: These changes make #879 obsolete, right?

src/System.IO.Abstractions.TestingHelpers/MockFileSystem.cs
```diff
@@ -1,6 +1,7 @@
-﻿using System.Collections.Generic;
+﻿    using System.Collections.Generic;
```
fgreinacher, line 1:
```suggestion
using System.Collections.Generic;
```

src/System.IO.Abstractions.TestingHelpers/TimeAdjustments.cs
```diff
@@ -0,0 +1,30 @@
...
+        /// <summary>
+        /// Adjusts the <see cref="MockFileData.LastWriteTime"/>
+        /// </summary>
+        LastWriteTime = 1 << 2,
+        /// <summary>
+        /// Adjusts all times on the <see cref="MockFileData"/>
+        /// </summary>
+        All = ~0
+    }
+}
```
fgreinacher, line 30:
```suggestion
}

```

src/System.IO.Abstractions.TestingHelpers/MockDirectoryInfo.cs
vbreuss: Yeah, more or less.
It changes the behaviour, that `DateTimeKind.Unspecified` is now consistently interpreted as local time...

fgreinacher: @xtopaz Can you double check that this changes also solves the issue you have been seeing?

xtopaz: No, the changes unfortunately do not fix the bug described in #879. 
Since in the timestamp setters the value is always implicitly converted to `DateTimeOffset` and thus `DateTimeKind.Unspecified` is treated as local, there is a further time shift here.

vbreuss: OK, now it gets complicated @xtopaz .
Your proposed solution works, when you read/write UTC times (e.g. `LastAccessTimeUtc`), that they remain unchanged when providing `DateTimeKind.Unspecified`.
My solution works, when you read/write the local times (e.g. `LastAccessTime`), that they remain unchanged when providing `DateTimeKind.Unspecified`.
However the real file system works in both cases, that it doesn't change the time, but only adjusts the `DateTimeKind`, so both the following tests against the real file system are successful:

```csharp
        [Test]
        public void RealFileInfo_LastAccessTimeUtcWithUnspecifiedDateTimeKind_ShouldSetLastAccessTimeUtcOfFileInFileSystem()
        {
            var date = DateTime.SpecifyKind(DateTime.Now.AddHours(-4), DateTimeKind.Unspecified);
            var fileSystem = new FileSystem();
            fileSystem.Directory.CreateDirectory(@"c:\test");
            fileSystem.File.WriteAllText(@"c:\test\a.txt", "Demo text content");
            var fileInfo = new FileInfo(@"c:\test\a.txt")
            {
                LastAccessTimeUtc = date
            };

            Assert.AreEqual(date, fileInfo.LastAccessTimeUtc);
            Assert.AreNotEqual(DateTimeKind.Unspecified, fileInfo.LastAccessTimeUtc);
        }
        [Test]
        public void RealFileInfo_LastAccessTimeWithUnspecifiedDateTimeKind_ShouldSetLastAccessTimeOfFileInFileSystem()
        {
            var date = DateTime.SpecifyKind(DateTime.Now.AddHours(-4), DateTimeKind.Unspecified);
            var fileSystem = new FileSystem();
            fileSystem.Directory.CreateDirectory(@"c:\test");
            fileSystem.File.WriteAllText(@"c:\test\a.txt", "Demo text content");
            var fileInfo = new FileInfo(@"c:\test\a.txt")
            {
                LastAccessTime = date
            };

            Assert.AreEqual(date, fileInfo.LastAccessTime);
            Assert.AreNotEqual(DateTimeKind.Unspecified, fileInfo.LastAccessTime);
        }
```

vbreuss: @fgreinacher : I would suggest to close this pull request and fix the behaviour with `DateTimeKind.Unspecified` explicitely with #879 .
I could provide or help with a pull request (based on my changes here), that implements the same behaviour as the real file system in both cases...

xtopaz: The fix implemented in #879 already fixes the behavior in `MockFileInfo`, but `FileInfoWrapper` is not affected by the bug described there.

Please check https://github.com/TestableIO/System.IO.Abstractions/pull/879#discussion_r969527170

fgreinacher: > I would suggest to close this pull request

What about just removing the changes related to timezone handling from this PR? The changes that update the timestamps look good to me?

vbreuss: @fgreinacher:
It would have been harder to remove the timezone handling from this PR, than to also fix #879 here.
I included the tests from @xtopaz from #879 and they ran successful.

fgreinacher: Great, thanks to both of you @xtopaz and @vbreuss 👍

I'll merge this one. I think we don't need #879 anymore then, right?

xtopaz: @fgreinacher Yes, i've closed #879.

DaveakX:
Using v19.1.5 this is working great for files, however it doesn't seem to work for directories  (or I've doing something wrong) as their last modified times are coming back as now rather than my MockTime.

e.g. the following for a file passes
`
var now = DateTime.Now;
var fs = new MockFileSystem().MockTime(() => now.AddMonths(-6));
fs.File.CreateText("/temp/myfile").Close();
var fileinfo = fs.FileInfo.New("/temp/myfile");
Assert.That(fileinfo.LastAccessTimeUtc, Is.LessThan(now));
`

but this, for a directory fails

`
var now = DateTime.Now;
var fs = new MockFileSystem().MockTime(() => now.AddMonths(-6));
var dirinfo = fs.Directory.CreateDirectory("/temp/myfolder");
Assert.That(dirinfo.LastWriteTimeUtc, Is.LessThan(now));
`

PR #878, opened 2022-08-22
feat: Add GetFileSystemEntries with SearchOption

### LINKED ISSUE #775: Directory.GetFileSystemEntries is missing the SearchOption parameter
The windows method for Directory.GetFileSystemEntries supports a parameter SearchOption to return not only values of the top most entries, but also the possibility to search within all sub directories. However, the mock hereby has no option for it yet. 

It would be great to use this method here. Is there a chance, that you add this functionality?

(Function gets used within a .Net Core ASP.NET MVC application)

rkm:
Fixes #775.

### DIFF
--- a/src/System.IO.Abstractions.TestingHelpers/MockDirectory.cs
+++ b/src/System.IO.Abstractions.TestingHelpers/MockDirectory.cs
@@ -653,25 +653,19 @@ namespace System.IO.Abstractions.TestingHelpers
         /// <inheritdoc />
         public override IEnumerable<string> EnumerateFileSystemEntries(string path)
         {
-            var fileSystemEntries = new List<string>(GetFiles(path));
-            fileSystemEntries.AddRange(GetDirectories(path));
-            return fileSystemEntries;
+            return GetFileSystemEntries(path);
         }
 
         /// <inheritdoc />
         public override IEnumerable<string> EnumerateFileSystemEntries(string path, string searchPattern)
         {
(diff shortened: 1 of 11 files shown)


src/System.IO.Abstractions.TestingHelpers/MockDirectory.cs
```diff
@@ -380,6 +380,15 @@ public override string[] GetFileSystemEntries(string path, string searchPattern)
...
 
+        /// <inheritdoc />
+        public override string[] GetFileSystemEntries(string path, string searchPattern, SearchOption searchOption)
+        {
+            var dirs = GetDirectories(path, searchPattern, searchOption);
+            var files = GetFiles(path, searchPattern, searchOption);
+
+            return dirs.Union(files).ToArray();
+        }
+
```
fgreinacher, line 391: We already have a similar implementation in [`EnumerateFileSystemEntries`](https://github.com/TestableIO/System.IO.Abstractions/blob/bb3dde6e2be7acc7a76d22c717b4c554f0aa513e/src/System.IO.Abstractions.TestingHelpers/MockDirectory.cs#L654). What about calling `GetFileSystemEntries` from there for consistency?

fgreinacher (commented):
Thanks for your work on this @rkm!

I left one suggestion, otherwise this looks great!

fgreinacher (approved):
Perfect, thanks again!

PR #879, opened 2022-08-25
fix: set last access time with unspecified date time kind

xtopaz:
This fixes a missing `ToLocalTime` conversion in the `MockFileInfo.LastAccessTimeUtc` setter, which leads to inconsistent results when dealing with `DateTime` values having `DateTimeKind.Unspecified`.

### DIFF
--- a/src/System.IO.Abstractions.TestingHelpers/MockFileInfo.cs
+++ b/src/System.IO.Abstractions.TestingHelpers/MockFileInfo.cs
@@ -137,7 +137,7 @@ namespace System.IO.Abstractions.TestingHelpers
             set
             {
                 var mockFileData = GetMockFileDataForWrite();
-                mockFileData.LastAccessTime = value;
+                mockFileData.LastAccessTime = value.ToLocalTime();
             }
         }
 
--- a/src/System.IO.Abstractions.TestingHelpers/MockFileInfo.cs
+++ b/src/System.IO.Abstractions.TestingHelpers/MockFileInfo.cs
@@ -1,6 +1,5 @@
 ﻿using System.Runtime.Versioning;
 using System.Security.AccessControl;
-using System.Text;
 
 namespace System.IO.Abstractions.TestingHelpers
 {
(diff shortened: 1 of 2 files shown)


src/System.IO.Abstractions.TestingHelpers/MockFileInfo.cs
```diff
@@ -137,7 +137,7 @@ public override DateTime LastAccessTimeUtc
             set
             {
                 var mockFileData = GetMockFileDataForWrite();
-                mockFileData.LastAccessTime = value;
+                mockFileData.LastAccessTime = value.ToLocalTime();
```
fgreinacher, line 140: Should we also do this for the other `*TimeUtc` properties?

fgreinacher: Not sure whether `ToLocalTime` is correct here. I think we rather want something like this: https://github.com/dotnet/runtime/blob/96d1c8e3ba4bc94afdcf15ee2896a3bc984820e7/src/libraries/System.Private.CoreLib/src/System/IO/FileSystemInfo.cs#L108

fgreinacher (changes requested):
Thanks for your work @xtopaz!

I left two comments :)

src/System.IO.Abstractions.TestingHelpers/MockFileInfo.cs
xtopaz: @fgreinacher   Sorry for the late reply, just got back from vacation.

All other `*TimeUtc` properties already contain the `ToLocalTime` call. This way the behavior has been unified.

`File.File.GetUtcDateTimeOffset`cannot be used because it is internal. However, this behaves analogously to the proposed implementation, since it takes Utc for `DateTimeKind.Unspecified`.

xtopaz: @fgreinacher  I've incorporated the recent changes from #875 and all test look fine.

xtopaz:
Obsolete by #875.

Issue #877, opened 2022-08-22
Throws exception on call methods


szymonpiekny:
**Is your feature request related to a problem? Please describe.**
It should be possible to throw exception while call methods on FileSystemMock.

**Describe the solution you'd like**
`_fileSystemMock.Setup(q => q.File.WriteAllText(It.IsAny<string>(), It.IsAny())).Throws<OutOfMemoryException>();`

fgreinacher:
@szymonpiekny You should already be able to this, see https://github.com/TestableIO/System.IO.Abstractions/issues/826#issuecomment-1081702068 for one possibility using `Moq`.

fgreinacher:
I'm closing this. Feel free to reopen if something is unclear!

Issue #805, opened 2022-02-03
Observing changes in the MockFileSystem


SuperJMN:
**Describe the solution you'd like**
I'd like to be able to observe common (interesting from a testing perspective) that happen into a `MockFileSystem`. 

Why? I'm creating a cache service that uses an `IFileSystem` and I need to test that a file is only written once.

**Describe alternatives you've considered**
I've tried to implement my own `MockFileSystem` using System.Reactive, but the work is a bit overwhelming.

fgreinacher:
This is the same as https://github.com/TestableIO/System.IO.Abstractions/issues/168 which has gone stale.

As mentioned over there I generally like the idea and would accept a contribution here unless it's getting overly complex.

Issue #365, opened 2018-10-06
Should the namespace really be System.IO.Abstractions?


bentorkington:
I get that it clearly communicates that these are abstractions for System.IO, but isn't it a bit cheeky to put the library in the System.IO namespace? That might suggest to someone unfamiliar that it is part of .NET.

fgreinacher:
I agree this naming choice was a bit cheeky more than eight years ago, but changing that now would be a severe breaking change for all users, probably causing a lot more trouble than the confusion people might currently encounter.

Additional thoughts @jpreese @tathamoddie?

bentorkington:
True - I agree such a breaking change would be unwelcome.

Incidentally, this was how I learned that `namespace System.IO.Abstractions { ... }` implicitly adds the System.IO namespace. In my code, I removed all my `using System.IO` lines so I didn't inadvertently leave any classes outside the abstraction layer, only to find that I needed it for constants like System.IO.FileMode, etc. anyway.

When browsing the Abstractions code, I saw the library uses these constants without fully qualified names, and without `using System.IO` anywhere.  I haven't been able to find this behaviour of the namespace keyword explicitly documented anywhere.

jpreese:
Yeah, it's a slightly unfortunate name, but I don't think theres enough value return to go about changing it. 

We considered it when Microsoft barred the use of the System.IO namespace a bit ago, making it impossible to deploy package updates to nuget, but thats been since resolved. We were essentially grandfathered in.

bentorkington:
Thanks for the discussion, and such a useful testing library!

Issue #168, opened 2016-08-31
Add events on major events


zvirja:
Sometimes it's needed to test sequence of events with file system and currently it's impossible to do that with current implementation of `MockFileSystem`. Basically, we should add a couple of events, so we could subscribe on that events in our test. Moreover, this way we could simulate various situations (e.g. throw sharing violation exception when particular file is opened).

List of events:   
- File is being opened
- File is being created
- File is being deleted
- File is being updated
- File stream is opened
- File stream is closed
- Directory is being created
- Directory is being deleted

I could implement that and create a PR for you. Will you approve such kind of PR?
Do you see some other ways to achieve the desired behavior?

manne:
@Zvirja which kind of API do you have in your mind? and can you give example(s).

Zvirja:
For instance, that could be a public property exposed directly from the `MockFileSystem`:
```csharp
var fs = new MockFileSystem();
fs.Events.OnFileCreated += (sender, args) => {}
...
```

Basically, nothing special - any way it's convenient to implement. The only requirement is to be able to hook such events :)

Let me know if you want more input from me.

manne:
I think we need an API for the project `TestingHelpers` (MockFileSystem).
Some pull requests/issues indicate the need for this API.
see #189, #177 & #183

ericnewton76:
I wouldnt recommend events on the MockFileSystem at the moment.  

You're basically then implementing a FileSystemWatcher instead.  We should make sure the goal is parity with mscorlib.

manne:
I prefer to have a FileSystemWatcher.
@Zvirja is a FileSystemWatcher sufficient for your purpose?

Zvirja:
I'm not sure that watcher could cover my initial need why I fired this issue. Initially I wanted to have a way to throw exception when file is opened. That will not be achieved by watcher as it doesn't provide us with appropriate event.

Also FileSystemWatcher isn't synchronous by itself (if I'm not mistaken) which also makes it more difficult to  simulate concurrency issues.

rkoeninger:
I outlined an attempt to do this in #373 and it didn't go over well. It was agreed that this would be too involved for this library. I assume the same answer would apply to this issue.

fgreinacher:
In my opinion adding synchronous events like described here would enable quite some scenarios, so I would not reject a PR in this direction.

PR #883, opened 2022-08-31
feat: Support event callbacks and custom exceptions in MockFileSystem

vbreuss:
**Observing Changes**
*Implements #805*
Allows registering a callback to be executed whenever a file or directory event is triggered in the file system:
```csharp
    var fs = new MockFileSystem()
        .OnFileEvent(f => { /* Do something */ })
        .OnDirectoryEvent(d => { /* Do something */ });
```

The callback parameter provides the path of the file or directory and the event type.

**Throw custom exceptions**
*Implements #877*
Additionally the callback allows throwing a custom exception. This callback is executed before the actual change in the  `MockFileSystem`, thus allowing simulating e.g. `OutOfMemoryException` or other edge cases...


fgreinacher:
I like the general direction here 👍

Any reason to not expose this as regular events, as e.g. suggested [here](https://github.com/TestableIO/System.IO.Abstractions/issues/168#issuecomment-263130564)?

> This callback is executed before the actual change

Maybe we could use the *ing/*ed scheme e.g. (`FileCreating`/`FileCreated`) for naming the events to support callbacks both before and after the actual change is made?

vbreuss:
@fgreinacher 
> Any reason to not expose this as regular events, as e.g. suggested https://github.com/TestableIO/System.IO.Abstractions/issues/168#issuecomment-263130564?

It probably boils down to personal preference :-)
I think callbacks are easier and for unit testing I don't see the requirement to support registering multiple event handlers. Also I think it is clearer for callbacks to allow throwing an exception. With events I wouldn't be sure how exceptions are treated, as I could have registered multiple event handlers and only some of them throw exceptions.

> Maybe we could use the *ing/*ed scheme e.g. (FileCreating/FileCreated) for naming the events to support callbacks both before and after the actual change is made?

I think this is a good idea, but I am not sure about the naming.
If I split it into too many different events ("FileCreating", "FileUpdating", "FileDeleting", "FilePermissionChanging", ...) and its counterparts it gets hard to register to all relevant events. That's why I only suggested `OnFileEvent` and `OnDirectoryEvent` and then I can filter on the events I want to handle.

What about `OnFileEventOccuring` and `OnFileEventOccured`?

fgreinacher:
Ah sorry @vbreuss, I got derailed by other topics 🙇 

The main problem I see with the callback approach is that it's not easily possible to unregister the handlers. 

I think we should for something like this:

```csharp
fs.Events.DirectoryCreating += (sender, args) => {}
fs.Events.FileCreated += (sender, args) => {}
...
```

Yes, it's a bit verbose if you want to register the same handler to different events. But on the other hand it's quite intuitive and future-proof (no breaking change when adding new events). 

WDYT?

vbreuss:
@fgreinacher :
I think in the end it boils down to preference.
Especially in unit tests I like the fluent syntax (I also use FluentAssertions in my xUnit tests). Regarding the unregistering, you could return an `IDisposable` which would be quite easy to handle.

Unfortunately, **I won't be able to continue working on this pull request!**
In my application, where I used this library I had quite some inconsistencies, that were not detected by the unit test suite due to incorrect handling of times on different operating systems. Even after my previous contribution here, I stumbled on differences between Linux, Mac and Windows. That was the reason, why I started my own implementation of a broader approach to testing abstractions including a [FileSystemMock](https://github.com/Testably/Testably.Abstractions) where I build an extensive test suite that also runs on a real file system to ensure, that the behaviour matches the real file system.

There I already implemented notification handling as follows:
```csharp
var fileSystem = new FileSystemMock();
fileSystem.Intercept.Creating(FileSystemTypes.Directory, _ => throw new Exception("custom exception"));
fileSystem.Notify.OnCreated(FileSystemTypes.Directory, c => DirectoryWasCreated(c.Path));
```
Maybe a similar approach would also work for you?

fgreinacher:
@vbreuss All good, fully understood! I'm maintaining this at super low capacity and want to keep changes as non-invasive as possible. I personally also like the fluent syntax a lot, but none of the existing APIs use it, and in doubt I also favor consistency.

I like your library a lot! I'm wondering whether it would make sense if you implement the System.IO.Abstractions interfaces instead of duplicating them. In that case I'd be happy to link your lib from the README here, as an alternative, way of mocking things. Let me know what you think 🙇

vbreuss:
@fgreinacher : 
In the beginning I started doing exactly that, however the interfaces are not completely identical:
- I separated the interfaces from the default implementation with wrappers in separate projects. In this project they are both in the main DLL, so I can't just use the interfaces, but would always get the implementations as well.
- You included the ACL-related functionality directly in the interfaces. I extracted them to a separate DLL (Testably.Abstractions.AccessControl), so if they are not required, I don't need a reference to [System.IO.FileSystem.AccessControl](https://www.nuget.org/packages/System.IO.FileSystem.AccessControl/)
- I replaced the `Stream` in the `IFileStreamFactory` with a custom stream class `FileSystemStream` (similar to #793)
- I added and used a `ITimeSystem` interface to simulate the system time, which I need for determining the correct time. This interface is not available in "System.IO.Abstraction" (and wouldn't really fit, as it is not IO-related)
- You didn't replace the `WaitForChangedResult` as return value in the `IFileSystemWatcher` interface. As this struct only has internal constructors, it is not possible to mock. Changing these now, would unfortunately be a breaking change...
- I would also be sceptical of using two completely distinct namespaces ("System.IO.Abstractions" and "Testably.Abstractions.Testing") and adding more projects under the "System.IO.*" namespace would not be possible, if I understand [this comment](https://github.com/TestableIO/System.IO.Abstractions/issues/365#issuecomment-427991975) correctly...

All in all I really like the idea, but my goal with [Testably.Abstractions](https://github.com/Testably/Testably.Abstractions) is for a broader approach to unit testing helpers and I don't see, how we could make these two libraries 100% compatible :-(

Do you have an idea how to overcome these challenges?

fgreinacher:
> @fgreinacher : In the beginning I started doing exactly that, however the interfaces are not completely identical:
> 
> * I separated the interfaces from the default implementation with wrappers in separate projects. In this project they are both in the main DLL, so I can't just use the interfaces, but would always get the implementations as well.

I would have no problem changing that so that the interface are separate.

> * You included the ACL-related functionality directly in the interfaces. I extracted them to a separate DLL (Testably.Abstractions.AccessControl), so if they are not required, I don't need a reference to [System.IO.FileSystem.AccessControl](https://www.nuget.org/packages/System.IO.FileSystem.AccessControl/)

I tried to mirror what .NET does and I think on modern versions we don’t actually need that reference. Or at least it is a no-op. 

> * I replaced the `Stream` in the `IFileStreamFactory` with a custom stream class `FileSystemStream` (similar to [feat: Add `Name` to `IFile.Create` (and others) #793](https://github.com/TestableIO/System.IO.Abstractions/pull/793))

As said the PR I like this, just a matter of getting it merged IMO 😀

> * I added and used a `ITimeSystem` interface to simulate the system time, which I need for determining the correct time. This interface is not available in "System.IO.Abstraction" (and wouldn't really fit, as it is not IO-related)

Agreed, this is not something I’d like to have in SIOA. But is this really a problem at the Interfaxe level?

> * You didn't replace the `WaitForChangedResult` as return value in the `IFileSystemWatcher` interface. As this struct only has internal constructors, it is not possible to mock. Changing these now, would unfortunately be a breaking change...

I would have no problem with changing this. Yes, it’s breaking, but easy to adapt. 

> * I would also be sceptical of using two completely distinct namespaces ("System.IO.Abstractions" and "Testably.Abstractions.Testing") and adding more projects under the "System.IO.*" namespace would not be possible, if I understand [this comment](https://github.com/TestableIO/System.IO.Abstractions/issues/365#issuecomment-427991975) correctly...
> 

Yes, new things need to go with the `TestableIO` prefix. Any option would be fine for me, either you stay on your own or you move to this org. 

All in all I’d say nothing really blocking. Let me know what you think @vbreuss

vbreuss:
@fgreinacher 
> > @fgreinacher : In the beginning I started doing exactly that, however the interfaces are not completely identical:
> > 
> > * I separated the interfaces from the default implementation with wrappers in separate projects. In this project they are both in the main DLL, so I can't just use the interfaces, but would always get the implementations as well.
> 
> I would have no problem changing that so that the interface are separate.

Would this be possible? I think you can't create new projects in the `System.IO.Abstractions` namespace?
Or do you suggest to move the interface to e.g. `TestableIO.Abstractions.Interface` and use them in the `System.IO.Abstractions` library? This would also mean a breaking change to consumers, as they all need to adjust their using-statements...

And I fear this would become quite confusing, especially in my library:
- `TestableIO.Abstractions.Interface`
   Defines the IO-related interfaces
- `System.IO.Abstractrions`
   Contains the default implementation of the IO-related interfaces
- `Testably.Abstractions.Interface`
   Defines the time and random related interfaces
- `Testably.Abstractions`
   Contains the default implementation of the time and random related interfaces
- `Testably.Abstractions.Testing`
   Testing helper

If I were to split my interface DLL, so that one only contains the file-related interfaces (e.g. `Testably.Abstractions.Interface.IO`), would it be possible for you to use these interfaces?
If you think this approach feasible, I could create a draft-PR...

> > * You included the ACL-related functionality directly in the interfaces. I extracted them to a separate DLL (Testably.Abstractions.AccessControl), so if they are not required, I don't need a reference to [System.IO.FileSystem.AccessControl](https://www.nuget.org/packages/System.IO.FileSystem.AccessControl/)
> 
> I tried to mirror what .NET does and I think on modern versions we don’t actually need that reference. Or at least it is a no-op.

In modern versions it was extracted to extension methods. I implemented the same behaviour in [Testably.Abstractions.AccessControl](https://github.com/Testably/Testably.Abstractions/tree/main/Source/Testably.Abstractions.AccessControl). For this to work, I had to implement the [IFileSystemExtensionContainer](https://github.com/Testably/Testably.Abstractions/blob/main/Source/Testably.Abstractions.Interface/IFileSystem.IFileSystemExtensionContainer.cs) interface, but this would be quite a trivial change in "System.IO.Abstractions", so no blocker...

> > * I replaced the `Stream` in the `IFileStreamFactory` with a custom stream class `FileSystemStream` (similar to [feat: Add `Name` to `IFile.Create` (and others) #793](https://github.com/TestableIO/System.IO.Abstractions/pull/793))
> 
> As said the PR I like this, just a matter of getting it merged IMO 😀

Agreed :-)

> > * I added and used a `ITimeSystem` interface to simulate the system time, which I need for determining the correct time. This interface is not available in "System.IO.Abstraction" (and wouldn't really fit, as it is not IO-related)
> 
> Agreed, this is not something I’d like to have in SIOA. But is this really a problem at the Interfaxe level?

Agreed, as said above, I think I could split my interface to only contain the file-related interfaces. However the `ITimeSystem` was quite handy, especially when implementing the "CreationTime", "LastAccessTime" and "LastWriteTime"...

> > * You didn't replace the `WaitForChangedResult` as return value in the `IFileSystemWatcher` interface. As this struct only has internal constructors, it is not possible to mock. Changing these now, would unfortunately be a breaking change...
> 
> I would have no problem with changing this. Yes, it’s breaking, but easy to adapt.
> 
> > * I would also be sceptical of using two completely distinct namespaces ("System.IO.Abstractions" and "Testably.Abstractions.Testing") and adding more projects under the "System.IO.*" namespace would not be possible, if I understand [this comment](https://github.com/TestableIO/System.IO.Abstractions/issues/365#issuecomment-427991975) correctly...
> 
> Yes, new things need to go with the `TestableIO` prefix. Any option would be fine for me, either you stay on your own or you move to this org.
> 
> All in all I’d say nothing really blocking. Let me know what you think @vbreuss

fgreinacher:
> Would this be possible? I think you can't create new projects in the System.IO.Abstractions namespace?
> Or do you suggest to move the interface to e.g. TestableIO.Abstractions.Interface and use them in the System.IO.Abstractions library? This would also mean a breaking change to consumers, as they all need to adjust their using-statements...

True, we cannot upload new NuGet packages with the `System.IO` prefix. But we can keep using the `System.IO` prefix for the namespaces.

I think a simple way would be keep the namespaces as-is and change the NuGet package structure a bit:

- Add `TestableIO.System.IO.Abstractions` with just the interfaces
- Add `TestableIO.System.IO.Abstractions.Wrappers` with the wrapper implementations, referencing `TestableIO.System.IO.Abstractions`
- Keep `System.IO.Abstractions` for backward compatibility, referencing `TestableIO.System.IO.Abstractions.Wrappers`

You could then reference just `TestableIO.System.IO.Abstractions` and plugin in your implementations.

fgreinacher:
> Agreed, as said above, I think I could split my interface to only contain the file-related interfaces. However the ITimeSystem was quite handy, especially when implementing the "CreationTime", "LastAccessTime" and "LastWriteTime"...

Agreed that it helps with the implementation, but I'd like to keep it out of the interfaces and implementation here. But that should be fine, right?

vbreuss:
> I think a simple way would be keep the namespaces as-is and change the NuGet package structure a bit:
> 
> * Add `TestableIO.System.IO.Abstractions` with just the interfaces
> * Add `TestableIO.System.IO.Abstractions.Wrappers` with the wrapper implementations, referencing `TestableIO.System.IO.Abstractions`
> * Keep `System.IO.Abstractions` for backward compatibility, referencing `TestableIO.System.IO.Abstractions.Wrappers`
> 
> You could then reference just `TestableIO.System.IO.Abstractions` from your `Testably.Abstractions.Interface` and plug in in your implementations.

@fgreinacher :
I created #905 to implement the split into the two libraries. Is this how you envisioned the change?
I would suggest to implement any adaptions to the interfaces in a separate PR, so that it is easier for GIT to detect the renamed files and preserve the history. Or would you prefer for me to add the changes into the same PR?
Do you need to change the build pipelines/deployments in any way?

vbreuss:
> I like your library a lot! I'm wondering whether it would make sense if you implement the System.IO.Abstractions interfaces instead of duplicating them. In that case I'd be happy to link your lib from the README here, as an alternative, way of mocking things. Let me know what you think 🙇

I now released [v2.0.0](https://www.nuget.org/packages/Testably.Abstractions.Interface/2.0.0#dependencies-body-tab) which references the interfaces from TestableIO.System.IO.Abstractions (see https://github.com/Testably/Testably.Abstractions/pull/225) and also mentioned the System.IO.Abstractions in the [Readme.md](https://github.com/Testably/Testably.Abstractions#readme)...
What do you think, @fgreinacher ?

Would it make sense to merge the two projects further together?

fgreinacher:
> I now released [v2.0.0](https://www.nuget.org/packages/Testably.Abstractions.Interface/2.0.0#dependencies-body-tab) which references the interfaces from TestableIO.System.IO.Abstractions (see https://github.com/Testably/Testably.Abstractions/pull/225) and also mentioned the System.IO.Abstractions in the [Readme.md](https://github.com/Testably/Testably.Abstractions#readme)...
> What do you think, @fgreinacher ?
> 
> Would it make sense to merge the two projects further together?

Looks great @vbreuss, I think we're good then. I'll prepare a small PR to mention the Testably library.

PR #887, opened 2022-09-23
fix: Fixing UNC path GetFullPath when Current Directory is a UNC path

michal-ciechan:
Currently, if you switch current directory in the `MockFileSystem` to a UNC path, and try to enumerate files, it blows up.

Attempting to fix it, although it feels like a hack.

### DIFF
--- a/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockPath.cs
+++ b/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockPath.cs
@@ -85,7 +85,7 @@ namespace System.IO.Abstractions.TestingHelpers
             // unc paths need at least two segments, the others need one segment
             var isUnixRooted = mockFileDataAccessor.StringOperations.StartsWith(
                 mockFileDataAccessor.Directory.GetCurrentDirectory(),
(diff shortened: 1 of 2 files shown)


src/System.IO.Abstractions.TestingHelpers/MockPath.cs
```diff
@@ -115,13 +115,16 @@ public override string GetFullPath(string path)
             {
                 fullPath = DirectorySeparatorChar + fullPath;
             }
-            else if (isUnixRooted)
+            else if (isUnixRooted || isUnc)
```
michal-ciechan, line 118: feels like we should only set isUnixRooted if "current directory is not a UNC path" which would resolve this issue as well. Although re-using the same slashes feels safest

fgreinacher: I agree, that sounds more correct.  I think `isUnixRooted` should only be `true` when the path starts with a forward slash.

michal-ciechan: Done, much cleaner

fgreinacher (approved):
Thanks a lot 👍

PR #901, opened 2022-10-28
draft: use testably abstractions interface

vbreuss:
@fgreinacher 
As discussed in #883 I tried myself at a combination of the two interfaces. Here is a draft with a compiling `System.IO.Abstractions` library that uses the interfaces from [Testably.Abstractions.Interface](https://github.com/Testably/Testably.Abstractions/tree/main/Source/Testably.Abstractions.Interface).

**This PR is only intended to discuss the general approach and also to explore other alternatives.**

I had some difficulties, which I highlighted in [the last commit](https://github.com/TestableIO/System.IO.Abstractions/commit/6f98f8fc7f4e24645688b212dccfab347c25e7ff) with corresponding "TODO" code comments:
- You don't have an explicit type implementing just `IFileSystemInfo`, but this would be required for implementing `ResolveLinkTarget`
- You have overloads for `WriteAllLinesAsync` with an array instead of `IEnumerable<string>`. Where does this overload come from?
- Currently I don't explicitly target .NET5. The `MoveTo` method with `override` parameter is available starting with .NET5, but not with netstandard2.1 (which is used in the background). Do you plan to remove explicit support for .NET5 (as it is no longer supported by Microsoft)?
- Currently the explicit cast in the `FileSystemWatcherWrapper` can't be implemented, as you don't have a reference to the `IFileSystem` in the static method. Is this really required?
  *Note: I have a `Wrap` method on the `IFileSystemWatcherFactory` interface for this scenario...*

*Changes in the Testably project are tracked by [Issue 158](https://github.com/Testably/Testably.Abstractions/issues/158).*


fgreinacher:
Many thanks for the initiative here  @vbreuss, really appreciated!

I'm out traveling for the long weekend, will have a look mid next week!

fgreinacher:
Don't get me wrong here, but I'd really like this library to stay independent. I'm totally open to adapt the interfaces to be more compatible downstream, but I'm not eager to take a hard dependency here.

That being said here are my thoughts regarding your questions:

> * You don't have an explicit type implementing just `IFileSystemInfo`, but this would be required for implementing `ResolveLinkTarget`

(Not really relevant given my comment above, but we could add that implementation)

> * You have overloads for `WriteAllLinesAsync` with an array instead of `IEnumerable<string>`. Where does this overload come from?

Good question, it has been there for a while 😄 If it hurts we could get rid of it, overload resolution should pick the `IEnumerable<>` overload automatically.

> * Currently I don't explicitly target .NET5. The `MoveTo` method with `override` parameter is available starting with .NET5, but not with netstandard2.1 (which is used in the background). Do you plan to remove explicit support for .NET5 (as it is no longer supported by Microsoft)?

No plans to remove it right now, maybe when .NET 7 is released.

> * Currently the explicit cast in the `FileSystemWatcherWrapper` can't be implemented, as you don't have a reference to the `IFileSystem` in the static method. Is this really required?
>   _Note: I have a `Wrap` method on the `IFileSystemWatcherFactory` interface for this scenario..._

We explicitly highlight that in the readme so this could hurt folks and I think we should not remove it.

vbreuss:
@fgreinacher 
> Don't get me wrong here, but I'd really like this library to stay independent. I'm totally open to adapt the interfaces to be more compatible downstream, but I'm not eager to take a hard dependency here.

All good!
My PR was not aimed to introduce any dependency, but rather to check and highlight the existing differences in the interfaces!
My bad, if I didn't communicate the intention clear enough 😄

> That being said here are my thoughts regarding your questions:
> 
> > * You don't have an explicit type implementing just `IFileSystemInfo`, but this would be required for implementing `ResolveLinkTarget`
> 
> (Not really relevant given my comment above, but we could add that implementation)
> > * You have overloads for `WriteAllLinesAsync` with an array instead of `IEnumerable<string>`. Where does this overload come from?
> 
> Good question, it has been there for a while 😄 If it hurts we could get rid of it, overload resolution should pick the `IEnumerable<>` overload automatically.
> 
> > * Currently I don't explicitly target .NET5. The `MoveTo` method with `override` parameter is available starting with .NET5, but not with netstandard2.1 (which is used in the background). Do you plan to remove explicit support for .NET5 (as it is no longer supported by Microsoft)?
> 
> No plans to remove it right now, maybe when .NET 7 is released.
> 
> > * Currently the explicit cast in the `FileSystemWatcherWrapper` can't be implemented, as you don't have a reference to the `IFileSystem` in the static method. Is this really required?
> >   _Note: I have a `Wrap` method on the `IFileSystemWatcherFactory` interface for this scenario..._
> 
> We explicitly highlight that in the readme so this could hurt folks and I think we should not remove it.

I will work on a pull request to implement [your suggestion in #883](https://github.com/TestableIO/System.IO.Abstractions/pull/883#issuecomment-1300891750)
- Move out the Interfaces in System.IO.Abstractions into a separate project "TestableIO.System.IO.Abstractions"
- Move the Wrapper implementation to a project "TestableIO.System.IO.Abstractions.Wrappers"
I will use this abstractions project as a reference in my interface project to replace the file-related interfaces and keep the `ITimeSystem` there.
There will definitely be some changes, but I will highlight them in the PR and we can adress them one by one...

PR #906, opened 2022-11-05
refactor: improve and consolidate interfaces

vbreuss:
Make the following changes for the interfaces:
1. Enable [Nullable reference types](https://learn.microsoft.com/en-us/dotnet/csharp/nullable-references)
2. Improve and cleanup XML documentation and sort properties and methods in interfaces alphabetically
3. Consolidate the naming in the factory methods to always use just `New`; so instead of `new DirectoryInfo(path)` you should use `IFileSystem.DirectoryInfo.New(path)`. The previous existing factory methods were kept and marked as `Obsolete`.
4. Added a `Wrap` method to the factory interfaces that can be used instead of a direct cast (which is not possible for interfaces) to create a wrapped interface from a "System.IO" type.
5. Removed the `WriteAllLinesAsync` with array overload, as it is not defined in `System.IO` and overload resolution will automatically use the `IEnumerable` overload.
6. Added missing methods: `ResolveLinkTarget` and `CreateAsSymbolicLink`
7. Change the return type for Stream-Methods to a custom defined `FileSystemStream`.
   *Note: This should fix the issues from #793 , right?*

Added the following extensibility measures:
- Add `IFileSystem FileSystem` property to more classes to enable implementing extension methods.
  This was simplified by adding a base interface `IFileSystemExtensionPoint`
- Add a `IFileSystemExtensibility Extensibility` property on `IFileSystemInfo` and `IFileSystemStream` which grants access to the underlying wrapped instance or to a metadata dictionary to store additional information together with the instance.


vbreuss:
@fgreinacher : 
I tried to refactor (especially the interfaces). If the general approach of this PR is OK, I would add unit tests and try to implement the missing methods `ResolveLinkTarget` and `CreateAsSymbolicLink`.

vbreuss:
The tests fail due to Parity-Errors, as I implemented some additional properties. What do I need to change in order to update the Parity-Tests?

tests/TestableIO.System.IO.Abstractions.Wrappers.Tests/ApiParityTests.cs
```diff
@@ -7,7 +7,7 @@
 
 namespace System.IO.Abstractions.Tests
 {
-    [TestFixture]
+    [TestFixture(IgnoreReason = "TODO: temporarily disable in #906 as it is unclear how the parity tests are updated.")]
```
vbreuss, line 10: This needs to be fixed before merging this PR!

vbreuss:
@fgreinacher : 
Did you have time to have a look at this PR?
If this is through, I plan to add two further updates, but in separate PRs:
1. Extract the ACL-related method to extension methods in [`TestableIO.System.IO.Abstractions.Wrappers`](https://github.com/TestableIO/System.IO.Abstractions/tree/main/src/TestableIO.System.IO.Abstractions.Wrappers)
2. Add support for the new methods introduced in [.NET 7](https://devblogs.microsoft.com/dotnet/announcing-dotnet-7/)

fgreinacher:
> @fgreinacher : Did you have time to have a look at this PR? If this is through, I plan to add two further updates, but in separate PRs:
> 
> 1. Extract the ACL-related method to extension methods in [`TestableIO.System.IO.Abstractions.Wrappers`](https://github.com/TestableIO/System.IO.Abstractions/tree/main/src/TestableIO.System.IO.Abstractions.Wrappers)
> 2. Add support for the new methods introduced in [.NET 7](https://devblogs.microsoft.com/dotnet/announcing-dotnet-7/)

Sorry @vbreuss, I'll have a look today!

src/TestableIO.System.IO.Abstractions.Wrappers/FileSystemWatcherBase.cs
```diff
@@ -77,7 +83,7 @@ public void Dispose()
                 return null;
             }
 
-            return new FileSystemWatcherWrapper(watcher);
+            return new FileSystemWatcherWrapper(null, watcher);
```
fgreinacher, line 86: Can we do the same like for the other wrappers, e.g. https://github.com/TestableIO/System.IO.Abstractions/blob/8a24fd27db06503654d0a9856818e797e4898a7f/src/TestableIO.System.IO.Abstractions.Wrappers/FileInfoBase.cs#L92?

```suggestion
            return new FileSystemWatcherWrapper(new FileSystem), watcher);
```

src/TestableIO.System.IO.Abstractions/IDirectory.cs
```diff
@@ -45,105 +89,112 @@ public interface IDirectory
...
+        /// <inheritdoc cref="Directory.GetParent(string)" />
+        IDirectoryInfo? GetParent(string path);
+
+        /// <inheritdoc cref="Directory.Move(string, string)" />
         void Move(string sourceDirName, string destDirName);
 
+#if FEATURE_CREATE_SYMBOLIC_LINK
+        /// <inheritdoc cref="Directory.ResolveLinkTarget(string, bool)" />
+        IFileSystemInfo? ResolveLinkTarget(string linkPath, bool returnFinalTarget);
+#endif
```
fgreinacher, line 170: Can you please move this change out to separate `feat` PR so that have only refactoring here?

src/TestableIO.System.IO.Abstractions/IDirectory.cs
```diff
@@ -20,17 +18,63 @@ public interface IDirectory
...
         void Delete(string path);
-        /// <inheritdoc cref="Directory.Delete(string,bool)"/>
+
+        /// <inheritdoc cref="Directory.Delete(string, bool)" />
         void Delete(string path, bool recursive);
-        /// <inheritdoc cref="Directory.Exists"/>
-        bool Exists(string path);
 
+        /// <inheritdoc cref="Directory.EnumerateDirectories(string)" />
+        IEnumerable<string> EnumerateDirectories(string path);
```
fgreinacher, line 34: I see you did some reordering here, that's cool!

Did you use an automated tool for this? If so, which?

src/TestableIO.System.IO.Abstractions/IFileSystemExtensionPoint.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace System.IO.Abstractions
+{
+    /// <summary>
+    ///     Interface to support implementing extension methods on top of nested <see cref="IFileSystem" /> interfaces.
+    /// </summary>
+    public interface IFileSystemExtensionPoint
```
fgreinacher, line 6: Can we maybe give this name a bit more meaning?

```suggestion
    public interface IFileSystemEntity
```

src/TestableIO.System.IO.Abstractions/IFileSystemExtensibility.cs
```diff
@@ -0,0 +1,37 @@
+﻿using System.Diagnostics.CodeAnalysis;
+
+namespace System.IO.Abstractions
+{
+    /// <summary>
+    ///     A container to give access to the wrapped instance and to store metadata on <see cref="IFileSystem" /> entities.
+    ///     <br />
+    ///     This is useful for writing extension methods.
+    /// </summary>
+    public interface IFileSystemExtensibility
```
fgreinacher, line 10: I don't fully understand why we need this. Can you give a specific example where this is useful?

tests/TestableIO.System.IO.Abstractions.Wrappers.Tests/ApiParityTests.cs
fgreinacher: Each mismatched snapshot will be created in a `__mismatch__` directory below https://github.com/TestableIO/System.IO.Abstractions/tree/main/tests/TestableIO.System.IO.Abstractions.Wrappers.Tests/__snapshots__.

Just overwrite files from the `__snapshots__` directory with the mismatched ones to acknowledge them.

See https://swisslife-oss.github.io/snapshooter/docs/snapshot-mismatch-handling

fgreinacher (changes requested):
Thanks a lot @vbreuss, this looks great, I left some suggestions/questions!

src/TestableIO.System.IO.Abstractions/IDirectory.cs
vbreuss: I use [ReSharper](https://www.jetbrains.com/resharper/) to keep a consistent file layout and code style.
But due to the many `#if` / `#endif` structures, the alphabetical sorting does not work very good, so I had to do it by hand.

Without it, it would be hard to double-check it against my existing interface implementation :-)

src/TestableIO.System.IO.Abstractions/IFileSystemExtensibility.cs
vbreuss: On .NET core, Microsoft implemented the ACL-Functionality using extension methods (see [FileSystemAclExtensions](https://learn.microsoft.com/en-us/dotnet/api/system.io.filesystemaclextensions?view=net-6.0).
When looking for a way to do something similar, I implemented this extensibility method, so I could
- Get the underlying instance if it is there and execute the extension method on it or
- Store/Retrieve some data on the mocked instance and work with this storage mechanism instead

After reading your comment, I think it would make more sense, if I remove it from this PR and add it in again later, together with the ACL-Refactoring, when the use case makes it clear what it is for...

src/TestableIO.System.IO.Abstractions/IDirectory.cs
vbreuss: Will revert it here and add a separate PR, after the interface changes are finished...
I probably got a bit carried away here ;-)

tests/TestableIO.System.IO.Abstractions.Wrappers.Tests/ApiParityTests.cs
vbreuss: Thanks for the hint, I re-enabled the parity tests. Please note, that I had to change some mappings in [ApiParityTests](https://github.com/TestableIO/System.IO.Abstractions/pull/906/commits/4da6ca728b4b455e22a63864374383874c1b54c1#diff-3c3ddade875d30a4c4d1b8088475ca0128dc09de21ae6dda4b3e913d1ab81a92) due to the changed return values:
- `FileSystemStream` instead of `Stream`
- `IWaitForChangedResult` instead of `WaitForChangedResult`

vbreuss:
Thanks for the review, @fgreinacher !

I implemented all review findings and answered your questions :-)
*Could you please ignore the Codacy-Issue, as I can't remove the `partial` from `IFile` (seems like a false positive)?*

Are the next planned steps also OK for you?
1. Extract the ACL-related method to extension methods in [TestableIO.System.IO.Abstractions.Wrappers](https://github.com/TestableIO/System.IO.Abstractions/tree/main/src/TestableIO.System.IO.Abstractions.Wrappers)
2. Add support for the new methods introduced in [.NET 7](https://devblogs.microsoft.com/dotnet/announcing-dotnet-7/)
3. Re-Implement the "ResolveLinkTarget" functionality

src/TestableIO.System.IO.Abstractions/IDirectory.cs
fgreinacher: Ah, I see. Was just wondering if we should add some more automation to automatically handle this. All good :)

src/TestableIO.System.IO.Abstractions/IFile.cs
```diff
@@ -1,59 +1,75 @@
...
 using System.Text;
 
 namespace System.IO.Abstractions
 {
-    /// <inheritdoc cref="File" />
-    public partial interface IFile
+    /// <summary>
+    /// Abstractions for <see cref="File" />.
+    /// </summary>
+    public partial interface IFile : IFileSystemEntity
```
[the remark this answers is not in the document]
vbreuss: I could add this, but do you really think it adds value?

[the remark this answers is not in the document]
fgreinacher: That's always the question with static code analysis tools 😀

We can keep it as-is!

fgreinacher (approved):
Thanks again for your hard work here!

The next steps are also fine from my side 👍

vbreuss:
@fgreinacher : 
It seems Codacy didn't trigger a new scan.
> [Pull Request Scan](https://app.codacy.com/gh/TestableIO/System.IO.Abstractions/pullRequest?prid=10505853)

Is there anything I can do?

PR #912, opened 2022-11-15
fix: change rollForward to "latestMinor" in global.json

vbreuss:
**Currently it is impossible to build System.IO.Abstractions with the latest Visual Studio version (v17.4.0) :-(**

"latestMinor" means:
> Uses the highest installed minor, feature band, and patch level that matches the requested major with a minor, feature band, and patch level that is greater or equal than the specified value.

*See [Microsoft Documentation](https://learn.microsoft.com/en-us/dotnet/core/tools/global-json#rollforward)*

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "7.0.100",
-    "rollForward": "latestFeature"
+    "version": "6.0.403",
+    "rollForward": "latestMajor"
   }
 }
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "6.0.403",
-    "rollForward": "latestMinor"
+    "version": "7.0.100",
+    "rollForward": "latestFeature"
   }
 }
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "sdk": {
     "version": "6.0.403",
-    "rollForward": "latestFeature"
+    "rollForward": "latestMajor"
   }
 }
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "sdk": {
     "version": "6.0.403",
-    "rollForward": "latestMajor"
+    "rollForward": "latestMinor"
   }
 }


fgreinacher:
Does VS not include any .NET 6 SDK anymore?

vbreuss:
> Does VS not include any .NET 6 SDK anymore?

Unfortunately after updating to v17.4.0 I couldn't build the solution anymore. I had to delete `global.json` locally.

Would it be OK, to change the "rollForward" to "latestMajor", effectively allowing any newer version of the SDK?

fgreinacher:
> Unfortunately after updating to v17.4.0 I couldn't build the solution anymore.

What error message did you get?

> Would it be OK, to change the "rollForward" to "latestMajor", effectively allowing any newer version of the SDK?

I'm a big fan of pinning versions for reproducability. `latestMajor` is equivalent to deleting the `global.json` so I'd rather not do like to go that way. `latestMinor` would be the maximum for me I think.

fgreinacher:
Also when using `latestMinor`?

vbreuss:
> Also when using `latestMinor`?

Yes, unfortunatly the same:
image

vbreuss:
@fgreinacher :

Please note, that the problem will be more imminent, when adding .NET 7 support, as you then **have to** use this version of visual studio, but the [SDK-Version 7.0.100](https://learn.microsoft.com/en-us/dotnet/core/porting/versioning-sdk-msbuild-vs) is not yet available in GitHub actions...

fgreinacher:
Can you build from the command line?

What does `dotnet --list-sdks` tell you?

vbreuss:
> Can you build from the command line?

No, unfortunately not.

> What does `dotnet --list-sdks` tell you?

image


> Sounds similar to https://developercommunity.visualstudio.com/t/Updating-Visual-Studio-2022-to-174-prev/10154141?q=.net+6+removed. Maybe just install .NET 6 SDK manually?
.
**After manually installing the [.NET 6 SDK](https://dotnet.microsoft.com/en-us/download/dotnet/6.0) the build works**

However Microsoft displays a warning when I download the .NET6 SDK :-(
https://dotnet.microsoft.com/en-us/download/dotnet/thank-you/sdk-6.0.403-windows-x64-installer
image

*So the immediate pain is fixed for me, thanks for the hint @fgreinacher*

fgreinacher:
Wow, this sounds so complicated 😬

Would you change `rollForward` to `latestMinor` so that we all .NET 6 versions?

vbreuss:
> Wow, this sounds so complicated 😬
> 
> Would you change `rollForward` to `latestMinor` so that we all .NET 6 versions?

Of course I can, but I don't understand what this is supposed to help...

fgreinacher:
That way any .NET 6 SDK will work, which will prevent things from breaking unexpectedly. Thanks!

PR #913, opened 2022-11-15
refactor: extract ACL related functionality

vbreuss:
Remove the ACL-Features from the interfaces and instead implement them as extension methods implemented in "TestableIO.System.IO.Abstractions.Wrappers".

This removes the dependency on `System.IO.FileSystem.AccessControl` from the interface project.

This is part of the approach discussed in #883

BREAKING CHANGE: This refactoring removes ACL-related methods from the interface project and replaces them with extension methods in `TestableIO.System.IO.Abstractions.Wrappers`.


vbreuss:
@fgreinacher :
I finished part 1 of the next PR. Could you please have a look?

src/System.IO.Abstractions.TestingHelpers/MockDirectoryInfo.cs
```diff
@@ -27,6 +27,15 @@ public MockDirectoryInfo(IMockFileDataAccessor mockFileDataAccessor, string dire
...
             originalPath = directoryPath;
+
+            if (directoryPath== null)
+            {
+                throw new ArgumentNullException("path", StringResources.Manager.GetString("VALUE_CANNOT_BE_NULL"));
+            }
+            if (directoryPath.Trim() == string.Empty)
+            {
+                throw CommonExceptions.PathIsNotOfALegalForm("path");
+            }
```
siprbaum, line 38: Should this (similar to the proposed change in MockFileInfo) also rely on `mockFileSystem.PathVerifier.IsLegalAbsoluteOrRelative(path, "path");` to check for a valid path?
```suggestion
            if (directoryPath== null)
            {
                throw new ArgumentNullException("path", StringResources.Manager.GetString("VALUE_CANNOT_BE_NULL"));
            }
            if (directoryPath.Trim() == string.Empty)
            {
                throw CommonExceptions.PathIsNotOfALegalForm("path");
            }
            
            originalPath = directoryPath;
```

vbreuss: @siprbaum :
> Should this (similar to the proposed change in MockFileInfo) also rely on `mockFileSystem.PathVerifier.IsLegalAbsoluteOrRelative(path, "path");` to check for a valid path?

To answer your question: I started implementing it using `PathVerifier.IsLegalAbsoluteOrRelative`, but it resulted in 47 failed tests; most often about "Illegal characters in path.".
I didn't have time to analyse them in depth to see, if I could refactor the constructor and/or the tests, to find a better solution. This was the change with the compromise with the least impact and I think changing more would need to be part of a different pull request (if at all).

vbreuss:
@fgreinacher : This time the two issues found by codacy are false positives.
I need these parameters as they are interface implementations, right?

fgreinacher:
I'll have a deeper look at this by the middle of the week.

@siprbaum Do you want to do another pass?

src/TestableIO.System.IO.Abstractions.Wrappers/DirectoryAclExtensions.cs
```diff
@@ -0,0 +1,102 @@
...
+            if (extensibility.TryGetWrappedInstance(out DirectoryInfo di))
+            {
+                di.Create(directorySecurity);
+            }
+            else
+            {
+                extensibility.StoreMetadata("AccessControl:DirectorySecurity",
+                    directorySecurity);
+                directoryInfo.Create();
+            }
```
fgreinacher, line 34: I don't fully understand why we need this code path. Couldn't we just throw a `NotSupportedException` when the underlying type does not support the ACL APIs?

vbreuss: I need to store the `DirectoryInfo` so that I can return the value when calling `GetAccessControl` on a mocked instance.
The idea is to
- Use the functionality of the underlying file system when working on a real file system
- Otherwise store the value as metadata, so that I can retrieve it in other methods.

src/TestableIO.System.IO.Abstractions.Wrappers/FileAclExtensions.cs
```diff
@@ -0,0 +1,73 @@
...
+        /// <inheritdoc cref="File.GetAccessControl(string)"/>
+#endif
+        [SupportedOSPlatform("windows")]
+        public static FileSecurity GetAccessControl(
+            this IFile file, string path)
+        {
+            IFileInfo fileInfo = file.FileSystem.FileInfo.New(path);
+            IFileSystemExtensibility extensibility =
+                fileInfo.Extensibility;
+            return extensibility.TryGetWrappedInstance(out FileInfo fi)
```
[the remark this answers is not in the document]
vbreuss: I don't understand how you would solve this otherwise.
In order to implement the extension methods, I need access to the underlying instance (if it is available) and otherwise be able to store or retrieve metadata.
That's now just one extension method that I implemented in the TestableIO.System.IO.Abstractions.Wrappers project; but I need the same functionality, when I implement it in [Testably.Abstractions.AccessControl](https://github.com/Testably/Testably.Abstractions/tree/main/Source/Testably.Abstractions.AccessControl), so the interface itself needs to be public.

What I could do is to "remove" it from the `IDirectoryInfo` and "just add it as a separate interface to `MockDirectoryInfo` and `DirectoryInfoWrapper`. In the extension methods, I would then change it to 
```csharp
public static DirectorySecurity GetAccessControl(
            this IDirectoryInfo directoryInfo,
            AccessControlSections includeSections)
        {
            IFileSystemExtensibility extensibility =
                directoryInfo as IFileSystemExtensibility;
            return extensibility.TryGetWrappedInstance(out DirectoryInfo di)
                ? di.GetAccessControl(includeSections)
                : extensibility.RetrieveMetadata<DirectorySecurity>(
                    "AccessControl:DirectorySecurity") ?? new DirectorySecurity();
        }
```
Is this a better approach?

src/TestableIO.System.IO.Abstractions.Wrappers/DirectoryAclExtensions.cs
fgreinacher: > Otherwise store the value as metadata, so that I can retrieve it in other methods.

But in the end ACLs only work on the real file system, so why not prevent changing them in the beginning?

src/TestableIO.System.IO.Abstractions.Wrappers/FileAclExtensions.cs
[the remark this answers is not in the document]
fgreinacher: Hm, what I don't like about this approach is the "broadness" of the interface `IFileSystemExtensibility`.

Could we make this more specific, e.g. like this:

```csharp
public static DirectorySecurity GetAccessControl(
            this IDirectoryInfo directoryInfo,
            AccessControlSections includeSections)
{
    return directoryInfo is IDirectoryInfoAclSupport directoryInfoAclSupport
        ? directoryInfoAclSupport.GetAccessControl(includeSections)
        : null
}
```

And for the write operations:


```csharp
public static DirectorySecurity SetAccessControl(
            this IDirectoryInfo directoryInfo,
            DirectorySecurity directorySecurity)
{
    return directoryInfo is IDirectoryInfoAclSupport directoryInfoAclSupport
        ? directoryInfoAclSupport.SetAccessControl(directorySecurity)
        : throw new NotSupportedException();
}
```

Your custom `Mock` classes could then easily implement whatever works best for `IDirectoryInfoAclSupport`.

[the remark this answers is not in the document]
vbreuss: If I add explicit types to the interface, I still need the reference to `System.IO.FileSystem.AccessControl`, which is the whole point of this pull request :-(

I could do the following:
1. Remove the `Extensibility` properties from the interfaces
2. Add the extensibility interface instead on the corresponding wrapper or mock classes, e.g.
   ```csharp
       public class MockDirectoryInfo : DirectoryInfoBase, IFileSystemExtensibility
   ```
3. Implement the extensibility interface, e.g.
   ```csharp
        /// <inheritdoc cref="IFileSystemExtensibility.TryGetWrappedInstance{T}(out T)" />
        public bool TryGetWrappedInstance<T>(out T wrappedInstance)
            => GetExtensibility().TryGetWrappedInstance<T>(out wrappedInstance);

        /// <inheritdoc cref="IFileSystemExtensibility.StoreMetadata{T}(string, T)" />
        public void StoreMetadata<T>(string key, T value)
            => GetExtensibility().StoreMetadata<T>(key, value);

        /// <inheritdoc cref="IFileSystemExtensibility.RetrieveMetadata{T}(string)" />
        public T RetrieveMetadata<T>(string key)
        => GetExtensibility().RetrieveMetadata<T>(key);
        
        private IFileSystemExtensibility GetExtensibility()
        {
            var mockFileData = mockFileDataAccessor.GetFile(directoryPath);
            return mockFileData?.Extensibility ?? FileSystemExtensibility.GetNullObject(
                () => CommonExceptions.CouldNotFindPartOfPath(directoryPath));
        }
   ```
4. Adjust the ACL extension methods, e.g.
   ```csharp
        public static void Create(this IDirectoryInfo directoryInfo, DirectorySecurity directorySecurity)
        {
            IFileSystemExtensibility extensibility = directoryInfo as IFileSystemExtensibility;
            if (extensibility == null)
            {
                throw new NotSupportedException("The directory info does not support extensibility methods.")
            }

            if (extensibility.TryGetWrappedInstance(out DirectoryInfo di))
            {
                di.Create(directorySecurity);
            }
            else
            {
                directoryInfo.Create();
                extensibility.StoreMetadata("AccessControl:DirectorySecurity",
                    directorySecurity);
            }
        }
   ```

This way, the extensibility is "hidden" in normal use and more like an internal interface, even though it is public ;-)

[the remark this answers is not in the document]
fgreinacher: Ah sorry, I was back and forth with my example. I meant this:

```csharp

```csharp
public static object GetAccessControl(
            this IDirectoryInfo directoryInfo,
            object includeSections)
{
    return directoryInfo is IDirectoryInfoAclSupport directoryInfoAclSupport
        ? directoryInfoAclSupport.GetAccessControl(includeSections)
        : null
}
```

And for the write operations:


```csharp
public static void SetAccessControl(
            this IDirectoryInfo directoryInfo,
            object directorySecurity)
{
    if (directoryInfo is IDirectoryInfoAclSupport directoryInfoAclSupport)
        directoryInfoAclSupport.SetAccessControl(directorySecurity);
    else 
        throw new NotSupportedException();
}
```

Your suggestion is a little bit better, but still has the broad extensibility interface.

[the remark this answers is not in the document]
vbreuss: I replaced the `IFileSystemExtensibility` with the following interface:
```csharp
    /// <summary>
    ///     Support ACL functionality on file system instances.
    /// </summary>
    public interface IFileSystemAclSupport
    {
        /// <summary>
        ///     Gets a access control object that encapsulates the access control list (ACL) entries for the file or directory in the file system.
        /// </summary>
        [SupportedOSPlatform("windows")]
        object GetAccessControl();

        /// <summary>
        ///     Gets a access control object that encapsulates the access control list (ACL) entries for the file or directory in the file system.
        /// </summary>
        /// <param name="includeSections">One of the <see cref="AccessControlSections" /> values that specifies the type of access control list (ACL) information to receive.</param>
        [SupportedOSPlatform("windows")]
        object GetAccessControl(AccessControlSections includeSections);

        /// <summary>
        ///     Applies access control list (ACL) entries described by the <paramref name="value"/> object to the file or directory in the file system.
        /// </summary>
        [SupportedOSPlatform("windows")]
        void SetAccessControl(object value);

        /// <summary>
        /// Specifies which sections of a security descriptor to save or load.</summary>
        [Flags]
        public enum AccessControlSections
        {
            /// <summary>
            /// No sections.
            /// </summary>
            None = 0,

            /// <summary>
            /// The system access control list (SACL).
            /// </summary>
            Audit = 1,

            /// <summary>
            /// The discretionary access control list (DACL).
            /// </summary>
            Access = 2,

            /// <summary>
            /// The owner.
            /// </summary>
            Owner = 4,

            /// <summary>
            /// The primary group.
            /// </summary>
            Group = 8,

            /// <summary>
            /// The entire security descriptor.
            /// </summary>
            All = Group | Owner | Access | Audit
        }
    }
```

The FileSystemExtensibility is now internal in the TestingHelpers project.

Is this, how you meant your suggestion, @fgreinacher ?

src/System.IO.Abstractions.TestingHelpers/FileSystemExtensibility.cs
```diff
@@ -0,0 +1,32 @@
...
+        {
+            if (_metadata.TryGetValue(key, out object value))
+            {
+                return value;
+            }
+
+            return default;
+        }
+
+        internal void CloneFrom(FileSystemExtensibility template)
```
siprbaum, line 24: Clone or CloneFrom suggests to me that a new identical object is created.
But this method only copies  the values from the `template`. Any existing metadata are kept.

Perhaps change to `CopyFrom` ?

src/System.IO.Abstractions.TestingHelpers/MockDirectoryData.cs
```diff
@@ -25,9 +21,16 @@ public new DirectorySecurity AccessControl
...
                 // This let's us use this class as long as we don't use AccessControl property.
-                return accessControl ?? (accessControl = new DirectorySecurity());
+                var value = Extensibility.RetrieveMetadata("AccessControl:DirectorySecurity");
+                if (value is not DirectorySecurity directorySecurity)
+                {
+                    directorySecurity = new DirectorySecurity();
+                    Extensibility.StoreMetadata("AccessControl:DirectorySecurity", directorySecurity);
+                }
+
+                return directorySecurity;
```
siprbaum, line 31: May it is just me not being used to the ` value is not DirectorySecurity directorySecurity`, but I find this much more readable. I was staring for over a minute on the code to see were `directorySecurity` is defined or if it is a global variable. [ Feel free to ignore this style, very likely just personal taste, but I want to toss it out here ...]
```suggestion
                var value = Extensibility.RetrieveMetadata("AccessControl:DirectorySecurity") as DirectorySecurity;
                if (value == null)
                {
                    directorySecurity = new DirectorySecurity();
                    Extensibility.StoreMetadata("AccessControl:DirectorySecurity", directorySecurity);
                }

                return directorySecurity;
```

src/System.IO.Abstractions.TestingHelpers/MockDirectoryInfo.cs
```diff
@@ -26,6 +27,15 @@ public MockDirectoryInfo(IMockFileDataAccessor mockFileDataAccessor, string dire
         {
             this.mockFileDataAccessor = mockFileDataAccessor ?? throw new ArgumentNullException(nameof(mockFileDataAccessor));
 
+            if (directoryPath== null)
```
siprbaum, line 30: Style, spaces around ` == `
```suggestion
            if (directoryPath == null)
```

src/System.IO.Abstractions.TestingHelpers/MockFileData.cs
```diff
@@ -188,9 +191,16 @@ public FileSecurity AccessControl
...
                 // This let's us use this class as long as we don't use AccessControl property.
-                return accessControl ?? (accessControl = new FileSecurity());
+                var value = Extensibility.RetrieveMetadata("AccessControl:FileSecurity");
+                if (value is not FileSecurity fileSecurity)
+                {
+                    fileSecurity = new FileSecurity();
+                    Extensibility.StoreMetadata("AccessControl:FileSecurity", fileSecurity);
+                }
+
+                return fileSecurity;
```
siprbaum, line 201: Similar suggestion  to the other MockDirectoryData one. [ Feel free to ignore]
```suggestion
                var value = Extensibility.RetrieveMetadata("AccessControl:FileSecurity") as FileSecurity fileSecurity;
                if (value == null)
                {
                    fileSecurity = new FileSecurity();
                    Extensibility.StoreMetadata("AccessControl:FileSecurity", fileSecurity);
                }

                return fileSecurity;
```

src/System.IO.Abstractions.TestingHelpers/FileSystemExtensibility.cs
vbreuss: I will change it to "CopyMetadataFrom".

src/System.IO.Abstractions.TestingHelpers/MockDirectoryData.cs
vbreuss: Will change it as you suggest. I am just used to this new style, but can understand keeping it simple :-)

src/System.IO.Abstractions.TestingHelpers/MockFileData.cs
vbreuss: Will change it and also similar issues in
- DirectoryInfoAclExtensions
- FileInfoAclExtensions
- FileStreamAclExtensions

src/System.IO.Abstractions.TestingHelpers/MockDirectoryInfo.cs
vbreuss: Fixed in latest commit

vbreuss:
@fgreinacher :
I think it would be good to finish this PR before bumping the major version in #919, as this PR also changes the dependencies...

fgreinacher:
> @fgreinacher : I think it would be good to finish this PR before bumping the major version in #919, as this PR also changes the dependencies...

@vbreuss I'd like to push out a clean v18 first. We can then just bump major again for this change, no problem :)

vbreuss:
> > @fgreinacher : I think it would be good to finish this PR before bumping the major version in #919, as this PR also changes the dependencies...
> 
> @vbreuss I'd like to push out a clean v18 first. We can then just bump major again for this change, no problem :)

@fgreinacher: I merged the latest changes from `main` including the v18 refactoring.
I am not sure, if the change really needs a major version bump; it just removes a reference, from the interface and replaces it with extension methods. However, if somebody only uses the interface project (which didn't exist previously), the ACL-methods are now missing...

src/System.IO.Abstractions.TestingHelpers/FileSystemExtensibility.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System.Collections.Generic;
```
fgreinacher, line 1: I think this file can be deleted

src/TestableIO.System.IO.Abstractions.TestingHelpers/MockFileData.cs
```diff
@@ -34,10 +34,13 @@ public class MockFileData
...
         /// <summary>
-        /// The access control of the <see cref="MockFileData"/>.
+        /// The extensibility of the <see cref="MockFileData"/>.
         /// </summary>
+        internal FileSystemExtensibility Extensibility
+            => extensibility;
+
         [NonSerialized]
-        private FileSecurity accessControl;
+        private FileSystemExtensibility extensibility = new FileSystemExtensibility();
```
fgreinacher, line 43: What's the benefit of going through this abstraction  here? We could just keep the field I think?

src/TestableIO.System.IO.Abstractions/IFileSystemAclSupport.cs
```diff
@@ -0,0 +1,65 @@
...
+        ///     Gets a access control object that encapsulates the access control list (ACL) entries for the file or directory in the file system.
+        /// </summary>
+        [SupportedOSPlatform("windows")]
+        object GetAccessControl();
+
+        /// <summary>
+        ///     Gets a access control object that encapsulates the access control list (ACL) entries for the file or directory in the file system.
+        /// </summary>
+        /// <param name="includeSections">One of the <see cref="AccessControlSections" /> values that specifies the type of access control list (ACL) information to receive.</param>
+        [SupportedOSPlatform("windows")]
```
fgreinacher, line 20: I think we can get rid of this attribute as we have it on the class?

```suggestion
```

fgreinacher (changes requested):
Thanks for your rework @vbreuss - I like it a lot know.

I have a few small suggestions/questions and then we're good to go 🚀

src/TestableIO.System.IO.Abstractions.Wrappers/FileAclExtensions.cs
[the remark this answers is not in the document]
fgreinacher: Looks great now!

fgreinacher:
> I am not sure, if the change really needs a major version bump; it just removes a reference, from the interface and replaces it with extension methods. However, if somebody only uses the interface project (which didn't exist previously), the ACL-methods are now missing...

It will break people that are currently mocking the methods on the interface level, so I'd rather bump major to not surprise folks :)

Could you update the description to include a `BREAKING CHANGE` notice like in https://github.com/TestableIO/System.IO.Abstractions/pull/919?

src/TestableIO.System.IO.Abstractions/IFileSystemAclSupport.cs
vbreuss: I removed all these attributes in the interface...

src/System.IO.Abstractions.TestingHelpers/FileSystemExtensibility.cs
fgreinacher: @vbreuss Let's delete this file? 😀

fgreinacher (approved):
LGTM, thanks @vbreuss 🎉

fgreinacher (approved):

PR #922, opened 2022-12-09
chore: ignore .NET 7

fgreinacher:
Let's stick with LTS versions for the moment.

If we decide to upgrade we can probably also bump C# language version and maybe more.

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "7.0.100",
+    "version": "6.0.403",
     "rollForward": "latestMinor"
   }
 }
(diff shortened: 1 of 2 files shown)


global.json
```diff
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "7.0.100",
+    "version": "6.0.403",
```
vbreuss, line 3: I am currently working on a pull request to support .NET 7.
As .NET 7 has quite some additions to the file system, I think it is important to keep up-to-date.

vbreuss:
I started with the .NET7 implementation in #923...

fgreinacher:
> @fgreinacher : I started with the .NET7 implementation in #923...

Ah OK, then let's close this 👍

PR #940, opened 2023-01-31
feat: Add an explicit cast operator for FileSystemStream

Birnsen:
This commit adds an explicit cast to change the fact, that the underlaying FileStream can not be accessed in any way.
This has also the advantage that if the library is updated no code change is necessary if a cast was present from a stream.


fgreinacher (approved):
Helpful little addition, thanks!

Issue #942, opened 2023-02-05
IsReadOnly is true when file does not exist


brinkdinges:
**Describe the bug**
When you get the FileInfo of a file that does not exist, IsReadOnly is true while the Exists property is false.

**To Reproduce**

```
FileSystem = new MockFileSystem();
var fileInfo = FileSystem.FileInfo.New("C:\\Folder\\Donuts.sldasm");
var isReadOnly = fileInfo.IsReadOnly;
```

**Expected behavior**
According to the [docs](https://learn.microsoft.com/en-us/dotnet/api/system.io.fileinfo.isreadonly?view=netframework-4.8.1), IsReadOnly should throw a FileNotFoundException when the file does not exist. 

**Additional context**
I've looked through previous issues but could not find any mentioning this. 

In my code, I was checking the FileInfo attributes to see if it was read-only. If a file does not exist, the attributes return -1, which, when you bitwise AND with any value, returns the positive value, apparently. Maybe that's happening internally as well.

![image](https://user-images.githubusercontent.com/15387410/216839619-de93ce04-33b8-4573-a26b-b0376c1d281c.png)

fgreinacher:
Thanks for reporting! PRs welcome 🤗

brinkdinges:
I created a console app in .NET 7 to let me compare System.IO with System.IO.Abstractions. It looks like the official docs are not the same as the current .NET code because IsReadOnly just returns true when the file does not exist:

![image](https://user-images.githubusercontent.com/15387410/217849734-7ad41f3d-0710-4983-9085-6b916030f51d.png)

When the file does exist:
![image](https://user-images.githubusercontent.com/15387410/217851002-ebdf832b-1433-430e-a1af-9b7955c5420a.png)

So is this a .NET issue? That doesn't seem right 😄

vbreuss:
@brinkdinges : I tested this issue in https://github.com/Testably/Testably.Abstractions/pull/233:
The tests run on Windows, MacOS and Linux and a wide range of .NET versions. In all circumstances `IsReadOnly` returns `true` when it does not exist.

I agree with you, that it looks strange, but as we try to mimic the real file system, this is the intended behaviour...

brinkdinges:
Thanks for checking it out. When I reread the docs, they probably only throw a FileNotFoundException when setting the property, not when getting it. I've sent in feedback on that page.

fgreinacher:
[...]

PR #944, opened 2023-02-09
fix: Fixed typos in test names, variables and method parameter name

brinkdinges:
While searching for a fix for #942, I noticed a few typos in (mostly) test method names so I fixed them. I didn't change any public methods that had any references.

I kept 'typos' like variable names that consist of multiple words but did not use camelcase, like 'newcontents'.


vbreuss (approved):
Looks good. Thanks @brinkdinges !

fgreinacher (approved):

PR #951, opened 2023-02-25
feat: introduce convenience methods into MockFileSystem

rcdailey:
- `AddEmptyFile()` added which is a convenience wrapper for `AddFile()` with an empty `MockFileData`.
- `AddFile()` overload added which consumes an `IFileInfo`.
- `AddDirectory()` overload added which consumes an `IDirectoryInfo`.
- `GetFile()` overload added which consumes an `IFileInfo`.

Unit test added for `AddEmptyFile()` but not the others, as the test code for them would have very little value.

Fixes TestableIO/System.IO.Abstractions.Extensions#32

### DIFF
--- a/tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests/MockFileSystemTests.cs
+++ b/tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests/MockFileSystemTests.cs
@@ -120,6 +120,17 @@ namespace System.IO.Abstractions.TestingHelpers.Tests
         }
 
         [Test]
+        public void MockFileSystem_AddEmptyFile_ShouldBeEmpty()
+        {
+            var path = XFS.Path(@"c:\some\file.txt");
+            var fileSystem = new MockFileSystem();
+
+            fileSystem.AddEmptyFile(path);
+
+            Assert.That(fileSystem.GetFile(path).TextContents, Is.EqualTo(""));
+        }
+
+        [Test]
         public void MockFileSystem_ByDefault_IsSerializable()
         {
             var file1 = new MockFileData("Demo\r\ntext\ncontent\rvalue");
--- a/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockFileSystem.cs
+++ b/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockFileSystem.cs
@@ -209,6 +209,53 @@ namespace System.IO.Abstractions.TestingHelpers
             }
         }
 
+        /// <summary>
+        /// Add a new file that is empty.
+        /// </summary>
+        /// <param name="path">A string representing the path of the new file to add.</param>
+        public void AddEmptyFile(string path)
+        {
+            AddFile(path, new MockFileData(""));
+        }
+
+        /// <summary>
+        /// Add a new file that is empty.
+        /// </summary>
+        /// <param name="path">An <see cref="IFileInfo"/> representing the path of the new file to add.</param>
+        public void AddEmptyFile(IFileInfo path)
+        {
+            AddEmptyFile(path.FullName);
+        }
+
+        /// <summary>
+        /// Add a new, empty directory.
+        /// </summary>
+        /// <param name="path">An <see cref="IDirectoryInfo"/> representing the path of the new directory to add.</param>
+        public void AddDirectory(IDirectoryInfo path)
+        {
+            AddDirectory(path.FullName);
+        }
+
+        /// <summary>
+        /// Add a new file with its contents set to a specified <see cref="MockFileData"/>.
+        /// </summary>
+        /// <param name="path">An <see cref="IFileInfo"/> representing the path of the new file to add.</param>
+        /// <param name="data">The data to use for the contents of the new file.</param>
+        public void AddFile(IFileInfo path, MockFileData data)
+        {
+            AddFile(path.FullName, data);
+        }
+
+        /// <summary>
+        /// Gets a file.
+        /// </summary>
+        /// <param name="path">The path of the file to get.</param>
+        /// <returns>The file. <see langword="null"/> if the file does not exist.</returns>
+        public MockFileData GetFile(IFileInfo path)
+        {
+            return GetFile(path.FullName);
+        }
+
         /// <inheritdoc />
         public void AddDirectory(string path)
         {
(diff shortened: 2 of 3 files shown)


rcdailey:
I realize the PR title workflow is failing, however I would like the project maintainers to introduce whatever prefix they deem appropriate when they squash merge my PR into the mainline. Thank you.

vbreuss:
Just a quick question from my side:
In https://github.com/TestableIO/System.IO.Abstractions.Extensions/issues/32 you discussed adding it as extension methods to a separate library or adding it directly to the `MockFileSystem`.

Why not add it as extension methods in this repository?

rcdailey:
> Just a quick question from my side: In [TestableIO/System.IO.Abstractions.Extensions#32](https://github.com/TestableIO/System.IO.Abstractions.Extensions/issues/32) you discussed adding it as extension methods to a separate library or adding it directly to the `MockFileSystem`.
> 
> Why not add it as extension methods in this repository?

I didn't see a reason to add them as extension methods. Typically I only do that when I do not have access to the class in question. However, in my mind at least, the entire point of contributing these upstream (into this repo) was to make them a part of the class. I already have these as extension methods in my own code base and they were working fine.

I'm unfamiliar with any design policies or preferences in this repo or of its its maintainers, so I went about this using my own judgement. Is there a problem with having them in the class itself?

vbreuss:
> > Just a quick question from my side: In [TestableIO/System.IO.Abstractions.Extensions#32](https://github.com/TestableIO/System.IO.Abstractions.Extensions/issues/32) you discussed adding it as extension methods to a separate library or adding it directly to the `MockFileSystem`.
> > Why not add it as extension methods in this repository?
> 
> I didn't see a reason to add them as extension methods. Typically I only do that when I do not have access to the class in question. However, in my mind at least, the entire point of contributing these upstream (into this repo) was to make them a part of the class. I already have these as extension methods in my own code base and they were working fine.
> 
> I'm unfamiliar with any design policies or preferences in this repo or of its its maintainers, so I went about this using my own judgement. Is there a problem with having them in the class itself?

@rcdailey :
Please don't get me wrong. Your solution is fine and I think that the convenience methods are a nice addition here!

I am just curious, if there is a specific reason why you implemented it this way, as you had to adapt the `IMockFileDataAccessor` interface and add quite a lot of methods there. Normally I try to keep interface definitions as minimal as possible.
By adding your implementation as extension methods in this repository, the interface could remain smaller.

 Are there disadvantages to extension methods that I am not aware of :-)

rcdailey:
I added them to the interface because everything else seemed to have been added there. I honestly didn't think too hard about it. Should I remove them from there? For my own understanding, how do you decide between adding to the interface or not? Happy to learn & adjust. I am not too strongly opinionated on how this was done. I'm relying on you fine folks to let me know how it should be done :-)

Thanks for the discussion!

rcdailey:
For what it's worth, I have these already in my own project as a set of extension methods. I'm happy to drop them in here just like this if that's what you want. Code is [here](https://github.com/recyclarr/recyclarr/blob/master/src/Recyclarr.TestLibrary/MockFileSystemExtensions.cs).

I think the reason I jumped right into making them part of the class was @fgreinacher's comment on the linked issue:

> I'm also fine adding them to [MockFileSystem](https://github.com/TestableIO/System.IO.Abstractions/blob/main/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockFileSystem.cs?rgh-link-date=2023-01-09T14%3A05%3A16Z).

Maybe I read that too literally. It's honestly not a big deal either way. Let me know how you'd like it done.

src/TestableIO.System.IO.Abstractions.TestingHelpers/IMockFileDataAccessor.cs
```diff
@@ -86,8 +87,40 @@ public interface IMockFileDataAccessor : IFileSystem
...
+        /// <param name="path">An <see cref="IFileInfo"/> representing the path of the new file to add.</param>
+        /// <param name="data">The data to use for the contents of the new file.</param>
+        void AddFile(IFileInfo path, MockFileData data);
+
+        /// <summary>
+        /// Gets a file.
+        /// </summary>
+        /// <param name="path">The path of the file to get.</param>
+        /// <returns>The file. <see langword="null"/> if the file does not exist.</returns>
+        MockFileData GetFile(IFileInfo path);
```
fgreinacher, line 124: I don' think these new methods need to be here. This interface is mostly used by other `Mock*` classes. (Actually it should probably be `internal`, but that would be a severe breaking change :/)

tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests/MockFileSystemTests.cs
```diff
@@ -103,20 +103,14 @@ public void MockFileSystem_AddFile_ShouldHandleNullFileDataAsEmpty()
         }
 
         [Test]
-        public void MockFileSystem_AddFile_ShouldReplaceExistingFile()
```
fgreinacher, line 106: Did you delete this deliberately?

fgreinacher:
> I added them to the interface because everything else seemed to have been added there.

I'd also tend to add them to `MockFileSystem` directly for consistency. I don't see a lot of value introducing another class for these methods. But I might very well be missing something :)

tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests/MockFileSystemTests.cs
rcdailey: No, this was accidental. I have restored it. Thank you for catching this!

src/TestableIO.System.IO.Abstractions.TestingHelpers/IMockFileDataAccessor.cs
rcdailey: I have removed all changes to this file.

rcdailey:
I've force-pushed the requested changes and I've rebased onto latest changes in `main`.

fgreinacher:
Looks good to me.

CI seems to be failing due to a .NET bug. We have to wait a bit until the fix is rolled out: https://github.com/dotnet/core/issues/8194#issuecomment-1445175415

Issue #943, opened 2023-02-09
`FileSystemStream` Issues


JasonBock:
**Describe the bug**
There's a documentation and API issue with [this type](https://github.com/TestableIO/System.IO.Abstractions/blob/main/src/TestableIO.System.IO.Abstractions/FileSystemStream.cs). There isn't a way to reproduce this, more that the design should be reviewed.

* The XML docs say `Wraps the <see cref="FileStream" />`, but the wrapped field, `_stream`, is of type `Stream`. There's no runtime check to ensure that the stream provided in the constructor is of type `FileStream`. This is confusing because the docs state one thing, but the implementation doesn't seem to match that.
* The `path` parameter for the constructor is defined as `string?`, but the constructor will throw an `ArgumentNullException` if `path` is null. This is confusing, as annotating a parameter as nullable means that the method can accept a null value. If `path` cannot be `null`, then remove the nullable annotation.

vbreuss:
Thanks for the hint @JasonBock.

The intention of this class is to wrap the `FileStream` but make it testable. So the interface wraps the `FileStream` and in the "normal" usage in production the provided stream should be a `FileStream`.
Only in the testing wrapper, we will replace it with a different stream for testing purposes.
I agree, that the XML documentation is short and could be improved, but the intention of it wrapping a `FileStream` is in essence correct.

With your second point I agree completely. The `path` parameter should be `string`.

A pull request would be welcome!

ZaffyPuck:
Hi, I am a college student looking trying to get more involved with open source and looking for a first-time "bug fix" and figured I could help out with this! I am a little confused with what XML docs you are referring to. Perhaps a link would help. As for the string issue, it should be `string` as opposed to `string?` because you want the user to pass data through the constructor. Is there anything I could help with as far as fixing the documentation or code? Let me know!

vbreuss:
@ZaffyPuck : The second point was already fixed in #956!
A [link to the file](https://github.com/TestableIO/System.IO.Abstractions/blob/main/src/TestableIO.System.IO.Abstractions/FileSystemStream.cs) is mentioned in the description.

PR #956, opened 2023-03-04
Make FileSystemStream path argument non-nullable

oreze:
Make "path" argument nullable in FileSystemStream constructor

Solving the second point #943


vbreuss (approved):
Looks good!

PR #960, opened 2023-03-10
fix: ensure `FlushAsync` behaves like `Flush`

### LINKED ISSUE #959: `FlushAsync` doesn't update file contents
**Describe the bug**
While `Flush` does update the contents of the file, allowing other streams to read it later, `FlushAsync` does not.

**To Reproduce**
```csharp
var file = fs.FileInfo.New("/file");
await using var first = file.OpenWrite();
await first.WriteAsync(new byte[] { 42 });
await first.FlushAsync();

await using var second = file.OpenRead();
var buffer = new byte[1];
Assert.Equal(1, await second.ReadAsync(buffer));
Assert.Equal(42, buffer[0]);
```

**Expected behavior**
Behave identically to the synchronous version, which does successfully pass.
```csharp
var file = fs.FileInfo.New("/file");
using var first = file.OpenWrite();
first.Write(new byte[] { 42 });
first.Flush();

using var second = file.OpenRead();
var buffer = new byte[1];
Assert.Equal(1, second.Read(buffer));
Assert.Equal(42, buffer[0]);
```

cryocz:
Added an overwrite for `Stream.FlushAsync` to the `MockFileStream` class, which ensures that the internal flush implementation is called. This way `FlushAsync` will correctly project the changes to the underlying `MockFile`, exactly like `Flush` does. 

Closes #959

### DIFF
--- a/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockFileStream.cs
+++ b/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockFileStream.cs
@@ -198,6 +198,13 @@ namespace System.IO.Abstractions.TestingHelpers
             InternalFlush();
         }
 
+        /// <inheritdoc />
+        public override Task FlushAsync(CancellationToken cancellationToken)
+        {
+            InternalFlush();
+            return Task.CompletedTask;
+        }
+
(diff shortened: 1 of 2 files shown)


fgreinacher (approved):
Looking great, thanks!

PR #964, opened 2023-04-09
feat: Implemented MockDirectory.CreateTempSubDirectory

gregington:
Implemented MockDirectory.CreateTempSubdirectory(). This implementation generates a random 6 character string with an optional user specified prefix and creates the directory in the temporary directory returned by `Path.GetTempPath()`.

The 6 random characters seems to be the algorithm used on my platform (Mac) by the System.IO implementation.

Notes on this PR
1. There's a potential temporary directory namespace of $62^6$ or about 56.8 billion. As we get closer to this limit, the likelihood of generating a duplicate increases as well as the likelihood of slowdowns of the random generation. At the extreme case, if all possible temporary directories are created, creating a random directory name will effectively become an infinite loop. I assume we don't need to handle conditions close to the limit.
2. `MockDirectory` now contains a `Random` instance to generate random temp directory names. `Random.Shared` can't be used as it was introduced in .NET 6 it it appears we are supporting earlier .NET versions.
3. `MockDirectory` is marked as `[Serializable]`, however the `Random` instance cannot be serialized. I have marked the `Random` instance as `[NonSerialized]` and created a new `Random` instance on deserializing. The state of the random number generator will be different when deserialized. I don't think this is an issue as there was no way to specify the seed of the `Random` instance when creating the `MockDirectory` instance, so there's no way to generate the same sequence of random directories again (besides a very slim chance).
4. I have had to add `<EnableUnsafeBinaryFormatterSerialization>` on the test project and `#pragma warning disable SYSLIB0011` on the deserialize method to allow testing that deserialization of a `MockDirectory` also instantiates a new random number generator. Deserialization is heavily deprecated and this is only on test classes, so I am assuming this is OK.

### DIFF
--- a/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectory.cs
+++ b/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectory.cs
@@ -32,6 +32,7 @@ namespace System.IO.Abstractions.TestingHelpers
                 mockFileDataAccessor ?? throw new ArgumentNullException(nameof(mockFileDataAccessor));
         }
 
+
         /// <inheritdoc />
         public override IDirectoryInfo CreateDirectory(string path)
         {
--- a/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectory.cs
+++ b/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectory.cs
@@ -2,7 +2,6 @@
 using System.Globalization;
 using System.Linq;
 using System.Runtime.Serialization;
-using System.Text;
 using System.Text.RegularExpressions;
 
 namespace System.IO.Abstractions.TestingHelpers
--- a/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectory.cs
+++ b/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectory.cs
@@ -1,7 +1,6 @@
 ﻿using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
-using System.Runtime.Serialization;
 using System.Text.RegularExpressions;
 
 namespace System.IO.Abstractions.TestingHelpers
--- a/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectory.cs
+++ b/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectory.cs
@@ -16,7 +16,6 @@ namespace System.IO.Abstractions.TestingHelpers
     {
         private readonly IMockFileDataAccessor mockFileDataAccessor;
         private string currentDirectory;
-        
         /// <inheritdoc />
         public MockDirectory(IMockFileDataAccessor mockFileDataAccessor, FileBase fileBase, string currentDirectory) :
             this(mockFileDataAccessor, currentDirectory)
--- a/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectory.cs
+++ b/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectory.cs
@@ -16,6 +16,7 @@ namespace System.IO.Abstractions.TestingHelpers
     {
         private readonly IMockFileDataAccessor mockFileDataAccessor;
         private string currentDirectory;
+
         /// <inheritdoc />
         public MockDirectory(IMockFileDataAccessor mockFileDataAccessor, FileBase fileBase, string currentDirectory) :
             this(mockFileDataAccessor, currentDirectory)
--- a/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectory.cs
+++ b/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectory.cs
@@ -133,15 +133,7 @@ namespace System.IO.Abstractions.TestingHelpers
             // Perform directory name generation is a loop, just in case the randomly generated name already exists.
             do
             {
-                var randomChars = new StringBuilder();
-                lock (random)
-                {
-                    for (var i = 0; i < 6; i++)
-                    {
-                        randomChars.Append(ValidRandomSubdirChars[random.Next(ValidRandomSubdirChars.Length)]);
-                    }
-                }
-                var randomDir = $"{prefix}{randomChars}";
+                var randomDir = $"{prefix}{Path.GetRandomFileName()}";
                 potentialTempDirectory = Path.Combine(Path.GetTempPath(), randomDir);
             } while (Exists(potentialTempDirectory));
 
--- a/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectory.cs
+++ b/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectory.cs
@@ -15,16 +15,9 @@ namespace System.IO.Abstractions.TestingHelpers
     [Serializable]
     public class MockDirectory : DirectoryBase
     {
-        private static readonly char[] ValidRandomSubdirChars = 
-            "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789".ToCharArray();
-
         private readonly IMockFileDataAccessor mockFileDataAccessor;
         private string currentDirectory;
         
-        // Random is not thread-safe, create our own instance and lock when accessing it.
-        [NonSerialized]
-        private Random random;
-
         /// <inheritdoc />
         public MockDirectory(IMockFileDataAccessor mockFileDataAccessor, FileBase fileBase, string currentDirectory) :
             this(mockFileDataAccessor, currentDirectory)
@@ -38,16 +31,7 @@ namespace System.IO.Abstractions.TestingHelpers
             this.currentDirectory = currentDirectory;
             this.mockFileDataAccessor =
                 mockFileDataAccessor ?? throw new ArgumentNullException(nameof(mockFileDataAccessor));
-            this.random = new Random();
         }
-
-        [OnDeserializing]
-        private void OnDeserializing(StreamingContext context)
-        {
-            // Random can't be serialized, create a new random when deserialized.
-            this.random = new Random();
-        }
-
 
         /// <inheritdoc />
         public override IDirectoryInfo CreateDirectory(string path)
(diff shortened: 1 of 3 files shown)


tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests/MockDirectoryTests.cs
```diff
@@ -813,6 +814,104 @@ public void MockDirectory_CreateDirectory_ShouldSucceedIfTryingToCreateUNCPathSh
...
+            var result = fileSystem.Directory.CreateTempSubdirectory(prefix);
+
+            // Assert
+            var name = Path.GetFileName(result.FullName);
+            for (var i = prefix.Length; i < name.Length; i++)
+            {
+                var c = name[i];
+                if (!char.IsUpper(c) && !char.IsLower(c) && !char.IsDigit(c))
+                {
+                    Assert.Fail($"Character '{c}' at position {i} of directoy {name} is not upper, lower or digit", c, i);
```
vbreuss, line 885: Typo `directoy` should be `directory`

tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests/MockDirectoryTests.cs
```diff
@@ -813,6 +814,104 @@ public void MockDirectory_CreateDirectory_ShouldSucceedIfTryingToCreateUNCPathSh
...
+
+            // Act
+            var result = fileSystem.Directory.CreateTempSubdirectory();
+
+            // Assert
+            var name = Path.GetFileName(result.FullName);
+            for (var i = 0; i < name.Length; i++)
+            {
+                var c = name[i];
+                if (!char.IsUpper(c) && !char.IsLower(c) && !char.IsDigit(c))
```
vbreuss, line 846: This check is incorrect (except on .NET Framework)!
I ran these tests an the real file system ([see here](https://github.com/Testably/Testably.Abstractions/pull/286)) and the result is e.g. `hnkjxa50.lqs`, so it contains a dot as well!

src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectory.cs
```diff
@@ -110,7 +127,25 @@ public override IFileSystemInfo CreateSymbolicLink(string path, string pathToTar
...
         public override IDirectoryInfo CreateTempSubdirectory(string prefix = null)
         {
-            throw CommonExceptions.NotImplemented();
+            prefix ??= "";
+            string potentialTempDirectory;
+
+            // Perform directory name generation is a loop, just in case the randomly generated name already exists.
+            do
+            {
+                var randomChars = new StringBuilder();
```
vbreuss, line 136: Can't you use `Path.GetRandomFileName()` to set the `potentialTempDirectory`?

vbreuss (changes requested):
@gregington : Thank you for working on this missing implementation!

tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests/MockDirectoryTests.cs
gregington: Removed this test, so this typo no longer exists. Test removed as I am now using `Path.GetRandomFileName()` so are now dependent on the .NET standard libraries to determine what is or isn't a valid character for a random file name.

src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectory.cs
gregington: Done. Also removed the tests that validated the character types of the random name as we now depend on the system library and whatever it's implementation decides is a valid name.

tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests/MockDirectoryTests.cs
gregington: That's interesting. I'm not sure how that was failing on the Windows platform as the name generation shouldn't have been generating `.` characters. The implementation has changed now, and this test was removed due to reliance on the `Path.GetRandomFileName()`.

gregington:
I changed one of the tests to hopefully work with MacOS and whether the `/private` directory prefix is returned for temporary directories.

I don't have access to a Windows platform so I couldn't reproduce failure with the `.` in the temporary filename. In any case, the implementation has now been changed to delegate to `Path.CreateTemporaryFileName()` so the generation of file names is no longer the concern of that method, so the tests regarding temporary file naming have been removed.

src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectory.cs
```diff
@@ -110,7 +110,17 @@ public override IFileSystemInfo CreateSymbolicLink(string path, string pathToTar
...
+            string potentialTempDirectory;
+
+            // Perform directory name generation is a loop, just in case the randomly generated name already exists.
+            do
+            {
+                var randomDir = $"{prefix}{Path.GetRandomFileName()}";
+                potentialTempDirectory = Path.Combine(Path.GetTempPath(), randomDir);
+            } while (Exists(potentialTempDirectory));
+
+            return CreateDirectoryInternal(Path.Combine(Path.GetTempPath(), potentialTempDirectory));
```
vbreuss, line 123: `potentialTempDirectory` already contains the `Path.GetTempPath()`, so you should not prefix it with the path to the temporary directory twice...

src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectory.cs
```diff
@@ -110,7 +110,17 @@ public override IFileSystemInfo CreateSymbolicLink(string path, string pathToTar
         /// <inheritdoc />
         public override IDirectoryInfo CreateTempSubdirectory(string prefix = null)
         {
-            throw CommonExceptions.NotImplemented();
+            prefix ??= "";
+            string potentialTempDirectory;
+
+            // Perform directory name generation is a loop, just in case the randomly generated name already exists.
```
vbreuss, line 116: Typo (`is a loop` => `in a loop`)

tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests.csproj
```diff
@@ -7,6 +7,7 @@
     <RootNamespace>System.IO.Abstractions.TestingHelpers.Tests</RootNamespace>
     <IsPackable>false</IsPackable>
     <IsTestable>true</IsTestable>
+    <EnableUnsafeBinaryFormatterSerialization>true</EnableUnsafeBinaryFormatterSerialization>
```
vbreuss, line 10: I think you no longer need this, as you removed the corresponding test, right?

tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests/MockDirectoryTests.cs
```diff
@@ -1,5 +1,6 @@
 using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.Serialization.Formatters.Binary;
```
vbreuss, line 3: I think you no longer need this, as you removed the corresponding test, right?

vbreuss (changes requested):

src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectory.cs
gregington: Good catch. Fixed.

src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectory.cs
gregington: Fixed.

tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests.csproj
gregington: Fixed.

tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests/MockDirectoryTests.cs
gregington: Fixed.

src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectory.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
+using System.Runtime.Serialization;
```
vbreuss, line 4: Is this using statement still required?

vbreuss (approved):
Looks good to me...

src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectory.cs
gregington: Sorry, missed this one too.

vbreuss (approved):

src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectory.cs
```diff
@@ -16,7 +16,7 @@ public class MockDirectory : DirectoryBase
     {
         private readonly IMockFileDataAccessor mockFileDataAccessor;
         private string currentDirectory;
-
+
```
fgreinacher, line 19:
```suggestion

```

src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectory.cs
```diff
@@ -16,7 +16,6 @@ public class MockDirectory : DirectoryBase
     {
         private readonly IMockFileDataAccessor mockFileDataAccessor;
         private string currentDirectory;
```
fgreinacher, line 18:
```suggestion
        private string currentDirectory;

```

src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectory.cs
```diff
@@ -32,7 +32,6 @@ public class MockDirectory : DirectoryBase
                 mockFileDataAccessor ?? throw new ArgumentNullException(nameof(mockFileDataAccessor));
         }
 
-
         /// <inheritdoc />
```
fgreinacher, line 35:
```suggestion

        /// <inheritdoc />
```

fgreinacher (approved):
LGTM, thanks!

PR #969, opened 2023-04-16
fix: throw IOException when creating a directory with a conflicting file

### LINKED ISSUE #968: MockFileSystem.Directory.CreateDirectory should throw IOException if conflicting file exists
**Describe the bug**
The MockFileSystem allows you to create a directory that conflicts with an existing file. The real System.IO.Directory.CreateDirectory method will throw an IOException if a conflicting file exists.

**To Reproduce**
```
fileSystem.File.WriteAllText(@"c:\test", "test file");
fileSystem.Directory.CreateDirectory(@"c:\test");
```

If you reverse the lines above and create the directory first, you get an UnauthorizedAccessException:
```
fileSystem.Directory.CreateDirectory(@"c:\test");
fileSystem.File.WriteAllText(@"c:\test", "test file");
```

**Expected behavior**
I would expect creation of a directory to fail when there is a conflicting file, just as creating a file fails if there is a conflicting directory.

**Additional context**
I was trying to write a negative test for a case when I'm trying to create a directory and a conflicting file already exists. This is how I discovered that the MockFileSystem behavior in this case doesn't match the real file system.

vbreuss:
Fixes #968:
When creating a directory, check if a file with the same name already exists. If so, throw an IOException.

### DIFF
--- a/System.IO.Abstractions.sln
+++ b/System.IO.Abstractions.sln
@@ -25,6 +25,7 @@ EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "_", "_", "{BBF7AD8D-5522-48C0-A906-00CBB72308A0}"
 	ProjectSection(SolutionItems) = preProject
 		Directory.Build.props = Directory.Build.props
+		global.json = global.json
 	EndProjectSection
 EndProject
 Global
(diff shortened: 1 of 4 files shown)


fgreinacher (approved):
Thanks!

PR #970, opened 2023-04-16
fix: always keep `FileAttributes.Directory` on directories

### LINKED ISSUE #957: Clearning the `Directory` flag causes a directory to be turned into a regular file
**Describe the bug**
Clearing the `FileAttributes.Directory` flag on a mocked directory will convert it into a file. Though it should be noted that `fs.Directory.Exists` will still report it as present.

**To Reproduce**
```csharp
var dir = fs.DirectoryInfo.New("dir");
dir.Attributes = 0;
fs.File.WriteAllText("/dir/file", "");
```
Will result in an exception, even though this would still work on a real IO.

vbreuss:
Fix #957:
Always keep `FileAttributes.Directory` on directories (and never on files).


fgreinacher (approved):
Nice gotcha, thanks!

PR #972, opened 2023-04-16
fix: update Name of `DirectoryInfo` when using `MoveTo`

### LINKED ISSUE #736: Using the MoveTo function to move a mocked directory doesn't update the DirectoryInfo information

vbreuss:
Fixes #736:
> In a unit test which consists on moving directory using the MoveTo function from IDirectoryInfo doesn't update the IDirectoryInfo entries and the moved directory info keeps the same initial values. However using the same function on the debug and release modes( FileSystem) does update the fields and properties of the DirectorytInfo

*Note: This should make #864 obsolete!*


fgreinacher (approved):
Wow, thanks for fixing all the things 🥇

PR #976, opened 2023-04-22
fix: null-handling in Wrap-Method in factories for `IDirectoryInfo` or `IFileInfo`

### LINKED ISSUE #975: FileInfoFactory and DirectoryInfoFactory Wrap implementations don't match the interface for nullability
**Describe the bug**
The implementations for `FileInfoFactory.Wrap` and `DirectoryInfoFactory.Wrap` require non-null references, while the interfaces are defined as `[return: NotNullIfNotNull("fileInfo")] IFileInfoFactory.Wrap(FileInfo? fileInfo)` and likewise for `IDirectoryInfoFactory.Wrap`.

**To Reproduce**
1. `new FileSystem().FileInfo.Wrap(null);`
2. The analyzer and compiler accept the syntax since `IFileInfoFactory.Wrap` accepts `FileInfo?` and returns `IFileInfo?`
3. An `ArgumentNullException` is thrown

**Expected behavior**
I expect to get a `null` `IFileInfo?` back if the input is `null`, and no exception thrown.

**Additional context**
Fixing this will help compatibility with [System.CommandLine](https://learn.microsoft.com/en-us/dotnet/standard/commandline/) where defining options as `FileInfo?` is a common practice, i.e.:

```csharp
var inputFileOption = new Option<FileInfo?>("--input");
var command = new Command("test") { inputFileOption };
command.SetHandler(
    (FileInfo? inputFile) => {
        var fileSystem = new FileSystem();
        var concreteCommand = new ConcreteCommand(fileSystem:fileSystem) {
            InputFile = fileSystem.FileInfo.Wrap(inputFile)
        }
        concreteCommand.Run();
    },
    inputFileOption
);
```

vbreuss:
Fixes #975:
When providing `null` as parameter to the implementations of `IFileInfoFactory` or `IDirectoryInfoFactory` return `null` instead of throwing an exception.

### DIFF
--- a/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectoryInfoFactory.cs
+++ b/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectoryInfoFactory.cs
@@ -33,6 +33,11 @@ namespace System.IO.Abstractions.TestingHelpers
         /// <inheritdoc />
         public IDirectoryInfo Wrap(DirectoryInfo directoryInfo)
         {
+            if (directoryInfo == null)
+            {
+                return null;
+            }
+
             return new MockDirectoryInfo(mockFileSystem, directoryInfo.Name);
         }
     }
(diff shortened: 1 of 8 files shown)


fgreinacher (approved):
Looks great, thanks for the quick fix!

PR #982, opened 2023-05-04
fix: Propagate null values in Wrap* methods

### LINKED ISSUE #981: IFileSystemInfo.ResolveLinkTarget throws exception instead of returning null if the path is not a link
**Describe the bug**
[System.IO.FileSystemInfo.ResolveLinkTarget](https://learn.microsoft.com/en-us/dotnet/api/system.io.filesysteminfo.resolvelinktarget?view=net-6.0#system-io-filesysteminfo-resolvelinktarget(system-boolean)) returns null when the directory is not a link.

IFileSystemInfo.ResolveLinkTarget ends up crashing with a null reference exception.

[ResolveLinkTarget implementation in DirectoryInfoWrapper.cs](https://github.com/TestableIO/System.IO.Abstractions/blob/main/src/TestableIO.System.IO.Abstractions.Wrappers/DirectoryInfoWrapper.cs#L43)
```C#
public override IFileSystemInfo ResolveLinkTarget(bool returnFinalTarget)
{
    return instance.ResolveLinkTarget(returnFinalTarget)
        .WrapFileSystemInfo(FileSystem);
}
```

[WrapFileSystemInfo implementation in Converters.cs](https://github.com/TestableIO/System.IO.Abstractions/blob/788d196afbf70dc7872dc8d72753e8052315fde6/src/TestableIO.System.IO.Abstractions.Wrappers/Converters.cs#L15)
```C#
private static FileSystemInfoBase WrapFileSystemInfo(IFileSystem fileSystem, FileSystemInfo item)
{
	if (item is FileInfo fileInfo)
	{
		return WrapFileInfo(fileSystem, fileInfo);
	}
	else if (item is DirectoryInfo directoryInfo)
	{
		return WrapDirectoryInfo(fileSystem, directoryInfo);
	}
	else
	{
		throw new NotImplementedException(string.Format(
			CultureInfo.InvariantCulture,
			"The type {0} is not recognized by the System.IO.Abstractions library.",
			item.GetType().AssemblyQualifiedName
		));
	}
}
```

The null reference we end up hitting is because `item` is null in WrapFileSystemInfo.

```
System.NullReferenceException
  HResult=0x80004003
  Message=Object reference not set to an instance of an object.
  Source=System.Private.CoreLib
  StackTrace:
   at System.Object.GetType()
   at System.IO.Abstractions.Converters.WrapFileSystemInfo(IFileSystem fileSystem, FileSystemInfo item) in /home/runner/work/System.IO.Abstractions/System.IO.Abstractions/src/TestableIO.System.IO.Abstractions.Wrappers/Converters.cs:line 42
```

**To Reproduce**
```C#
IFileSystem fs = new FileSystem();
IDirectoryInfo dirinfo = fs.Directory.CreateDirectory(@"C:\Some\Dir");
IFileSystemInfo? target = dirinfo.ResolveLinkTarget(true); //Crash
```

**Expected behavior**
ResolveLinkTarget should propagate the null return of the underlying System.IO API one way or another.
That API could do it directly, though it seems like WrapFileSystemInfo really ought to do the propagation itself.

Fix option 1:
```C#
private static FileSystemInfoBase WrapFileSystemInfo(IFileSystem fileSystem, FileSystemInfo item)
{
	if (item == null)
	{
		return null;
	}
	...
}
```

Fix option 2:
```C#
public override IFileSystemInfo ResolveLinkTarget(bool returnFinalTarget)
{
    return instance.ResolveLinkTarget(returnFinalTarget)
        ?.WrapFileSystemInfo(FileSystem);
}
```

zivarah:
When wrapping `FileSystemInfo`, `DirectoryInfo`, and `FileInfo`, null inputs will be correctly propagated.
Prior to this change, the wrappers for all three of these types would throw exceptions:

- `NullReferenceException` for `WrapFileSystemInfo`
- `ArgumentNullException` for `WrapDirectoryInfo`/`WrapFileInfo` (thrown by `DirectoryInfoWrapper`/`FileInfoWrapper`)

Basic unit tests were also added for these null cases.

Fixes #981

### DIFF
--- a/src/TestableIO.System.IO.Abstractions.Wrappers/Converters.cs
+++ b/src/TestableIO.System.IO.Abstractions.Wrappers/Converters.cs
@@ -29,6 +29,11 @@ namespace System.IO.Abstractions
 
         private static FileSystemInfoBase WrapFileSystemInfo(IFileSystem fileSystem, FileSystemInfo item)
         {
+            if (item is null)
+            {
+                return null;
+            }
+
             if (item is FileInfo fileInfo)
             {
                 return WrapFileInfo(fileSystem, fileInfo);
@@ -47,8 +52,10 @@ namespace System.IO.Abstractions
             }
         }
 
-        private static FileInfoBase WrapFileInfo(IFileSystem fileSystem, FileInfo f) => new FileInfoWrapper(fileSystem, f);
+        private static FileInfoBase WrapFileInfo(IFileSystem fileSystem, FileInfo f)
+            => f is null ? null : new FileInfoWrapper(fileSystem, f);
 
-        private static DirectoryInfoBase WrapDirectoryInfo(IFileSystem fileSystem, DirectoryInfo d) => new DirectoryInfoWrapper(fileSystem, d);
+        private static DirectoryInfoBase WrapDirectoryInfo(IFileSystem fileSystem, DirectoryInfo d)
+            => d is null ? null : new DirectoryInfoWrapper(fileSystem, d);
     }
 }
--- a/tests/TestableIO.System.IO.Abstractions.Wrappers.Tests/ConvertersTests.cs
+++ b/tests/TestableIO.System.IO.Abstractions.Wrappers.Tests/ConvertersTests.cs
@@ -1,5 +1,6 @@
 using System.Collections;
 using System.Collections.Generic;
+using System.Linq;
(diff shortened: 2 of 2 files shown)


fgreinacher (approved):
Thanks a lot!

PR #984, opened 2023-05-08
fix: Refresh FileInfo & DirectoryInfo after adding to MockFileSystem

rcdailey:
Modify the `IFileInfo` and `IDirectoryInfo` overloads of `AddFile()`, `AddEmptyFile()`, or `AddDirectory()` in `MockFileSystem`, to call `Refresh()` on those objects after the paths they refer to are created.


rcdailey:
Latest push has no changes; just rebased to pull in latest on `main`.

fgreinacher (approved):
Thanks, makes sense!

rcdailey:
@fgreinacher Thanks for merging this, I really appreciate it!

One thing I'm noticing, is that this doesn't cover all use cases. For example, if there are other `IFileInfo` objects pointing at the same file, they do not get refreshed. I know one of the driving philosophies for the mock filesystem is that it should model the behavior of .NET's file system objects. To that end, I'm not sure how this is officially handled with real `FileInfo` objects. If I were just "solving the problem", I'd probably go for some observable behavior where 'Info' objects subscribe to MockFileSystem for file events based on a particular path, that way all of those objects will get refreshed when a file is added. Again, however, if the official version of `FileSystem` does not do this, then I don't that would be an acceptable solution here, either.

Do you have any insight here you can share? This is somewhat a pain point I only experience in unit tests, having to manually refresh file info objects.

Thanks for any feedback you can share.

vbreuss:
@rcdailey :
The functionality on the real file system is a bit complicated:
- .NET Framework does not update the exists cache at all, so the `Exists` property gets set on the first read and never updated afterwards
- .NET Core evaluates the `Exists` cache also on first read, but updates it for when calling `Create()` for the current instance (but not for other instances, that point to the same location)

The following test is successful on the real file system on all platforms:
```csharp
	[Fact]
	public void Create_ShouldRefreshExistsCacheForCurrentItem_ExceptOnNetFramework()
	{
		string path = "file-path-to-check",
		IFileInfo sut1 = FileSystem.FileInfo.New(path);
		IFileInfo sut2 = FileSystem.FileInfo.New(path);
		IFileInfo sut3 = FileSystem.FileInfo.New(path);
		sut1.Exists.Should().BeFalse();
		sut2.Exists.Should().BeFalse();
		// Do not call Exists for `sut3`

		using FileSystemStream stream = sut1.Create();

		if (Test.IsNetFramework)
		{
			sut1.Exists.Should().BeFalse();
			sut2.Exists.Should().BeFalse();
			sut3.Exists.Should().BeFalse();
		}
		else
		{
			sut1.Exists.Should().BeTrue();
			sut2.Exists.Should().BeFalse();
			sut3.Exists.Should().BeTrue();
		}

		FileSystem.File.Exists(path).Should().BeTrue();
	}
```


*For reference see https://github.com/Testably/Testably.Abstractions/pull/313*

PR #985, opened 2023-05-09
feat: Add Flush(bool) to FileSystemStream.

gregington:
`FileSystemStream` was missing the implementation of `FileStream.Flush(bool)` which allows the caller to specify whether `Flush` should ensure all buffers are written to disk.

### DIFF
--- a/src/TestableIO.System.IO.Abstractions/FileSystemStream.cs
+++ b/src/TestableIO.System.IO.Abstractions/FileSystemStream.cs
@@ -128,6 +128,10 @@ namespace System.IO.Abstractions
         public override void Flush()
             => _stream.Flush();
 
+        /// <inheritDoc cref="FileStream.Flush(bool)" />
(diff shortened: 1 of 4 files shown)


vbreuss (approved):
Thanks a lot!
Looks good to me.

src/TestableIO.System.IO.Abstractions/FileSystemStream.cs
```diff
@@ -128,6 +128,10 @@ public override void EndWrite(IAsyncResult asyncResult)
         public override void Flush()
             => _stream.Flush();
 
+        /// <inheritDoc cref="FileStream.Flush(bool)" />
+        public virtual void Flush(bool flushToDisk)
+            => _stream.Flush();
```
fgreinacher, line 133: Should this pass on `flushToDisk`?

fgreinacher (approved):
Awesome, thanks a lot!

PR #988, opened 2023-05-15
fix: Use searchPattern argument in GetFiles method to complete the logic

AK1llv:
It should use the searchPattern argument to complete the logic.

I found it while working with MockDirectoryInfo.EnumerateFiles(searchPattern, enumerationOptions)
It didnt respect the searchPattern and my tests failed, so I decided to check the source.


AK1llv:
@fgreinacher there is also a problem with the Codacy Static Code Analysis

AK1llv:
image

fgreinacher (approved):
Nice catch, thanks for the fix!

Issue #773, opened 2021-11-29
Test Implementation for Path.GetRelativePath


ZacheryPD:
**Is your feature request related to a problem? Please describe.**
I am trying to implement a static site generator, and one of the tasks I need to do is adjust file paths handed to me to be relative to the root of the "static site" I am trying to create. This is something that Microsoft makes available via Path.GetRelativePath, and the normal implementation of it in System.IO.Abstractions seems to work great, but the Mock implementation is not being overwritten so the result is strange. 

**Describe the solution you'd like**
I think it should be possible to write some kind of mock implementation that would use the mock file system as reference.

**Describe alternatives you've considered**
I could probably also implement this as a specific mock for my project, but wanted to see if this was something the community had an interest in.

**Additional context**
If desired, my plan is to start prototyping this out today, but I can't make any promises on how fast it'll go.

[This](https://github.com/ZacheryPD/S3G/blob/a86759261d7d41e8f52e1efdd3a2d0077a6cf296/StaticSiteGenerator/Files/FileProcessingStrategies/FileProcessingStrategy.cs) is the example code that I need to "mock" the behavior of and [this](https://github.com/ZacheryPD/S3G/blob/feature/copy-files/StaticSiteGenerator.UnitTests/Files/FileProcessingStrategies/FileProcessingStrategyTests.cs) is a prototype test that currently behaves very strangely because of the GetRelativePath mock implementation.

ZacheryPD:
After spending some more time debugging and looking into this issue, I realize now that what I am dealing with is probably a strange edge case, and not as pervasive as I previously thought, which is great!  I'm still a bit confused by it, though.  This example test always fails:

```
[Fact]
public void Foo()
{
        var fs = new MockFileSystem();

        fs.AddDirectory("input");
        fs.AddDirectory("output");

        fs.AddFile("input/a.txt", "foo");

        var result = fs.Path.GetRelativePath("/input", "a.txt");

        Assert.Equal("a.txt", result);
}
```

But if you change "/input" to "input" in the `GetRelativePath` call, it all works as expected.  In reality, in this example, `/input` is what is created either way, but result will have the path to my test DLL when run.

fgreinacher:
Thanks for reporting this and sharing some test cases!

Which OS are you running the test on? `/input` is something different than `input` so I'd not really expected these tests to succeed. Can you share the test results here so that its easier to understand what's going on?

ZacheryPD:
[...]

PR #1012, opened 2023-07-30
feat: implement Path.GetRelativePath in MockFileSystem

vbreuss:
Override the `Path.GetRelativePath` implementation on mocked file systems and use the current directory to change the parameters to a full path relative to the current directory in the mocked file system.

Fixes:
- #773


fgreinacher (approved):
Thanks!

PR #1024, opened 2023-08-20
feat: add `ReparsePoint` attribute when creating a symbolic link

vbreuss:
Files or directories which are a symbolic link should have the `ReparsePoint` attribute set.


fgreinacher (approved):
LGTM, thanks a lot!

PR #1026, opened 2023-08-22
fix: incorrectly thrown exception in `Directory.CreateSymbolicLink`

vbreuss:
Remove incorrectly thrown exception when the target does not exist in Directory.CreateSymbolicLink. The real file system does not throw an exception in this case.
> [See here for a succeeding test against the real file system](https://github.com/Testably/Testably.Abstractions/blob/main/Tests/Testably.Abstractions.Tests/FileSystem/Directory/CreateSymbolicLinkTests.cs#L50).


fgreinacher (changes requested):
Thanks! Should we indicate a breaking change?

vbreuss:
@fgreinacher 
I consider this a bug, as our goal is to mimic the behaviour of the real file system. As we did not change any interface, I would not consider this a breaking change.

fgreinacher:
Technically it's still a breaking change as folks might have relied on the exception. But I agree that in this case it's pretty unlikely.

fgreinacher (approved):

PR #1030, opened 2023-08-22
fix: incorrect thrown exception when calling `Path.GetFullPath` with white-space

vbreuss:
Throw the correct `ArgumentException` instead of a `NullReferenceException` when calling Path.GetFullPath with a WhiteSpace string.


fgreinacher (approved):
Great catch 👍

PR #1032, opened 2023-08-25
feat: enable net8.0

vbreuss:
Support [.NET 8.0.](https://dotnet.microsoft.com/en-us/download/dotnet/8.0)


fgreinacher:
Thanks a lot, also for all the preparation work!

fgreinacher (approved):

PR #1034, opened 2023-08-25
refactor: prepare serialization infrastructure for .NET 8+

### LINKED ISSUE #1033: BinaryFormatter becomes obsolete -> Remove it with .NET 8.0
I tried executing the preview version of .NET 8.0 with System.IO.Abstractions and got the following test errors (see #1032)
image

The reason is, that the [`BinaryFormatter` becomes obsolete](https://github.com/dotnet/designs/blob/main/accepted/2020/better-obsoletion/binaryformatter-obsoletion.md).

I suggest to remove the support for Serialization under .NET8.0.

vbreuss:
This fixes #1033:
As [BinaryFormatter becomes obsolete](https://github.com/dotnet/designs/blob/main/accepted/2020/better-obsoletion/binaryformatter-obsoletion.md), remove the tests for serialization on .NET8.0 or greater. Also move the `Serialization` attribute behind a feature flag, which is currently always active, but will be disabled with .NET9.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -15,6 +15,7 @@
     <DefineConstants Condition="'$(TargetFramework)' == 'net7.0' OR '$(TargetFramework)' == 'net6.0' OR '$(TargetFramework)' == 'net5.0'">$(DefineConstants);FEATURE_FILE_MOVE_WITH_OVERWRITE;FEATURE_SUPPORTED_OS_ATTRIBUTE;FEATURE_FILE_SYSTEM_WATCHER_FILTERS;FEATURE_ENDS_IN_DIRECTORY_SEPARATOR;FEATURE_PATH_JOIN_WITH_PARAMS;FEATURE_PATH_JOIN_WITH_FOUR_PATHS</DefineConstants>
     <DefineConstants Condition="'$(TargetFramework)' == 'net7.0' OR '$(TargetFramework)' == 'net6.0'">$(DefineConstants);FEATURE_FILE_SYSTEM_INFO_LINK_TARGET;FEATURE_CREATE_SYMBOLIC_LINK;FEATURE_FILESTREAM_OPTIONS</DefineConstants>
     <DefineConstants Condition="'$(TargetFramework)' == 'net7.0'">$(DefineConstants);FEATURE_PATH_EXISTS;FEATURE_FILE_SYSTEM_WATCHER_WAIT_WITH_TIMESPAN;FEATURE_FILE_ATTRIBUTES_VIA_HANDLE;FEATURE_CREATE_TEMP_SUBDIRECTORY;FEATURE_READ_LINES_ASYNC;FEATURE_UNIX_FILE_MODE</DefineConstants>
+    <DefineConstants>$(DefineConstants);FEATURE_SERIALIZABLE</DefineConstants>
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Nerdbank.GitVersioning" Version="3.6.133">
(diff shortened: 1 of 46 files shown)


fgreinacher:
Just a quick answer while I'm on vacation 😎

Reading https://github.com/dotnet/designs/blob/main/accepted/2020/better-obsoletion/binaryformatter-obsoletion.md I don't see `SerializableAttribute` being deprecated for .NET 8. What's the reason that we have to exclude it?

vbreuss:
> Just a quick answer while I'm on vacation 😎
> 
> Reading https://github.com/dotnet/designs/blob/main/accepted/2020/better-obsoletion/binaryformatter-obsoletion.md I don't see `SerializableAttribute` being deprecated for .NET 8. What's the reason that we have to exclude it?

@fgreinacher 
Probably we don't have to exclude it, but as [this attribute only applies when using the BinaryFormatter or SoapFormatter](https://stackoverflow.com/a/50991360), which both are obsolete in .NET8 I would  also remove  the attribute (see also [here](https://stackoverflow.com/questions/2982376/why-is-serializable-attribute-required-for-an-object-to-be-serialized)).

I can of course keep the attributes also for .NET8, but there is no way to test them anymore!

fgreinacher:
@vbreuss If I understand the upstream documentation (especially https://github.com/dotnet/designs/blob/main/accepted/2020/better-obsoletion/binaryformatter-obsoletion.md#binaryformatter-disabled-by-default-across-all-project-types-net-8) right, application can still opt in to binary serialization via the `EnableUnsafeBinaryFormatterSerialization` switch. As this library can be used in such scenarios I think we should not remove the attributes for .NET 8 and instead set the switch in our test project. WDYT?

vbreuss:
> If I understand the upstream documentation (especially https://github.com/dotnet/designs/blob/main/accepted/2020/better-obsoletion/binaryformatter-obsoletion.md#binaryformatter-disabled-by-default-across-all-project-types-net-8) right, applications can still opt in to binary serialization via the `EnableUnsafeBinaryFormatterSerialization` switch. As this library can be used in such scenarios I think we should not remove the attributes for .NET 8 and instead set the switch in our test project. 

@fgreinacher 
I interpreted this
[image](https://github.com/dotnet/designs/blob/main/accepted/2020/better-obsoletion/binaryformatter-obsoletion.md#net-8-nov-2023)
to mean, that the attribute should no longer be used in .NET 8 or newer for custom types.
As an upgrade of the .NET version can always result in a breaking change, I preferred to remove the `SerializableAttribute` altogether.

However, I see your point that the workaround via `EnableUnsafeBinaryFormatterSerialization` is not working anymore.
I can update the pull request, to keep the attributes. However, I don't think it necessary to explicitely update the tests, and instead I would just disable the serialization tests for .NET 8 or newer.

The attribute would then be removed with .NET 9 when the `BinaryFormatter` infrastructure gets removed, right?
[image](
image
)

fgreinacher:
> However, I don't think it necessary to explicitely update the tests, and instead I would just disable the serialization tests for .NET 8 or newer.

I think we should keep them because we should still support the legacy serialization if applications explicitly opt-in.

vbreuss:
@fgreinacher 
I moved the `SerializableAttribute` behind a feature flag, that is currently always active. This way it will be easy to disable for a specific framework version in future.
The tests for serialization will only run pre-.NET8.

I hope this is okay with you :-)

fgreinacher:
> I moved the SerializableAttribute behind a feature flag, that is currently always active. This way it will be easy to disable for a specific framework version in future.

That's cool, thanks a lot!

> The tests for serialization will only run pre-.NET8.

As mentioned above I'd prefer to add not exclude them and instead add this the test csprojs:

```xml
  <PropertyGroup>
    <!--
      Allow deprecated binary formatter functionality on .NET 8 so that we can test it
    -->
    <EnableUnsafeBinaryFormatterSerialization Condition="'$(TargetFramework)' == 'net8.0'">true</EnableUnsafeBinaryFormatterSerialization>
  </PropertyGroup>
```

fgreinacher (approved):
Thanks again @vbreuss 🙇

PR #1046, opened 2023-10-18
fix: prefix bug in MockDirectory.EnumerateDirectories (#815)

### LINKED ISSUE #815: MockFileSystem.Directory.GetDirectories returns absolute paths instead of relative paths

Phoenox:
Fixes #815 

`System.IO.Directory.EnumerateDirectories(path)` guarantees that returned paths are always prefixed with the value provided in the `path` parameter. However, `MockDirectory.EnumerateDirectories` always returned absolute paths, leading to an inconsistent behavior (e.g. when using relative paths).

We now check returned paths from `MockDirectory.EnumerateDirectories` and provide the proper prefix path.

### DIFF
--- a/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectory.cs
+++ b/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectory.cs
@@ -651,7 +651,9 @@ namespace System.IO.Abstractions.TestingHelpers
         private string FixPrefix(string path, string originalPath)
         {
             var normalizedOriginalPath = mockFileDataAccessor.Path.GetFullPath(originalPath);
-            return originalPath + path.Substring(normalizedOriginalPath.Length);
+            var pathWithoutOriginalPath = path.Substring(normalizedOriginalPath.Length)
+                .TrimStart(mockFileDataAccessor.Path.DirectorySeparatorChar);
+            return mockFileDataAccessor.Path.Combine(originalPath, pathWithoutOriginalPath);
         }
         
 #if FEATURE_ENUMERATION_OPTIONS
--- a/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectory.cs
+++ b/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectory.cs
@@ -627,16 +627,12 @@ namespace System.IO.Abstractions.TestingHelpers
         /// <inheritdoc />
         public override IEnumerable<string> EnumerateDirectories(string path)
         {
-            mockFileDataAccessor.PathVerifier.IsLegalAbsoluteOrRelative(path, "path");
-
             return EnumerateDirectories(path, "*");
         }
 
         /// <inheritdoc />
         public override IEnumerable<string> EnumerateDirectories(string path, string searchPattern)
         {
-            mockFileDataAccessor.PathVerifier.IsLegalAbsoluteOrRelative(path, "path");
-
             return EnumerateDirectories(path, searchPattern, SearchOption.TopDirectoryOnly);
         }
 
@@ -644,22 +640,26 @@ namespace System.IO.Abstractions.TestingHelpers
         public override IEnumerable<string> EnumerateDirectories(string path, string searchPattern, SearchOption searchOption)
         {
             mockFileDataAccessor.PathVerifier.IsLegalAbsoluteOrRelative(path, "path");
+            var originalPath = path;
             path = path.TrimSlashes();
             path = mockFileDataAccessor.Path.GetFullPath(path);
             return GetFilesInternal(mockFileDataAccessor.AllDirectories, path, searchPattern, searchOption)
-                .Where(p => !mockFileDataAccessor.StringOperations.Equals(p, path));
+                .Where(p => !mockFileDataAccessor.StringOperations.Equals(p, path))
+                .Select(p => FixPrefix(p, originalPath));
         }
-
+        
+        private string FixPrefix(string path, string originalPath)
+        {
+            var normalizedOriginalPath = mockFileDataAccessor.Path.GetFullPath(originalPath);
+            return originalPath + path.Substring(normalizedOriginalPath.Length);
+        }
+        
 #if FEATURE_ENUMERATION_OPTIONS
         /// <inheritdoc />
         public override IEnumerable<string> EnumerateDirectories(string path, string searchPattern, EnumerationOptions enumerationOptions)
         {
             var searchOption = enumerationOptions.RecurseSubdirectories ? SearchOption.AllDirectories : SearchOption.TopDirectoryOnly;
-            mockFileDataAccessor.PathVerifier.IsLegalAbsoluteOrRelative(path, "path");
-            path = path.TrimSlashes();
-            path = mockFileDataAccessor.Path.GetFullPath(path);
-            return GetFilesInternal(mockFileDataAccessor.AllDirectories, path, searchPattern, searchOption)
-                .Where(p => !mockFileDataAccessor.StringOperations.Equals(p, path));
+            return EnumerateDirectories(path, searchPattern, searchOption);
         }
 #endif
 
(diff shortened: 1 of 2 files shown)


tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests/MockDirectoryTests.cs
```diff
@@ -1486,6 +1486,22 @@ public void MockDirectory_EnumerateDirectories_ShouldThrowWhenPathIsNotMocked()
             // Assert
             Assert.Throws<DirectoryNotFoundException>(action);
         }
+        
+        
+        [TestCase("Folder", "Folder/SubFolder")]
+        [TestCase("/Folder", "/Folder/SubFolder")]
+        [TestCase("Folder/", "Folder/SubFolder")]
```
vbreuss, line 1493: In the real file system this returns an empty collection and not one that contains "Folder/SubFolder".

tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests/MockDirectoryTests.cs
```diff
@@ -1486,6 +1486,22 @@ public void MockDirectory_EnumerateDirectories_ShouldThrowWhenPathIsNotMocked()
             // Assert
             Assert.Throws<DirectoryNotFoundException>(action);
         }
+        
+        
+        [TestCase("Folder", "Folder/SubFolder")]
+        [TestCase("/Folder", "/Folder/SubFolder")]
```
vbreuss, line 1492: These test cases will only run under Linux and not under Windows.

vbreuss (changes requested):
@Phoenox:
Thanks for this correction, but please make sure, that it also works on Windows, because this library should target all platforms :-)

tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests/MockDirectoryTests.cs
Phoenox: Sorry, but I cannot reproduce your finding. I've checked both on GNU/Linux and Windows: Neither OS cares about the trailing path separator.

tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests/MockDirectoryTests.cs
Phoenox: Yeah, I removed that one, since I have no idea how i can handle the root directory case in a platform-agnostic way.

tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests/MockDirectoryTests.cs
```diff
@@ -1486,6 +1486,21 @@ public void MockDirectory_EnumerateDirectories_ShouldThrowWhenPathIsNotMocked()
             // Assert
             Assert.Throws<DirectoryNotFoundException>(action);
         }
+        
+        
+        [TestCase("Folder", @"Folder\SubFolder")]
+        [TestCase(@"Folder\", @"Folder\SubFolder")]
```
vbreuss, line 1492: I ran this test case on the real file system and it returned an empty collection.

src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectory.cs
```diff
@@ -627,39 +627,39 @@ public override void SetLastWriteTimeUtc(string path, DateTime lastWriteTimeUtc)
...
-                .Where(p => !mockFileDataAccessor.StringOperations.Equals(p, path));
+                .Where(p => !mockFileDataAccessor.StringOperations.Equals(p, path))
+                .Select(p => FixPrefix(p, originalPath));
         }
-
+        
+        private string FixPrefix(string path, string originalPath)
+        {
+            var normalizedOriginalPath = mockFileDataAccessor.Path.GetFullPath(originalPath);
+            return originalPath + path.Substring(normalizedOriginalPath.Length);
```
vbreuss, line 654: Will you always have a path separator here?
To be sure, you could use `Path.Combine` for concatinating the `originalPath` and the remaining path.

vbreuss (changes requested):

tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests/MockDirectoryTests.cs
vbreuss: I created a repro-branch here: https://github.com/Testably/Testably.Abstractions/pull/415/files

vbreuss: Sorry, that was my error. Looks good :-)

src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectory.cs
Phoenox: I'm pretty new to the domain, but it seems to me that we should always have a path separator here.
The code with the `Path.Combine` would look like this:
```
var pathWithoutOriginalPath = path.Substring(normalizedOriginalPath.Length)
    .TrimStart(mockFileDataAccessor.Path.DirectorySeparatorChar);
return mockFileDataAccessor.Path.Combine(originalPath, pathWithoutOriginalPath);
```
Do you think this would be an improvement?

vbreuss: I think, it would be a safer choice. When I look only at the `FixPrefix` method, there is no way to ensure that there is exactly one `DirectorySeparatorChar` either at the end of `originalPath` or at the beginning of `path.Substring(normalizedOriginalPath.Length)`. With your suggestion above, it would make sure, that this is always the case.

Phoenox: Alright, changed in 617eca2

vbreuss (approved):
Looks good 😊
Thanks @Phoenox

fgreinacher:
Thanks for the fix @Phoenox, please have a look at the failing ubuntu and macOS tests :bow:

fgreinacher (approved):
Sorry @Phoenox, this slipped through my notifications. Looks great, thanks a ton for the fix!

PR #1065, opened 2023-12-06
refactor: Convert Classic Assert to Constraint Model

vbreuss:
Convert NUnit asserts to the constraint model, to simplify the [migration to NUnit v4](https://docs.nunit.org/articles/nunit/release-notes/Nunit4.0-MigrationGuide.html)


fgreinacher (approved):
Thanks!

(I assume you automated this, so I did not look into each and every change 😄)

PR #1079, opened 2024-01-22
fix: use correct path for wrapped `FileInfo` and `DirectoryInfo`

### LINKED ISSUE #1049: IDirectoryInfoFactory.Wrap loses part of absolute path when used with MockFileSystem
**Describe the bug**
If you provide the MockFileSystem's DirectoryInfo.Wrap method with a DirectoryInfo that has an absolute path, the resulting IDirectoryInfo has the wrong file path.

**To Reproduce**
```cs
using Xunit;
using System.IO.Abstractions;

[Fact]
public void Test()
{
    var directoryInfo = new DirectoryInfo(@"C:\subfolder\file");
    var wrappedDirectoryInfo = fs.DirectoryInfo.Wrap(directoryInfo);

    Assert.Equal(directoryInfo.FullName, wrappedDirectoryInfo.FullName); // wrappedDirectoryInfo returns "C:\file"
}
```

**Expected behaviour**
The test above is expected to pass.

**Actual behaviour**
```
Assert.Equal() Failure
             ↓ (pos 3)
Expected: C:\subfolder\file
Actual:   C:\file
             ↑ (pos 3)
```

**Additional context**
Tested on Windows 10, .Net 7, System.IO.Abstractions version 19.2.69

vbreuss:
Use the full path when wrapping a `FileInfo` or `DirectoryInfo` to preserve the relevant information.

*Fixes #1049*

### DIFF
--- a/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectoryInfoFactory.cs
+++ b/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectoryInfoFactory.cs
@@ -40,7 +40,7 @@ namespace System.IO.Abstractions.TestingHelpers
                 return null;
             }
 
-            return new MockDirectoryInfo(mockFileSystem, directoryInfo.Name);
+            return new MockDirectoryInfo(mockFileSystem, directoryInfo.FullName);
         }
     }
 }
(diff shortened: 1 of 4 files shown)


fgreinacher (approved):
Thanks @vbreuss, LGTM!

PR #1135, opened 2024-07-24
fix: Throw correct exception if opening read-only file for writing

WestRyanK:
Mock classes did not match .net's behavior of throwing `UnauthorizedAccessException` when opening read-only files with `FileAccess.Write`, making it impossible to test those scenarios.

This commit fixes the mock classes so that the exception will be thrown correctly.


vbreuss (approved):
Thanks, @WestRyanK for reporting and fixing this error.
I just slightly adapted the PR title, otherwise the pull request looks good to me and will be released shortly!

PR #1160, opened 2024-11-05
feat: added support for GetUnixFileMode and SetUnixFileMode

tomwis:
* only overload with `string path` is added, overload with `SafeFileHandle fileHandle` is not added


vbreuss:
@tomwis, Thanks for providing this pull request. I will have a look at your PR in the next couple of days...

vbreuss (approved):
Looks good to me 👍

Issue #1173, opened 2024-12-10
FileSystem property of DirectoryInfo is not what it seems


montoner0:
**Is your feature request related to a problem? Please describe.**
On multiple occasions I've misinterpret the `DirectoryInfo.FileSystem` property as related to this exact DirectoryInfo object which isn't true at all.

**Describe the solution you'd like**
It would be nice to have some clear statement in the description that they aren't related (yes, I know that's derived from IFileSystemInfo interface but still). Something like "The property isn't related to its parent object".

vbreuss:
I am not sure I get what you mean. The description of the `FileSystem` property is:
```xml
/// <summary>
/// Exposes the underlying filesystem implementation. This is useful for implementing extension methods.
/// </summary>
```

What is your interpretation of the file system being "related to the parent object"?

montoner0:
I mean the property could be considered as acting within a scope of the object it belongs. 
As if `DirectoryInfo.FileSystem.File.Create(filename)` would create a file in the directory of DirectoryInfo.

vbreuss:
Ah, I see your point. You would expect the current directory to be changed, when you access the file system from a `IDirectoryInfo` instance.
Unfortunately, this is not really possible, because we always return the same instance and changing the (global) current directory on accessing a property would lead to very confusing behavior (especially, if the accessing of the property is hidden in e.g. an extension method).

Do you have a suggestion on how to improve the documentation to make this more clear?
Feel free to create a pull request...

PR #1175, opened 2024-12-14
docs: clarify meaning of the FileSystem property

montoner0:
for #1173

### DIFF
--- a/src/TestableIO.System.IO.Abstractions/IFileSystemInfo.cs
+++ b/src/TestableIO.System.IO.Abstractions/IFileSystemInfo.cs
@@ -8,6 +8,9 @@ namespace System.IO.Abstractions
         /// <summary>
         /// Exposes the underlying filesystem implementation. This is useful for implementing extension methods.
         /// </summary>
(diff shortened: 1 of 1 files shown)


src/TestableIO.System.IO.Abstractions/IFileSystemInfo.cs
```diff
@@ -8,6 +8,9 @@ public interface IFileSystemInfo
         /// <summary>
         /// Exposes the underlying filesystem implementation. This is useful for implementing extension methods.
         /// </summary>
+        /// <remarks>
+        /// The property is always a global object related to a global current directory.
```
vbreuss, line 12: I think it would be clearer if you say "**the** global current directory" (as there is always only one).
What do you think?

montoner0: I'm not a native speaker so any suggestions about wording are welcomed! I'll make the change.

vbreuss (approved):
Thank you for the change. Looks good!

PR #1177, opened 2024-12-16
feat: Add FileVersionInfo support

t4m45:
Added a wrapper for the immutable `FileVersionInfo`.

For testing, the `IFileVersionInfo` can be stored in the `MockFileData` as a property.
Since testing it is rarely neccessary, I didn't add a constructor for it but it's a mutable property, so it can be set anytime after initialization.
A `MockFileVersionInfo` has also been added which is `mutable`.

In a normal scenario, the `FileVersionInfo` can be accessed through the `FileVersionInfo` property of `IFileInfo`.

The `MockFileSystem`'s `AddFile` method will initialize this version if it's null, and the values will be the default (Only it's FileName will be set).

Since the `System.IO.FileInfo` does not contain a `FileVersionInfo` property, I excluded it from the `ApiParityTests`

PS.: Sorry, seems like were some auto formats on my side which removed some extra spaces and the Visual Studio's Diff viewer did not show these before commit. I'm not really familiar with github, thats why I didn't try to revert them after commit.

### DIFF
--- a/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockFileVersionInfo.cs
+++ b/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockFileVersionInfo.cs
@@ -68,12 +68,6 @@ namespace System.IO.Abstractions.TestingHelpers
             ProductPrivatePart = parsedProductVersion.PrivatePart;
         }
 
-        /// <inheritdoc />
-        public MockFileVersionInfo(string fileName)
-        {
-            FileName = fileName;
-        }
-
         /// <inheritdoc/>
         public override string FileName { get; }
 
--- a/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockFileVersionInfo.cs
+++ b/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockFileVersionInfo.cs
@@ -1,5 +1,4 @@
 ﻿using System.Diagnostics;
-using System.Runtime.InteropServices;
 using System.Text;
 
 namespace System.IO.Abstractions.TestingHelpers
--- a/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockFileVersionInfo.cs
+++ b/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockFileVersionInfo.cs
@@ -12,43 +12,30 @@ namespace System.IO.Abstractions.TestingHelpers
     {
         /// <inheritdoc />
         public MockFileVersionInfo(
-            [Optional] string comments,
-            [Optional] string companyName,
-            [Optional] int fileBuildPart,
-            [Optional] string fileDescription,
-            [Optional] int fileMajorPart,
-            [Optional] int fileMinorPart,
-            [Optional] string fileName,
-            [Optional] int filePrivatePart,
-            [Optional] string fileVersion,
-            [Optional] string internalName,
-            [Optional] bool isDebug,
-            [Optional] bool isPatched,
-            [Optional] bool isPrivateBuild,
-            [Optional] bool isPreRelease,
-            [Optional] bool isSpecialBuild,
-            [Optional] string language,
-            [Optional] string legalCopyright,
-            [Optional] string legalTrademarks,
-            [Optional] string originalFilename,
-            [Optional] string privateBuild,
-            [Optional] int productBuildPart,
-            [Optional] int productMajorPart,
-            [Optional] int productMinorPart,
-            [Optional] string productName,
-            [Optional] int productPrivatePart,
-            [Optional] string productVersion,
-            [Optional] string specialBuild)
+            string comments = null,
+            string companyName = null,
+            string fileDescription = null,
+            string fileName = null,
+            string fileVersion = null,
+            string internalName = null,
+            bool isDebug = false,
+            bool isPatched = false,
+            bool isPrivateBuild = false,
+            bool isPreRelease = false,
+            bool isSpecialBuild = false,
+            string language = null,
+            string legalCopyright = null,
+            string legalTrademarks = null,
+            string originalFilename = null,
+            string privateBuild = null,
+            string productName = null,
+            string productVersion = null,
+            string specialBuild = null)
         {
             Comments = comments;
             CompanyName = companyName;
-            FileBuildPart = fileBuildPart;
             FileDescription = fileDescription;
-            FileMajorPart = fileMajorPart;
-            FileMinorPart = fileMinorPart;
             FileName = fileName;
-            FilePrivatePart = filePrivatePart;
-            FileVersion = fileVersion;
             InternalName = internalName;
             IsDebug = isDebug;
             IsPatched = isPatched;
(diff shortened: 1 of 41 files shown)


vbreuss (changes requested):
Thanks for suggesting this change, @t4m45!

While I do like the idea of adding support for [`FileVersionInfo`](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.fileversioninfo?view=net-8.0), I don't like adding it to the `IFileInfo` interface, as this is different to the public API.

Instead I would suggest adding a `IFileInfoFactory` to the `IFileSystem` with a static method `GetVersionInfo(string)` which returns an instance of `IFileInfo`. This way you could use it as follows:
```csharp
// Instead of
FileVersionInfo version = FileVersionInfo.GetVersionInfo("my-library.dll");
// you can use
IFileVersionInfo version = fileSystem.FileVersionInfo.GetVersionInfo("my-library.dll");
```

This works the same way as the existing API. You could then adapt the parity tests, to also verify, that the interface `IFileVersionInfo` matches `FileVersionInfo`.

Would you be willing to adapt your pull request as suggested?

t4m45:
Hello @vbreuss!

Just for clarity, in the sentence "I would suggest adding a `IFileInfoFactory` to the IFileSystem with a static method GetVersionInfo(string) which returns an instance of `IFileInfo`"

Did you mean 'IFileVersionInfoFactory' and 'IFileVersionInfo' instead of 'IFileInfoFactory' and 'IFileInfo'?

t4m45:
I implemented the factory you asked for, and removed the `FileVersionInfo` from `IFileInfo`.

vbreuss (changes requested):
Even though I have some comments, I am impressed by the overall approach; great work!

tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests/MockFileVersionInfoFactoryTests.cs
```diff
@@ -0,0 +1,49 @@
...
+        [Test]
+        public void MockFileVersionInfoFactory_GetVersionInfo_ShouldThrowFileNotFoundExceptionIfFileDoesNotExist()
+        {
+            // Arrange
+            var fileSystem = new MockFileSystem(new Dictionary<string, MockFileData>
+            {
+                { @"c:\a.txt", new MockFileData("Demo text content") },
+                { @"c:\a\b\c.txt", new MockFileData("Demo text content") },
+            });
+            _ = new MockFileVersionInfoFactory(fileSystem); // file system injection
```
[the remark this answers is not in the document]
t4m45: Do you mean I should've inlined the delegate into the Assert.Throw?

tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests/MockFileVersionInfoFactoryTests.cs
```diff
@@ -0,0 +1,49 @@
...
+        [Test]
+        public void MockFileVersionInfoFactory_GetVersionInfo_ShouldReturnTheFileVersionInfoOfTheMockFileData()
+        {
+            // Arrange
+            var fileVersionInfo = new MockFileVersionInfo();
+            var fileSystem = new MockFileSystem(new Dictionary<string, MockFileData>
+            {
+                { @"c:\a.txt", new MockFileData("Demo text content") { FileVersionInfo = fileVersionInfo } }
+            });
+            _ = new MockFileVersionInfoFactory(fileSystem); // file system injection
```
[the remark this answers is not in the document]
t4m45: It should. I messed this up.. It was a long day

tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests/MockFileSystemTests.cs
```diff
@@ -286,6 +286,21 @@ public void MockFileSystem_AddFile_ShouldMatchCapitalization_PartialMatch_Furthe
...
+        public void MockFileSystem_AddFile_InitializesMockFileDataFileVersionInfoIfNull()
+        {
+            // Arrange
+            var fileSystem = new MockFileSystem();
+
+            // Act
+            fileSystem.AddFile(XFS.Path(@"C:\file.txt"), string.Empty);
+
+            // Assert
+            IFileVersionInfo fileVersionInfo = fileSystem.GetFile(XFS.Path(@"C:\file.txt")).FileVersionInfo;
```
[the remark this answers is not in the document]
t4m45: Nice catch!

src/TestableIO.System.IO.Abstractions.TestingHelpers/MockFileVersionInfo.cs
```diff
@@ -0,0 +1,176 @@
...
+{
+    /// <inheritdoc />
+#if FEATURE_SERIALIZABLE
+    [Serializable]
+#endif
+    public class MockFileVersionInfo : FileVersionInfoBase
+    {
+        /// <inheritdoc />
+        public MockFileVersionInfo(
+            [Optional] string comments,
```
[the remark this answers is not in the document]
t4m45: My bad, I didn't check if you use the Optional attribute. I'll change that.

About the revision.. if you create a new file in the OS, the default value of the revision is 0, but I'll change it as you wish.

src/TestableIO.System.IO.Abstractions.TestingHelpers/MockFileVersionInfo.cs
```diff
@@ -0,0 +1,176 @@
...
+            [Optional] string companyName,
+            [Optional] int fileBuildPart,
+            [Optional] string fileDescription,
+            [Optional] int fileMajorPart,
+            [Optional] int fileMinorPart,
+            [Optional] string fileName,
+            [Optional] int filePrivatePart,
+            [Optional] string fileVersion,
```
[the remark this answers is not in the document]
t4m45: Good idea. Then it would work the same as an AssemblyInfo.

src/TestableIO.System.IO.Abstractions.TestingHelpers/MockFileVersionInfo.cs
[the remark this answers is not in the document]
vbreuss: The revision is named here "PrivatePart" (it's the fourth part of the version).

t4m45:
When it comes to the `MockFileVersionInfoFactory`, I'm not sure how to fix the issue reported by Codacy: "_Remove this assignment of 'mockFileSystem' or initialize it statically._"

If you're okay with it, I can remove the constructor, create a static method called `SetFileDataAccessor`, and inject it that way.
But if you have another idea, I'm all ears.

tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests/MockFileVersionInfoFactoryTests.cs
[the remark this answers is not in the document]
t4m45: After enough sleep, I realized what you meant..
The unneccessary constructor call has already been removed in a previous commit so this is also fixed now.

vbreuss:
> When it comes to the `MockFileVersionInfoFactory`, I'm not sure how to fix the issue reported by Codacy: "_Remove this assignment of 'mockFileSystem' or initialize it statically._"
> 
> If you're okay with it, I can remove the constructor, create a static method called `SetFileDataAccessor`, and inject it that way. But if you have another idea, I'm all ears.

You should not use a static field in the `MockFileVersionInfoFactor`.
If you make the following changes, it should build without errors:
image

*I will have a more in-depth look at the pull request after christmas...*

t4m45:
> You should not use a static field in the `MockFileVersionInfoFactory`. If you make the following changes, it should build without errors: image

Well, I know that. The only reason I made the method static is because thats what you asked for.

> Instead I would suggest adding a `IFileInfoFactory` to the `IFileSystem` with a static method `GetVersionInfo(string)` which returns an instance of `IFileInfo`.

But I will change it then.

vbreuss:
> Well, I know that. The only reason I made the method static is because thats what you asked for.
> 

Sorry, that was misleading by me. I meant, that you should mimick the static method in the System.Diagnostics namespace the same way we mimick the static `File` methods (by instance methods on a factory class)...

src/TestableIO.System.IO.Abstractions.TestingHelpers/MockFileVersionInfo.cs
```diff
@@ -0,0 +1,207 @@
...
+namespace System.IO.Abstractions.TestingHelpers
+{
+    /// <inheritdoc />
+#if FEATURE_SERIALIZABLE
+    [Serializable]
+#endif
+    public class MockFileVersionInfo : FileVersionInfoBase
+    {
+        /// <inheritdoc />
+        public MockFileVersionInfo(
```
vbreuss, line 14: Why not set a default value for the optional arguments?
Then I could also specify, e.g.
```csharp
new MockFileVersionInfo("myFile.dll", companyName: "My company");
```
when I am only interested in a specific part of the `IFileVersionInfo`

This should make the other constructors obsolete...

tests/TestableIO.System.IO.Abstractions.Wrappers.Tests/ApiParityTests.cs
```diff
@@ -24,6 +24,13 @@ public void FileInfo() =>
                 typeof(System.IO.Abstractions.FileInfoBase)
             );
 
+        [Test]
```
vbreuss, line 27: Can you also add a parity test fo the `IFileVersionInfoFactory`?

vbreuss (commented):
I have only found some minor comments.
Please also update the version in [version.json](https://github.com/TestableIO/System.IO.Abstractions/blob/main/version.json) to "21.2".

src/TestableIO.System.IO.Abstractions.TestingHelpers/MockFileVersionInfo.cs
```diff
@@ -0,0 +1,207 @@
...
+
+        private static void ParseVersion(string version, out int major, out int minor, out int build, out int revision)
+        {
+            var parts = version.Split('.');
+            if (parts.Length != 4)
+            {
+                throw new FormatException("Version string must have the format 'major.minor.build.revision'.");
+            }
+
+            if (!int.TryParse(parts[0], out major) ||
```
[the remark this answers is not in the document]
t4m45: > The product version is even more liberal (e.g. `<ProductVersion>2.3.4-abd</ProductVersion>` is valid), so please have a more in-depth look into this!

How? Based on the [documentation](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.fileversioninfo.productversion?view=net-9.0), all four sets of the product version are numbers.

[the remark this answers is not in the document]
t4m45: After checking it, I'm very surprised. For the string value "2.3.4-abd" is accepted indeed and in this case, the value 0 is parsed for the ProcuctPrivatePart.
I'll work on this.

src/TestableIO.System.IO.Abstractions.TestingHelpers/MockFileVersionInfo.cs
t4m45: When I tried that, your "Codacy static code analysis" workflow failed. [Link](https://github.com/TestableIO/System.IO.Abstractions/runs/34494707382)

But at that time, all of the parameters were optional. Maybe thats what caused the error.
I will try it again, but this time, the file name will be mandatory.

src/TestableIO.System.IO.Abstractions.TestingHelpers/MockFileVersionInfo.cs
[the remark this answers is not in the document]
t4m45: Done in commit [2e6fa6655927c94f413336c1a4d1510da37eeee5](https://github.com/TestableIO/System.IO.Abstractions/pull/1177/commits/2e6fa6655927c94f413336c1a4d1510da37eeee5)

tests/TestableIO.System.IO.Abstractions.Wrappers.Tests/ApiParityTests.cs
t4m45: I'm confused. There are no parity tests for the factories. To what should I compare the `IFileVersionInfoFactory`?

vbreuss: Sorry, my bad. I have parity tests for factories in [Testably.Abstractions](https://github.com/Testably/Testably.Abstractions/blob/main/Tests/Testably.Abstractions.Parity.Tests/ParityTests.cs) and got confused :-)

t4m45: All good :) Then I resolve this one.

src/TestableIO.System.IO.Abstractions.TestingHelpers/MockFileVersionInfo.cs
t4m45: Well.. I tried and it fails again with "Use the overloading mechanism instead of the optional parameters."

src/TestableIO.System.IO.Abstractions.TestingHelpers/ProductVersionParser.cs
```diff
@@ -0,0 +1,99 @@
+﻿using System.Reflection;
+using System.Text.RegularExpressions;
+
+namespace System.IO.Abstractions.TestingHelpers
+{
+    /// <summary>
+    /// Provides functionality to parse a product version string into its major, minor, build, and private parts.
+    /// </summary>
+    public static class ProductVersionParser
```
vbreuss, line 9: Please mark this class as `internal`, as it should not be required outside of the project.

src/TestableIO.System.IO.Abstractions.TestingHelpers/MockFileVersionInfo.cs
vbreuss: I will have a look...

src/TestableIO.System.IO.Abstractions.TestingHelpers/MockFileVersionInfo.cs
```diff
@@ -0,0 +1,175 @@
+﻿using System.Diagnostics;
+using System.Runtime.InteropServices;
```
vbreuss, line 2:
```suggestion
```
I think this using statement is not required, right?

tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests/MockFileVersionInfoFactoryTests.cs
```diff
@@ -0,0 +1,47 @@
+﻿using NUnit.Framework;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
```
vbreuss, line 6: Please cleanup the using statements:
```csharp
using NUnit.Framework;
```

tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests/MockFileVersionInfoTests.cs
```diff
@@ -0,0 +1,87 @@
+﻿using NUnit.Framework;
+using System.Collections;
```
vbreuss, line 2:
```suggestion
```
Please cleanup the using statements

src/TestableIO.System.IO.Abstractions.TestingHelpers/MockFileVersionInfo.cs
vbreuss: I deactivated this rule in Codacy and it is now working...

t4m45: Cool. Thanks.

vbreuss (approved):
Looks good to me, thanks a lot!

t4m45:
Awesome! Thank you for the help!

PR #1190, opened 2025-01-22
fix: Align MockDirectory and MockFile behavior with .NET's System.IO classes

WestRyanK:
The `MockDirectory` and `MockFile` classes previously did not replicate the .NET behavior of throwing exceptions in the following scenarios:  
- Attempting to delete a directory using `MockFile`.  
- Attempting to delete a file using `MockDirectory`.  

This pull request addresses these inconsistencies by updating the mocks to correctly throw exceptions in these cases, ensuring their behavior aligns with the corresponding `System.IO` classes they are designed to mimic.


vbreuss (approved):
Thanks, @WestRyanK for this bugfix.

Looks good to me!

Issue #1138, opened 2024-08-07
MockFileSystem.Directory.Move fails on Windows if destination has different case from source


El-Gor-do:
**Describe the bug**
On Windows, `FileSystem.Directory.Move` is successful when the destination has a different case than source but `MockFileSystem.Directory.Move` throws `System.IO.IOException: Source and destination path must be different.`

**To Reproduce**
Using `System.IO.Abstractions.TestingHelpers 21.0.29`

```
static void Test()
{
    FileSystem fs = new FileSystem();
    MockFileSystem mockFs = new MockFileSystem();

    Console.WriteLine("Real file system");
    MoveDir(fs);

    Console.WriteLine("Mock file system");
    MoveDir(mockFs);
}

static void MoveDir(IFileSystem fileSystem)
{
    string tempDir = fileSystem.Path.GetTempPath();
    string src = fileSystem.Path.Combine(tempDir, "src");
    string dest = fileSystem.Path.Combine(tempDir, "SRC");  // different case

    try
    {
        // create source directory
        IDirectoryInfo srcDir = fileSystem.DirectoryInfo.New(src);
        srcDir.Create();

        // move directory
        fileSystem.Directory.Move(src, dest);

        Console.WriteLine($"Successfully moved \"{src}\" to \"{dest}\"");
    }
    catch (Exception ex)
    {
        Console.WriteLine(ex.ToString());

        if (fileSystem.Directory.Exists(src))
            fileSystem.Directory.Delete(src);

        if (fileSystem.Directory.Exists(dest))
            fileSystem.Directory.Delete(dest);
    }
}
```

Actual output
```
Real file system
Successfully moved "C:\Users\El-Gor-do\AppData\Local\Temp\src" to "C:\Users\El-Gor-do\AppData\Local\Temp\SRC"
Mock file system
System.IO.IOException: Source and destination path must be different.
   at System.IO.Abstractions.TestingHelpers.MockDirectory.Move(String sourceDirName, String destDirName)
   at IDirectoryInfoMove.Program.MoveDir(IFileSystem fileSystem) in D:\dev\bugs\IDirectoryInfoMove\IDirectoryInfoMove\Program.cs:line 33
```

**Expected behavior**
MockFileSystem.Directory.Move should successfully move the directory.

```
Real file system
Successfully moved "C:\Users\El-Gor-do\AppData\Local\Temp\src" to "C:\Users\El-Gor-do\AppData\Local\Temp\SRC"
[...]

PR #1256, opened 2025-03-24
fix: `Directory.Move` fails on Windows if destination has different case from source

oni-shiro:
This PR address the issue #1138:

On Windows, `MockFileSystem.Directory.Move` no longer throws a `System.IO.IOException` when source and destination path only differ in casing.

### DIFF
--- a/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectory.cs
+++ b/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectory.cs
@@ -504,7 +504,7 @@ public class MockDirectory : DirectoryBase
         var fullDestPath = mockFileDataAccessor.Path.GetFullPath(destDirName).TrimSlashes();
 
         if (XFS.IsUnixPlatform() ? mockFileDataAccessor.StringOperations.Equals(fullSourcePath, fullDestPath)
-            : mockFileDataAccessor.StringOperations.Equals(fullSourcePath, fullDestPath, StringComparison.Ordinal))
+            : string.Equals(fullSourcePath, fullDestPath, StringComparison.Ordinal))
         {
             throw new IOException("Source and destination path must be different.");
         }
@@ -542,7 +542,7 @@ public class MockDirectory : DirectoryBase
                 throw CommonExceptions.CannotCreateBecauseSameNameAlreadyExists(fullDestPath);
             }
         }
-        else if (!mockFileDataAccessor.StringOperations.Equals(fullSourcePath, fullDestPath, StringComparison.OrdinalIgnoreCase))
+        else if (!string.Equals(fullSourcePath, fullDestPath, StringComparison.OrdinalIgnoreCase))
         {
             // In Windows, file/dir names are case sensetive, C:\\temp\\src and C:\\temp\\SRC and treated different
             if (mockFileDataAccessor.Directory.Exists(fullDestPath) || mockFileDataAccessor.File.Exists(fullDestPath))
--- a/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectory.cs
+++ b/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectory.cs
@@ -544,7 +544,7 @@ public class MockDirectory : DirectoryBase
         }
         else if (!mockFileDataAccessor.StringOperations.Equals(fullSourcePath, fullDestPath, StringComparison.OrdinalIgnoreCase))
         {
-            // In Windows, file/dir names are case case sensetive, src and SRC and treated different
+            // In Windows, file/dir names are case sensetive, C:\\temp\\src and C:\\temp\\SRC and treated different
             if (mockFileDataAccessor.Directory.Exists(fullDestPath) || mockFileDataAccessor.File.Exists(fullDestPath))
             {
                 throw CommonExceptions.CannotCreateBecauseSameNameAlreadyExists(fullDestPath);
--- a/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectory.cs
+++ b/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectory.cs
@@ -503,7 +503,8 @@ public class MockDirectory : DirectoryBase
         var fullSourcePath = mockFileDataAccessor.Path.GetFullPath(sourceDirName).TrimSlashes();
         var fullDestPath = mockFileDataAccessor.Path.GetFullPath(destDirName).TrimSlashes();
 
-        if (mockFileDataAccessor.StringOperations.Equals(fullSourcePath, fullDestPath))
+        if (XFS.IsUnixPlatform() ? mockFileDataAccessor.StringOperations.Equals(fullSourcePath, fullDestPath)
+            : mockFileDataAccessor.StringOperations.Equals(fullSourcePath, fullDestPath, StringComparison.Ordinal))
         {
             throw new IOException("Source and destination path must be different.");
         }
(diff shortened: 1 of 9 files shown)


oni-shiro:
I added this check to handle this issue[https://github.com/TestableIO/System.IO.Abstractions/issues/1138](url). As it states, the wrapper class and the mock class differ in behaviour. 
Another fix would be removing this checks from mock class instead of adding them in wrapper class. Let me know your thoughts.

vbreuss (commented):
Could you please add a test (see #1138) to verify the changed behavior?

oni-shiro:
> Could you please add a test (see #1138) to verify the changed behavior?

Sure will do that.

oni-shiro:
Added the test, removed two outdated tests that does not align with this change. @vbreuss

tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests/MockDirectoryTests.cs
```diff
@@ -1857,24 +1842,7 @@ public async Task MockDirectory_Move_ShouldThrowADirectoryNotFoundExceptionIfDes
         // Assert
         await That(action, "Could not find a part of the path 'c:\a'.").Throws<DirectoryNotFoundException>();
     }
-
-    [Test]
-    public async Task MockDirectory_Move_ShouldThrowAnIOExceptionIfDestinationDirectoryExists()
```
vbreuss, line 1862: This is a valid test case, that should be supported by the `MockDirectory`.
Note: Also the real file system throws an `IOException` when the destination directory already exists!

tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests/MockDirectoryTests.cs
```diff
@@ -2204,4 +2172,19 @@ public async Task MockDirectory_Exists_ShouldReturnTrue_IfArgIsFrontSlashAndRoot
 
         await That(fileSystem.Directory.Exists("/")).IsEqualTo(true);
     }
+
+    [Test]
+    public async Task MockDirectory_Move_ShoudlNotThrowException_When_SourceAndDestLocationAreSame()
```
vbreuss, line 2177:
```suggestion
    public async Task MockDirectory_Move_ShouldNotThrowException_When_SourceAndDestinationDifferOnlyInCasing()
```

Issue #1138 is about not throwing an exception on windows, when source and destination only differ in casing. The other behavior should remain unchanged!

vbreuss (changes requested):
@oni-shiro I fear you misunderstood #1138. It mentions only the case that source and destination differ in casing and only on windows. You should not remove the existing test cases, as they verify the correct behavior in other cases that should not be changed.

oni-shiro:
Thank you @vbreuss for pointing my mistake 👍. I have made new changes, for this I had to add another `StringOperation.Equals` method that allows user to pass the `String.Comparison` behaviour. Currently when initializing the `StringOperation` we are passing the `XFS.IsUnixPlatForm()` as a flag to force the checks to be `Ordinal` or `OrdinalIgnoreCase`.
image
Thus I had to add another method that allows user to choose. I have updated the expected test cases alongside that also.

Let me know your thoughts.

tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests/MockDirectoryTests.cs
oni-shiro: Corrected on latest commits.

tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests/MockDirectoryTests.cs
oni-shiro: Corrected on latest commits.

vbreuss:
> Let me know your thoughts.

I don't like this additional overload in the `StringOperations`, just use `string.Equals(x, y, StringComparison)` where needed (we should avoid bloating the public API surface).

oni-shiro:
> > Let me know your thoughts.
> 
> I don't like this additional overload in the `StringOperations`, just use `string.Equals(x, y, StringComparison)` where needed (we should avoid bloating the public API surface).

I was not keen on using Both StringOperations.Equals, and string.Equals simultaneously to avoid any confusion. But I understand your point, I will revert the change for that.

oni-shiro:
Reverted the public api related changes on https://github.com/TestableIO/System.IO.Abstractions/pull/1256/commits/402f597001ec71a69b1dc401ec84996bf63d01c0

src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectory.cs
```diff
@@ -534,12 +535,21 @@ public override void Move(string sourceDirName, string destDirName)
         {
             throw CommonExceptions.CouldNotFindPartOfPath(destDirName);
         }
-
-        if (mockFileDataAccessor.Directory.Exists(fullDestPath) || mockFileDataAccessor.File.Exists(fullDestPath))
+        if (XFS.IsUnixPlatform())
```
vbreuss, line 538: If I understand this if/else-if check correctly, it is equivalent to the following code:
```csharp
if (mockFileDataAccessor.Directory.Exists(fullDestPath) || mockFileDataAccessor.File.Exists(fullDestPath))
{
    // In Windows, file/dir names are case sensetive, C:\\temp\\src and C:\\temp\\SRC and treated different
    if (XFS.IsUnixPlatform() ||
        !string.Equals(fullSourcePath, fullDestPath, StringComparison.OrdinalIgnoreCase)
    {
        throw CommonExceptions.CannotCreateBecauseSameNameAlreadyExists(fullDestPath);
    }
}
```

Am I right?

src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectory.cs
```diff
@@ -503,7 +503,8 @@ public override void Move(string sourceDirName, string destDirName)
         var fullSourcePath = mockFileDataAccessor.Path.GetFullPath(sourceDirName).TrimSlashes();
         var fullDestPath = mockFileDataAccessor.Path.GetFullPath(destDirName).TrimSlashes();
 
-        if (mockFileDataAccessor.StringOperations.Equals(fullSourcePath, fullDestPath))
+        if (XFS.IsUnixPlatform() ? mockFileDataAccessor.StringOperations.Equals(fullSourcePath, fullDestPath)
```
vbreuss, line 506: When `XFS.IsUnixPlatform()` the [`StringOperations` ](https://github.com/TestableIO/System.IO.Abstractions/blob/main/src/TestableIO.System.IO.Abstractions.TestingHelpers/StringOperations.cs) already use `StringComparison.Ordinal`, so this check could be simplified to
```csharp
if (string.Equals(fullSourcePath, fullDestPath, StringComparison.Ordinal))
```

vbreuss (commented):
Thanks, @oni-shiro, I only have some minor suggestions.

src/TestableIO.System.IO.Abstractions.TestingHelpers/MockDirectory.cs
oni-shiro: You would be absolutely right. I should have self-reviewed it once again. Thanks for pointing this out.

oni-shiro: Addressed here : https://github.com/TestableIO/System.IO.Abstractions/pull/1256/commits/99b2fd7276cfe8ddfa7d4d8e7b31efd7f66abe65

vbreuss (approved):
I adjusted the title and description, as they were no longer up-to-date.

The PR looks good to me. Will release it shortly!

Issue #1140, opened 2024-08-09
MockFileSystem.File.Move on Windows and .Net 8 doesn't rename a file name's case


El-Gor-do:
**Describe the bug**
Possibly related to https://github.com/TestableIO/System.IO.Abstractions/issues/1138
On Windows and .Net 8, `FileSystem.File.Move` successfully renames a file if the destination has a different case than source whereas `MockFileSystem.File.Move` doesn't rename the file.

**To Reproduce**
Using `System.IO.Abstractions.TestingHelpers 21.0.29`

```
static void Main(string[] args)
{
    FileSystem fs = new FileSystem();
    MockFileSystem mockFs = new MockFileSystem();

    Console.WriteLine("Real file system");
    MoveFile(fs);

    Console.WriteLine("Mock file system");
    MoveFile(mockFs);
}

static void MoveFile(IFileSystem fileSystem)
{
    string tempDir = fileSystem.Path.GetTempPath();
    string src = fileSystem.Path.Combine(tempDir, "src.txt");
    string dest = fileSystem.Path.Combine(tempDir, "SRC.TXT");  // different case

    // create source file
    using (_ = fileSystem.File.Create(src))
    {
    }

    // move file
    fileSystem.File.Move(src, dest);

    // check the moved file's name exact case
    string movedFile = fileSystem.Directory.GetFiles(tempDir, "src.txt").First();
    Console.WriteLine($"Successfully moved \"{src}\" to \"{movedFile}\"");

    if (fileSystem.File.Exists(src))
        fileSystem.File.Delete(src);

    if (fileSystem.File.Exists(dest))
        fileSystem.File.Delete(dest);
}
```

Actual output
```
Real file system
Successfully moved "C:\Users\El-Gor-do\AppData\Local\Temp\src.txt" to "C:\Users\El-Gor-do\AppData\Local\Temp\SRC.TXT"
Mock file system
Successfully moved "C:\temp\src.txt" to "C:\temp\src.txt"
```

**Expected behavior**
`MockFileSystem.File.Move` should successfully rename the file to `SRC.TXT`
```
Real file system
Successfully moved "C:\Users\El-Gor-do\AppData\Local\Temp\src.txt" to "C:\Users\El-Gor-do\AppData\Local\Temp\SRC.TXT"
Mock file system
Successfully moved "C:\temp\src.txt" to "C:\temp\SRC.TXT"
```

vbreuss:
[...]

PR #1259, opened 2025-04-02
fix: handle case-sensitive file overwrite on Windows and add tests

ana1250:
This PR fixes an issue (#1140) where copying a file with the same name but different casing on Windows resulted in an "already in use" error. It aligns the behavior with the native file system.  

## Changes  
- Fixed file overwrite logic to properly handle case-insensitive paths on Windows.  
- Added Windows-specific checks to mimic `System.IO.File.Copy` `System.IO.File.Move` and `System.IO.File.Replace` behavior.  
- Implemented test cases to verify expected behavior for case-sensitive and case-insensitive file systems. 
- Ensured that the fix does not affect Linux behavior.  

## Testing  
- Verified that the fix works correctly on Windows by checking file existence before and after copying,moving and replacing file with different cases 
- Ran test cases on both Windows and Linux to ensure no unintended side effects.

### DIFF
--- a/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockFile.cs
+++ b/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockFile.cs
@@ -48,7 +48,7 @@ public partial class MockFile : FileBase
             AppendAllBytes(path, bytes.ToArray());
         }
 #endif
-        
+
     /// <inheritdoc />
     public override void AppendAllLines(string path, IEnumerable<string> contents)
     {
@@ -161,6 +161,11 @@ public partial class MockFile : FileBase
             if (!overwrite)
             {
                 throw CommonExceptions.FileAlreadyExists(destFileName);
+            }
+
+            if (string.Equals(sourceFileName, destFileName, StringComparison.OrdinalIgnoreCase) && XFS.IsWindowsPlatform())
+            {
+                throw CommonExceptions.ProcessCannotAccessFileInUse(destFileName);
             }
 
             mockFileDataAccessor.RemoveFile(destFileName);
@@ -522,10 +527,29 @@ public partial class MockFile : FileBase
         mockFileDataAccessor.PathVerifier.IsLegalAbsoluteOrRelative(sourceFileName, nameof(sourceFileName));
         mockFileDataAccessor.PathVerifier.IsLegalAbsoluteOrRelative(destFileName, nameof(destFileName));
 
+        var sourceFile = mockFileDataAccessor.GetFile(sourceFileName);
+
+        if (sourceFile == null)
+        {
+            throw CommonExceptions.FileNotFound(sourceFileName);
+        }
+
+        if (!sourceFile.AllowedFileShare.HasFlag(FileShare.Delete))
+        {
+            throw CommonExceptions.ProcessCannotAccessFileInUse();
+        }
+
+        VerifyDirectoryExists(destFileName);
+
         if (mockFileDataAccessor.GetFile(destFileName) != null)
         {
             if (mockFileDataAccessor.StringOperations.Equals(destFileName, sourceFileName))
             {
+                if (XFS.IsWindowsPlatform())
+                {
+                    mockFileDataAccessor.RemoveFile(sourceFileName);
+                    mockFileDataAccessor.AddFile(destFileName, mockFileDataAccessor.AdjustTimes(new MockFileData(sourceFile), TimeAdjustments.LastAccessTime), false);
+                }
                 return;
             }
             else
@@ -533,18 +557,6 @@ public partial class MockFile : FileBase
                 throw new IOException("A file can not be created if it already exists.");
             }
         }
-
-        var sourceFile = mockFileDataAccessor.GetFile(sourceFileName);
-
-        if (sourceFile == null)
-        {
-            throw CommonExceptions.FileNotFound(sourceFileName);
-        }
-        if (!sourceFile.AllowedFileShare.HasFlag(FileShare.Delete))
-        {
-            throw CommonExceptions.ProcessCannotAccessFileInUse();
-        }
-        VerifyDirectoryExists(destFileName);
 
         mockFileDataAccessor.RemoveFile(sourceFileName, false);
         mockFileDataAccessor.AddFile(destFileName, mockFileDataAccessor.AdjustTimes(new MockFileData(sourceFile), TimeAdjustments.LastAccessTime), false);
@@ -822,6 +834,11 @@ public partial class MockFile : FileBase
             throw CommonExceptions.FileNotFound(destinationFileName);
         }
 
+        if (mockFileDataAccessor.StringOperations.Equals(sourceFileName, destinationFileName) && XFS.IsWindowsPlatform())
+        {
+            throw CommonExceptions.ProcessCannotAccessFileInUse();
+        }
+
         if (destinationBackupFileName != null)
         {
             Copy(destinationFileName, destinationBackupFileName, overwrite: true);
@@ -1066,7 +1083,7 @@ public partial class MockFile : FileBase
 
         mockFileDataAccessor.AddFile(path, mockFileDataAccessor.AdjustTimes(new MockFileData(bytes.ToArray()), TimeAdjustments.All));
     }
-        
+
 #if FEATURE_FILE_SPAN
         /// <inheritdoc cref="IFile.WriteAllBytes(string,ReadOnlySpan{byte})"/>
         public override void WriteAllBytes(string path, ReadOnlySpan<byte> bytes)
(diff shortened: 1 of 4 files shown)


tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests/MockFileTests.cs
```diff
@@ -730,4 +730,20 @@ public async Task MockFile_OpenRead_ShouldReturnReadOnlyStream()
         await That(stream.CanWrite).IsFalse();
         await That(() => stream.WriteByte(0)).Throws<NotSupportedException>();
     }
+
+    [Test]
+    [WindowsOnly(WindowsSpecifics.Drives)]
+    public async Task MockFile_Replace_SameSourceAndDestination_ShouldThrowIOException()
```
vbreuss, line 736:
```suggestion
    public async Task MockFile_Replace_SourceAndDestinationDifferOnlyInCasing_ShouldThrowIOException()
```

tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests/MockFileTests.cs
```diff
@@ -730,4 +730,20 @@ public async Task MockFile_OpenRead_ShouldReturnReadOnlyStream()
...
+    {
+        string sourceFilePath = @"c:\something\demo.txt";
```
vbreuss, line 738: I would prefer avoiding the constructor with a dictionary:
```csharp
        string sourceFilePath = @"c:\something\demo.txt";
        string destFilePath = @"c:\something\Demo.txt";
        string fileContent = "content";
        fileSystem.File.WriteAllText(sourceFilePath, fileContent);
```

This way it is easier to temporarily replace `MockFileSystem` with `FileSystem` to check if the behavior is identical.

tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests/MockFileTests.cs
```diff
@@ -730,4 +730,20 @@ public async Task MockFile_OpenRead_ShouldReturnReadOnlyStream()
...
+        string destFilePath = @"c:\something\Demo.txt";
+        string fileContent = "content";
+
+        var fileSystem = new MockFileSystem(new Dictionary<string, MockFileData>
+        {
+            { sourceFilePath, new MockFileData(fileContent) }
+        });
+
+        await That(() => fileSystem.File.Replace(sourceFilePath, destFilePath, null, true)).Throws<IOException>().HasMessage("The process cannot access the file because it is being used by another process.");
```
vbreuss, line 747: If you add the callback in a separate line, you can still follow the AAA-Pattern:

```csharp
        void Act() => fileSystem.File.Replace(sourceFilePath, destFilePath, null, true);
        
        await That(Act).Throws<IOException>()
            .WithMessage("The process cannot access the file because it is being used by another process.");
```

tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests/MockFileMoveTests.cs
```diff
@@ -249,6 +249,26 @@ public async Task MockFile_Move_ShouldThrowNotSupportedExceptionWhenDestinationP
         await That(action).Throws<NotSupportedException>();
     }
 
+    [Test]
+    [WindowsOnly(WindowsSpecifics.Drives)]
+    public async Task MockFile_Move_CaseOnlyRename_ShouldChangeCase()
+    {
+        string sourceFilePath = @"c:\something\demo.txt";
```
vbreuss, line 256: I would prefer avoiding the constructor with a dictionary:
```csharp
        string sourceFilePath = @"c:\something\demo.txt";
        string destFilePath = @"c:\something\DEMO.TXT";
        string sourceFileContent = "content";
        fileSystem.File.WriteAllText(sourceFilePath, sourceFileContent);
```

This way it is easier to temporarily replace `MockFileSystem` with `FileSystem` to check if the behavior is identical.

tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests/MockFileMoveTests.cs
```diff
@@ -249,6 +249,26 @@ public async Task MockFile_Move_ShouldThrowNotSupportedExceptionWhenDestinationP
...
+
+        var fileSystem = new MockFileSystem(new Dictionary<string, MockFileData>
+        {
+        {
+                sourceFilePath, new MockFileData(sourceFileContent)}
+        });
+
+        fileSystem.File.Move(sourceFilePath, destFilePath);
+
+        await That(fileSystem.FileExists(destFilePath)).IsTrue();
```
vbreuss, line 268: Also here, you would need to adjust the assertions to use the methods from `IFileSystem`:
```csharp
        await That(fileSystem.File.Exists(destFilePath)).IsTrue();
        await That(fileSystem.File.ReadAllText(destFilePath)).IsEqualTo(sourceFileContent);
```

tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests/MockFileCopyTests.cs
```diff
@@ -246,6 +246,19 @@ public async Task MockFile_Copy_ShouldThrowNotSupportedExceptionWhenSourcePathCo
...
+    [WindowsOnly(WindowsSpecifics.Drives)]
+    public async Task MockFile_Copy_ShouldThrowIOExceptionWhenOverwritingWithSameNameDifferentCase()
+    {
+        var fileSystem = new MockFileSystem();
+        string path = @"C:\Temp\file.txt";
+        string pathUpper = @"C:\Temp\FILE.TXT";
+
+        fileSystem.File.WriteAllText(path, "Hello");
+
+        await That(() => fileSystem.File.Copy(path, pathUpper, true)).Throws<IOException>().HasMessage($"The process cannot access the file '{pathUpper}' because it is being used by another process.");
```
vbreuss, line 259: If you add the callback in a separate line, you can still follow the AAA-Pattern:

```csharp
        void Act() => fileSystem.File.Copy(path, pathUpper, true);

        await That(Act).Throws<IOException>()
            .WithMessage($"The process cannot access the file '{pathUpper}' because it is being used by another process.");
```

vbreuss (approved):
Thanks @ana1250, looks really good.
I just have some minor suggestions on improving the tests...

tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests/MockFileCopyTests.cs
ana1250: Yes fixed it. Thanks!

tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests/MockFileTests.cs
ana1250: Updated as per last commit.

tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests/MockFileTests.cs
ana1250: Updated as per last commit.

tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests/MockFileMoveTests.cs
ana1250: Updated as per last commit.

tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests/MockFileTests.cs
ana1250: Updated as per last commit.

ana1250:
Thanks! Looking forward to contributing more.

src/TestableIO.System.IO.Abstractions.TestingHelpers/MockFile.cs
```diff
@@ -163,6 +163,11 @@ public override void Copy(string sourceFileName, string destFileName, bool overw
                 throw CommonExceptions.FileAlreadyExists(destFileName);
             }
 
+            if (string.Equals(sourceFileName, destFileName, StringComparison.OrdinalIgnoreCase) && XFS.IsWindowsPlatform())
```
siprbaum, line 166: in all the other cases we use
`mockFileDataAccessor.StringOperations.Equals(sourceFileName, destinationFileName)` instead of relying on 
`string.Equals(..., ..., StringComparison.OrdinalIgnoreCase)`

Is intentional or an oversight?

vbreuss: This seems to be an oversight.
We had one case [here](https://github.com/TestableIO/System.IO.Abstractions/pull/1256#discussion_r2050797816) where we used `string.Equals` on purpose, as it should compare case-sensitive in all operating systems, but here it seems I overlooked it.

PR #1305, opened 2025-07-02
fix: MockFileSystem's File.WriteAllText, File.ReadAllText, Directory.GetFiles, Directory.Create fail with \\?\C:\foo style paths

### LINKED ISSUE #1304: MockFileSystem's IDirectoryInfo.Create implementation fails with \\?\C:\Foo style paths
**Describe the bug**

I need to use `\\?\C:\Foo` style paths on windows, in order to support long paths. 
```
var longPath = @"\\?\C:\some\really\long\string\would\be\here.txt";
var parent = fileSystem.Directory.GetParent(longPath);
parent.Create(); // This throws an exception because of the `?` character in the path.
```

If I use a regular FileSystem instead of MockFileSystem in the code above, there is no exception thrown.

mostynb:
This is an attempt to fix #1304.

### DIFF
--- a/tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests/MockDirectoryTests.cs
+++ b/tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests/MockDirectoryTests.cs
@@ -747,6 +747,20 @@ public class MockDirectoryTests
         await That(createDelegate).Throws<ArgumentException>();
     }
 
+    [Test]
+    [WindowsOnly(WindowsSpecifics.UNCPaths)]
+    public async Task MockDirectory_CreateDirectory_ShouldSupportExtendedLengthPaths()
+    {
+        // Arrange
+        var fileSystem = new MockFileSystem();
+
+        // Act
+        fileSystem.Directory.CreateDirectory(XFS.Path(@"\\?\c:\bar"));
+
+        // Assert
+        await That(fileSystem.Directory.Exists(XFS.Path(@"\\?\c:\bar"))).IsTrue();
+    }
+
     // Issue #210
     [Test]
     public async Task MockDirectory_CreateDirectory_ShouldIgnoreExistingDirectoryRegardlessOfTrailingSlash()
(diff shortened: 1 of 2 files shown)


vbreuss:
@mostynb : Thanks for the PR. Could you please check the failing actions?

What happens on the real file system when you use `Directory.GetFiles` when the directory is prefixed with `\\?\`? Are the returned strings also prefixed with `\\?\`?

mostynb:
> @mostynb
> 
> > If I give System.IO.Directory.GetFiles() a `\\?\` style path on windows, it returns `\\?\` style paths.
> 
> But this does not work on the `MockFileSystem`, even after this pull request, right?

It did not, so I pushed some changes to make it work, assuming that you always use \\?\ style paths.

However, extended length paths aren't normalised internally in MockFileSystem. eg if I create `\\?\C:\foo\bar.txt` then I can read `\\?\C:\foo\bar.txt`, but `C:\foo\bar.txt` does not appear to exist.

tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests/MockDirectoryTests.cs
```diff
@@ -747,6 +747,30 @@ public async Task MockDirectory_CreateDirectory_ShouldThrowIfIllegalCharacterInP
...
+
+        // Act
+        var directoryInfo = fileSystem.Directory.CreateDirectory(XFS.Path(@"\\?\c:\bar"));
+        fileSystem.File.WriteAllText(@"\\?\c:\bar\grok.txt", "hello world\n");
+
+        // Assert
+        await That(fileSystem.Directory.Exists(XFS.Path(@"\\?\c:\bar"))).IsTrue();
+        await That(directoryInfo.FullName).IsEqualTo(@"\\?\c:\bar");
+        await That(fileSystem.File.ReadAllText(@"\\?\c:\bar\grok.txt")).IsEqualTo("hello world\n");
+        var count = 0;
```
vbreuss, line 765: You can simplify this check with 

```csharp
        await That(fileSystem.Directory.GetFiles(@"\\?\c:\bar")).HasSingle()
             .Which.IsEqualTo(@"\\?\c:\bar\grok.txt");
```

instead of 

```csharp
        var count = 0;
        foreach (var f in fileSystem.Directory.GetFiles(@"\\?\c:\bar"))
        {
            await That(f).IsEqualTo(@"\\?\c:\bar\grok.txt");
            count++;
        }
        await That(count).IsEqualTo(1);
```

vbreuss:
Thanks @mostynb, will publish v22.0.15 shortly!

PR #1340, opened 2025-09-13
fix: Add TypeForwardedTo attribute to forward other moved interfaces 

borisf94:
Following #1338 

This PR adds TypeForwardedToAttribute to fix a problem where a transitive dependency on ver 21 can't find interfaces that were moved to a separate assembly in version 22.

Using the attribute, allows third parties that are built with ver 21 to be redirected to Testably.Abstractions.FileSystem.Interface.dll when trying to resolve the interface in its old location: TestableIO.System.IO.Abstractions.dll

This change adds backwards compatibility with ver 21.
Users of ver 22 are not affected in any way.


vbreuss (approved):
LGTM

PR #1363, opened 2025-10-30
fix: file stream sharing

### LINKED ISSUE #1356: MockFileSystem incorrectly allows shared access to same file
**Describe the bug**
According to C#'s [filesystem API](https://learn.microsoft.com/en-us/dotnet/api/system.io.file.create?view=net-9.0#system-io-file-create(system-string)), if you open a file with `FileShare.None` then it should not be possible to create other file streams to that file. If you do this using the "real" file system on Windows, it will throw an `IOException` with the message: *"The process cannot access the file '-local file path-' because it is being used by another process."*. But using the `MockFileSystem` this does not produce any error, which seems wrong.

**To Reproduce**
Run this code:
```csharp
 MockFileSystem mockFileSystem = new();
 mockFileSystem.FileStream.New("SomeFile.txt", FileMode.Create, FileAccess.ReadWrite, FileShare.None); // Should cause open to fail
 // mockFileSystem.File.Create("SomeFile.txt"); // Does not fail either

 mockFileSystem.FileStream.New("SomeFile.txt", FileMode.Open, FileAccess.Read);


 FileSystem realFileSystem = new();
 realFileSystem.FileStream.New("SomeFile.txt", FileMode.Create, FileAccess.ReadWrite, FileShare.None); // Causes open to fail
 // realFileSystem.File.Create("SomeFile.txt"); Also fails

 realFileSystem.FileStream.New("SomeFile.txt", FileMode.Open, FileAccess.Read);
```

**Expected behavior**
It should fail when creating the "open" stream on the `MockFileSystem`, but it does not.

**Additional context**
 - TestableIO version: 22.0.16
 - .NET: net8.0-windows
 - Windows version: Microsoft Windows 11 Enterprise (10.0.22631 Build 22631)

HarrisonTCodes:
Closes #1356

This PR adds stateful handling of file streams opened with the `FileShare.None` option. If a file stream is attempted to be opened whilst another file stream of the same file path and `FileShare.None` is in use (that is to say, has been opened and not yet disposed), an `IOException` will be thrown and the creation of the second file stream will be disallowed.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,7 +13,7 @@ jobs:
         os: [ ubuntu-latest, windows-latest, macos-latest ]
     runs-on: ${{ matrix.os }}
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
         with:
           fetch-depth: 0
       - name: Setup .NET SDKs
@@ -46,7 +46,7 @@ jobs:
     env:
       DOTNET_NOLOGO: true
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
         with:
           fetch-depth: 0
       - name: Setup .NET SDKs
@@ -77,7 +77,7 @@ jobs:
       SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
       DOTNET_NOLOGO: true
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
         with:
           fetch-depth: 0
       - name: Setup .NET SDKs
@@ -118,7 +118,7 @@ jobs:
     env:
       DOTNET_NOLOGO: true
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
         with:
           fetch-depth: 0
       - name: Setup .NET SDKs
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
         os: [ ubuntu-latest, windows-latest, macos-latest ]
     runs-on: ${{ matrix.os }}
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
         with:
           fetch-depth: 0
       - name: Setup .NET SDKs
@@ -46,7 +46,7 @@ jobs:
     env:
       DOTNET_NOLOGO: true
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
         with:
           fetch-depth: 0
       - name: Setup .NET SDKs
@@ -78,7 +78,7 @@ jobs:
       SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
       DOTNET_NOLOGO: true
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
         with:
           fetch-depth: 0
       - name: Setup .NET SDKs
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -26,7 +26,7 @@
     <PackageVersion Include="NUnit" Version="4.4.0" />
     <PackageVersion Include="NUnit.Analyzers" Version="4.11.2" />
     <PackageVersion Include="NUnit3TestAdapter" Version="5.2.0" />
-    <PackageVersion Include="PublicApiGenerator" Version="11.5.0"/>
+    <PackageVersion Include="PublicApiGenerator" Version="11.5.1"/>
     <PackageVersion Include="aweXpect" Version="2.26.0"/>
     <PackageVersion Include="aweXpect.Testably" Version="0.12.0"/>
   </ItemGroup>


hangy (commented):
I like the idea!

It might be useful to check if [symbolic links](https://github.com/TestableIO/System.IO.Abstractions/blob/72ea54d229a20c2f2078b04f04526677f1fb72cf/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockFile.cs#L211) need additional handling

src/TestableIO.System.IO.Abstractions.TestingHelpers/MockFileStream.cs
```diff
@@ -54,6 +58,11 @@ public MockFileStream(
         this.path = path;
         this.options = options;
 
+        if (_fileShareNoneStreams.ContainsKey(path))
```
hangy: You may want to add some normalization before using the path as the dictionary key. Relative paths passed to the ctor could resolve to the same file system object. The `mockFileDataAccessor` does like this:

https://github.com/TestableIO/System.IO.Abstractions/blob/72ea54d229a20c2f2078b04f04526677f1fb72cf/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockFileStream.cs#L57
https://github.com/TestableIO/System.IO.Abstractions/blob/72ea54d229a20c2f2078b04f04526677f1fb72cf/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockFileSystem.cs#L476
https://github.com/TestableIO/System.IO.Abstractions/blob/72ea54d229a20c2f2078b04f04526677f1fb72cf/src/TestableIO.System.IO.Abstractions.TestingHelpers/MockFileSystem.cs#L131-L142

HarrisonTCodes: Thanks for your comment and comprehensive snippets! This is a great point.

I added some light normalization based on the above in [this commit](https://github.com/TestableIO/System.IO.Abstractions/pull/1363/commits/4e724f5f92fa0fa799fc4366a83e50d5a4428cc3), with the line `path = path.Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar).TrimSlashes();`.

Please let me know if you think this is sufficient. Maybe there is a better approach here, such as:

- a more aggressive refactor, turning the private `FixPath` method  you mentioned above into a more general-use utility method (or a new utility method of similar function), or 

- a similar, more rigorous `FixPath` method belonging to the `MockFileStream` class (however this feels like quite a lot of logic/functionality duplication...)

tests/TestableIO.System.IO.Abstractions.TestingHelpers.Tests/MockFileStreamTests.cs
```diff
@@ -390,4 +390,26 @@ async Task Act() =>
...
+    public async Task MockFileStream_WhenExclusiveStreamOpen_ShouldThrowIOException()
+    {
+        var fileSystem = new MockFileSystem();
+        fileSystem.File.WriteAllText("foo.txt", "");
+        using (new MockFileStream(fileSystem, "foo.txt", FileMode.Open, FileAccess.Read, FileShare.None))
+        {
+            await That(() => new MockFileStream(fileSystem, "foo.txt", FileMode.Open, FileAccess.Read)).Throws<IOException>();
+        }
+    }
+
```
vbreuss: I think you could combine these two tests into one and verify that after you dispose of the stream it no longer throws...

src/TestableIO.System.IO.Abstractions.TestingHelpers/MockFileStream.cs
vbreuss: I think it would be better to de-duplicate this logic into an internal method that can be used whenever we need to normalize a path to ensure it is done the same everywhere.

src/TestableIO.System.IO.Abstractions.TestingHelpers/MockFileStream.cs
```diff
@@ -51,9 +55,15 @@ public MockFileStream(
         ThrowIfInvalidModeAccess(mode, access);
 
         this.mockFileDataAccessor = mockFileDataAccessor ?? throw new ArgumentNullException(nameof(mockFileDataAccessor));
+        path = NormalizePath(path);
         this.path = path;
         this.options = options;
 
+        if (_fileShareNoneStreams.ContainsKey(path))
```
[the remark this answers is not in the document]
vbreuss: I fear the solution is not so easy, as you can have multiple streams with e.g. read access and read share open at the same time and only when all streams got disposed should you be able to open the file with write access. So you would need to create a disposable access reference for each stream and remove this instance upon disposal of the stream from some kind of storage.

For reference: in [Testably.Abstractions](https://github.com/Testably/Testably.Abstractions) I used a `ConcurrentDictionary<IStorageLocation, ConcurrentDictionary<Guid, FileHandle>>` with a custom [`FileHandle`](https://github.com/Testably/Testably.Abstractions/blob/main/Source/Testably.Abstractions.Testing/Storage/FileHandle.cs#L6) class (see [here](https://github.com/Testably/Testably.Abstractions/blob/main/Source/Testably.Abstractions.Testing/Storage/InMemoryStorage.cs#L022)).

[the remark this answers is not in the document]
HarrisonTCodes: Ah yes, multiple streams of the same file open at once...
Thanks for the reference, I'll checkout how its handled in Testably.Abstractions, maybe some of it could be directly ported over?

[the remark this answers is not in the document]
HarrisonTCodes: I believe I have implemented this functionality and the correct handling of the cases described above in this thread, currently using `ConcurrentDictionary<string, ConcurrentDictionary<Guid, (FileAccess access, FileShare share)>>`  (behaviour covered in `MockFileStreamTests.cs`). Do review when you get the chance and let me know any pain points, things I may have missed, or desired changes! (maybe a type alias/class like Testably's `FileHandle` for the access/share tuple...)

src/TestableIO.System.IO.Abstractions.TestingHelpers/MockFileSystem.cs
```diff
@@ -114,6 +119,9 @@ public MockFileSystem(IDictionary<string, MockFileData> files, MockFileSystemOpt
     public IFileSystem FileSystem => this;
     /// <inheritdoc />
     public PathVerifier PathVerifier => pathVerifier;
+    /// <inheritdoc />
+    public ConcurrentDictionary<string, ConcurrentDictionary<Guid, (FileAccess, FileShare)>> FileHandles
```
vbreuss: Why do we need a public property for this?
Could we leave it at least internal to avoid breaking changes if we e.g. decide to use a dedicated class later on?

vbreuss (commented):
I really like this change. I would just make some things internal to enable future refactorings without breaking changes...

src/TestableIO.System.IO.Abstractions.TestingHelpers/MockFileSystem.cs
vbreuss: Yes it is a common pattern. I was just wondering, because you mentioned "maybe a type alias/class like Testably's FileHandle for the access/share tuple", and I am not sure if we would have to extend this dictionary in future (e.g. to support symbolic links, etc.). So one solution would be to make it internal (because we could then change it without breaking changes); another would be to add a dedicated type class (that we can extend later on)).
Your argument is sound, so I would keep it public, but I think it would make sense to use a dedicated type instead of the concurrent dictionary (e.g. introduce a `FileHandles` or `AccessHandles` class), that we could extend later on.

src/TestableIO.System.IO.Abstractions.TestingHelpers/MockFileStream.cs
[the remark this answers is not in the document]
vbreuss: You changes look good. Regarding the `ConcurrentDictionary<string, ConcurrentDictionary<Guid, (FileAccess access, FileShare share)>>` see my other comment:
> I think it would make sense to use a dedicated type instead of the concurrent dictionary (e.g. introduce a FileHandles or AccessHandles class), that we could extend later on.

src/TestableIO.System.IO.Abstractions.TestingHelpers/MockFileSystem.cs
HarrisonTCodes: This makes a lot of sense, thanks :) I have introduced a `FileHandles` class in [a61e568](https://github.com/TestableIO/System.IO.Abstractions/pull/1363/commits/a61e56808b5c7e2ea93fe31f024e004166caac3d) and updated `MockFileSystem` and `MockFileStream` accordingly. As usual, if there is something I have missed or any desired changes, please do let me know!

PR #1454, opened 2026-04-01
fix: MockFileSystem.GetTempPath() returns real OS temp path by default

Mpdreamz:
## Problem

`MockFileSystem.Path.GetTempPath()` returns a hardcoded path derived by mechanically converting the Windows constant `C:\temp\` via `MockUnixSupport.Path`. On non-Windows this produces `/temp/` rather than the standard `/tmp/`, and on any platform where the temp directory has been customised (`TMPDIR` on Linux/macOS, `TEMP`/`TMP` on Windows) it will also diverge from what `System.IO.Path.GetTempPath()` actually returns.

This became apparent while developing [scoped-filesystem](https://github.com/nullean/scoped-filesystem/), a library that restricts file access to known locations and supports injection of special folders like the temp directory. It uses `System.IO.Path.GetTempPath()` to determine allowed prefixes and `mockFs.Path.GetTempPath()` to produce paths inside those prefixes. When the two values diverge, paths that should be permitted are rejected — failures that are easy to miss because the mock returns a well-formed, rooted, separator-terminated path with no hint that it differs from reality.

## Fix

Replace the hardcoded constant with `System.IO.Path.GetTempPath()` when initialising the default temp directory. Callers who need a fixed, reproducible temp path can still pass an explicit value via `MockFileSystemOptions` or the `MockPath` constructor — that behaviour is unchanged.

## Test

Added `GetTempPath_Default_MatchesRealOsTempPath` which asserts that `new MockFileSystem().Path.GetTempPath()` equals `System.IO.Path.GetTempPath()`.


vbreuss (commented):
Thanks for the PR, @Mpdreamz. Looks good in general, I have just some minor comments.

Mpdreamz:
Hey @vbreuss us there anything still outstanding?

vbreuss:
> Hey @vbreuss us there anything still outstanding?

The API and unit tests still fail, that's the main blocking issue.

PR #1455, opened 2026-04-09
fix: update `FileVersionInfos` when calling CopyTo, MoveTo and Replace on a IFileInfo

FrancescoLuzzi:
This PR fixes an issue where copying, moving or replacing a file would lose all the `FileVersionInfo` informations when using a `MockFileSystem`

Changes:
- When creating a new `MockFileData` also copy the `FileVersionInfo`
- Added tests to validate the fix

Open questions:
- should `FileVersionInfo` be cloned or is the current fix enough?


vbreuss (approved):
Thanks, @FrancescoLuzzi, Looks good to me. I don't think it is necessary to clone the `FileVersionInfo`.

